PR #67, opened 2021-06-14
Support running `docker-compose` from archs != amd64

jaimergp:
The proposed changes are needed in an Apple Silicon MBP to emulate `amd64` archs for `conda-store-server`, due to the `fakechroot` dependency.

I have also put a [PR](https://github.com/conda-forge/fakechroot-feedstock/pull/3) to build `fakechroot` on `aarch64` (and `ppc64le`), which might allow us to avoid the emulation, but for now this is needed.

### DIFF
--- a/conda-store-server/Dockerfile
+++ b/conda-store-server/Dockerfile
@@ -1,4 +1,4 @@
-FROM condaforge/miniforge3:latest
+FROM --platform=linux/amd64 condaforge/miniforge3:latest
 
 
 RUN apt-get update \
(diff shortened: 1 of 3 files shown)


jaimergp:
Update: `fakechroot` packages are now available on `conda-forge` for `aarch64` and `ppc54le`. That seems to be enough to make everything run on Apple Silicon without further changes, so this can be closed.

jaimergp:
Reopening because even if it runs correctly under `aarch64`, some parts of the codebase assume we are on Linux 64 anyway. In other words, we need this until #71 is handled.

costrouc (approved):
Thanks @jaimergp! Great to pin the build architecture for now.

PR #75, opened 2021-06-29
Use mamba if available; fallback to conda

jaimergp:
This PR adds support to create envs with Mamba instead of conda (if available).

* The base docker image is now `mambaforge` instead of `micromamba` (we end up downloading a full mamba install anyway due to dependencies)
* It will use mamba to create the envs if it's found in PATH. If not, fallback to `conda` and cross your fingers.
* Peak memory didn't go beyond 750MB for a simple env (1.5GB for normal conda, I recall?)
* Time taken: `conda` takes 3:30 min for a python+flask env. `mamba` did it in 55s.

### DIFF
--- a/conda-store-server/Dockerfile
+++ b/conda-store-server/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=linux/amd64 mambaorg/micromamba
+FROM --platform=linux/amd64 condaforge/mambaforge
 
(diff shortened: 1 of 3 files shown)


jaimergp:
Note this branch stemmed out of #74

costrouc (approved):

PR #76, opened 2021-06-29
Download compressed repodata.json

jaimergp:
Instead of downloading the raw `repodata.json`, get the bz2-compressed version, which we can decompress in memory.

Some profiling shows a significant speed-up with my not-so-stable connection.

`repodata.json`

```python
%%timeit
url = "https://conda.anaconda.org/conda-forge/linux-64/repodata.json"
resp = requests.get(url)
resp.raise_for_status()
data = resp.json()
# 5.76 s ± 921 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

`repodata.json.bz2`

```python
%%timeit
url = "https://conda.anaconda.org/conda-forge/linux-64/repodata.json.bz2"
resp = requests.get(url)
resp.raise_for_status()
data = json.loads(bz2.decompress(resp.content))
# 3.01 s ± 276 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Note this branch stemmed out of #74

### DIFF
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -207,7 +207,8 @@ def conda_build(conda_store):
         build.size = utils.disk_usage(build_path)
 
         build_conda_pack(conda_store, build_path, build)
-        build_docker_image(conda_store, build_path, build)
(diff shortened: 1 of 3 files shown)


costrouc (approved):
Thanks @jaimergp. It looks like this is what mamba does as well.

PR #77, opened 2021-07-09
use conda api to obtain host platform

jaimergp:
There were some instances of hardcoded `linux-64` strings as the default platform, which assumed that the server components were running on `linux/amd64` architectures. This is not necessarily true (e.g. MacOS M1 will report `aarch64`), so instead we use the Conda API to obtain the conda-style platform code.


costrouc:
@jaimergp these additions looks great to me. Looks like there are some black formatting issues.

jaimergp:
@costrouc ready now!

costrouc (approved):

PR #78, opened 2021-07-09
use creation timestamp in env paths

### LINKED ISSUE #44: Environment name unique naming should also a function of time not just sha256 of environment.yaml
# Current database model

Environment -> (specification_sha256, build_id)

build_id -> specification_sha256

specification_id (doesn't point to anything)

# Proposal 

If you build a conda environment at a set point in time it will result in a different set of packages than if it was built 1 day ago. Thus it doesn't make much sense for the artifact to have the specification sha 256 to be in the name.

I want to support a (rebuild) environment action. Thus I really don't think that a specficaition_id is useful when talking about an environment.

Environment -> build_id -> specification_sha256

jaimergp:
Closes #44

### DIFF
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -87,7 +87,7 @@ class Build(Base):
 
     def build_path(self, store_directory):
         store_path = pathlib.Path(store_directory).resolve()
-        return store_path / f"{self.specification.sha256}-{self.specification.name}"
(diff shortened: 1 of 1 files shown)


jaimergp:
This is ready but I wouldn't merge this until #79 is done!

costrouc:
Closing as implemented in #90 in https://github.com/Quansight/conda-store/blob/b260386d154d9394485e12fdf3ed1f3cc72a47ee/conda-store-server/conda_store_server/orm.py#L98

PR #79, opened 2021-07-09
test conda-store-build on docker

jaimergp:
Add a workflow entry so the Docker component for building environments can be tested in CI.

This needs a way to make `start_conda_build` behave like a one-shot function. I used `poll_interval` argument for this. If `n <= 0`, it will behave like one-shot. To provide a valid exit code, we keep track of the exit status of all attempted builds and disable the rescheduling after failures.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -93,6 +93,12 @@ jobs:
   test-compose:
     name: Test conda-store-build on Docker
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        storage-backend:
+          - s3
+          - filesystem
+
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


jaimergp:
Hm, interestingly, the new test reproduces the Docker issues I was seen on MacOS M1, @costrouc! And we are using a Linux host here o.o

jaimergp:
@costrouc is there any reason to have `black` pinned in `setup.py`? Currently: `"black==19.10b0"`. I have `21.6b0` locally and they reformat the source differently.

jaimergp:
Running CI again because apparently the `conda-docker` issues are gone here?

jaimergp:
Well, turns out it does work now. No clue what was causing the [earlier errors](https://github.com/Quansight/conda-store/runs/3047696322?check_suite_focus=true), but it got fixed with the latest additions to `main`. It still fails on MacOS hosts though.

jaimergp:
I'd say this is ready for review/merge once we clarify which `black` version we should use, @costrouc

jaimergp:
Putting this on hold until we have a Celery queue.

costrouc:
Closing due to #111

PR #86, opened 2021-07-14
Turn macos' docker-compose into an override template

jaimergp:
Instead of having a whole copy, just use it as an override like this:

```
docker compose -f docker-compose.yaml -f docker-compose.macos.yaml ...
```

, which will help maintenance.


costrouc:
macos yaml file is now removed so that you can easily use an override template now.

Issue #91, opened 2021-07-23
Multiple mamba proceses attempt to update the cache at the same time


costrouc:
So I'm getting a reproducible error (however maybe there is a workaround that conda-store can do). When mamba env create -f .... -p ... is run an there is not a chached repodata.json the cache runs into a collision error (presumably since the cached repodata is produced within the same second and the first mamba process overwrites the second?).

```
# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<    Traceback (most recent call last):
      File "/opt/conda/lib/python3.9/site-packages/conda/exceptions.py", line 1079, in __call__
        return func(*args, **kwargs)
      File "/opt/conda/lib/python3.9/site-packages/conda_env/cli/main.py", line 80, in do_call
        exit_code = getattr(module, func_name)(args, parser)
      File "/opt/conda/lib/python3.9/site-packages/conda_env/cli/main_create.py", line 141, in execute
        result[installer_type] = installer.install(prefix, pkg_specs, args, env)
      File "/opt/conda/lib/python3.9/site-packages/mamba/mamba_env.py", line 45, in mamba_install
        index = load_channels(
      File "/opt/conda/lib/python3.9/site-packages/mamba/utils.py", line 142, in load_channels
        repo = subdir.create_repo(pool)
    RuntimeError: Could not read JSON repodata file (/opt/conda/pkgs/cache/b556ea5a.json) parse error line 117659`$ /opt/conda/condabin/mamba create -p /var/lib/conda-store/bcfb7b66708f93d61b8684deaffcb5cd8f783a0735ddc9c516dceb755296d4de-20210723-191717-055668-2-python-flask-env -f /tmp/tmpjdlos5_3/environment.yaml`
```

costrouc:
Linked to issue https://github.com/mamba-org/mamba/issues/1104

costrouc:
This still appears to be an issue for mamba during the package download phase.

costrouc:
This is resolved in #417 where we now explicitely download/extract packages using conda internals

PR #100, opened 2021-07-30
support micromamba

jaimergp:
I started working on this as possible solution for #91, but `micromamba`'s lockfile doesn't work perfectly. 

Some extra changes were needed because `micromamba` has a slightly different CLI.


costrouc:
Closing since the newer mamba release should have support for the lockfiles!

PR #103, opened 2021-08-11
Add more authentication methods

### LINKED ISSUE #101: Generic OAuth2, JupyterHub, and Dummy Authentication
Would be nice to support multiple forms of authentication. Currently conda-store will allow any username/password for authentication. Conda-store authentication has been modeled after jupyterhub's authentication implmentation https://github.com/Quansight/conda-store/blob/928ba3874523f878b3dcf81db49da7388bc5393e/conda-store-server/conda_store_server/server/auth.py#L141.

Simply subclass the `Authenticaiton` class and modify the `login_html`, `authenticate`, and modify the routes (for oauth there will need to be a callback url route). 

The AuthorizationBackend, and AuthenticationBackend should give a strong base to build upon.

jaimergp:
Closes #101 

- [X] Configurable traitlet
- [X] Dummy (based on JupyterHub's)
- [x] OAuth

### DIFF
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -76,7 +76,9 @@ class CondaStoreServer(Application):
             app.register_blueprint(views.app_metrics)
 
         app.conda_store = CondaStore(parent=self, log=self.log)
-        app.authentication = self.authentication_class(parent=self, log=self.log)
+        app.authentication = self.authentication_class(
+            parent=self, log=self.log, app=app
+        )
 
         # add dynamic routes
         for route, method, func in app.authentication.routes:
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -94,6 +94,6 @@ class CondaStoreServer(Application):
         app.conda_store.configuration.update_storage_metrics(
             app.conda_store.db, app.conda_store.store_directory
         )
-        # app.conda_store.ensure_conda_channels()
+        app.conda_store.ensure_conda_channels()
 
         app.run(debug=True, host=self.address, port=self.port)
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -2,7 +2,7 @@ import logging
 
 from flask import Flask
 from flask_cors import CORS
-from traitlets import Bool, Unicode, Integer
+from traitlets import Bool, Unicode, Integer, Type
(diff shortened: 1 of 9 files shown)


conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -159,6 +161,20 @@ def ui_get_build(build_id):
     return render_template("build.html", **context)
 
 
+@app_ui.route("/user/", methods=["GET"])
+def ui_get_user():
```
costrouc, line 165: Ahh. I think you are thinking of this differently. Github has a route that should be called within the callback_url method https://api.github.com/user to get the username and organizations that the user is a member of.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -80,6 +94,6 @@ def start(self):
         app.conda_store.configuration.update_storage_metrics(
             app.conda_store.db, app.conda_store.store_directory
         )
-        app.conda_store.ensure_conda_channels()
+        # app.conda_store.ensure_conda_channels()
```
jaimergp, line 97: Oops I need to revert this.

conda-store-server/conda_store_server/server/views/ui.py
costrouc: And conda-store shouldn't keep any information about the user. I'm saying this because every authentication provider is a bit different and may not provide an email/username etc. Additionally not everything logged into conda-store is a user. Often times it will be a service that has authenticated via a conda-store token. So conda-store should only be concerned about storing the authorization that the given authenticated user has (unless we want to handle to complexity of is this a user or a service etc. which I think is unneeded).

jaimergp: I am doing that [here](https://github.com/Quansight/conda-store/pull/103/files#diff-8aca98d789e72a0f566677c4fd621c36710f5f7962296ae7ba9775a69bb3bebfR504) and then saving the user data as json cookie. I didn't get the organizations yet to save a query and because I want to know how generic this needs to be (organizations might be GH-specific).

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -159,6 +161,22 @@ def ui_get_build(build_id):
...
+    conda_store = get_conda_store()
+    auth = get_auth()
+
+    # TODO: how does authorization work?
```
jaimergp, line 169: @costrouc Any idea how to manage this here? I couldn't get this bit to work 😬

PR #109, opened 2021-08-20
add support for url prefixes

### LINKED ISSUE #98: Support for base prefix to all routes
QHub often deploys services behind a given url prefix so that only one single domain is needed for multiple services. Need to add a prefix argument similar to https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L617.

This will involve modifications to the app routes along with jinja templates for the ui.

jaimergp:
Closes #98 

---

* Uses `url_prefix` in all blueprints
* Needs a new blueprint for the custom routes provided by the authentication class. Might be a better way to handle this, specially since these URLs cannot be found through `url_for`. We should standardize the routes and corresponding method names in this case so we can use `url_for`.
* The `/` route (index) is currently targeted by the list environments task. We need to use this in several instances (mainly due to the auth routes) in the templates to indicate "absolute path + prefix if any". If this method changes names or is assigned a different route, we need to manually update it. Maybe there's a way to alias a route to a 2nd method name that is just an alias?


conda-store-server/conda_store_server/server/app.py
```diff
@@ -71,16 +80,16 @@ def start(self):
         app.secret_key = self.secret_key
```
costrouc, line 80: This route also needs to account for the `url_prefix`.

costrouc (commented):
Aside from the CORS bit this PR looks great. I also like how using the url_for makes the templates significantly more robust for routes changes. Thanks!

jaimergp:
Nice! CORS fixed.

I am not very proud of the `/`-index bits in the template, but we can iron those out in the future.

costrouc:
Yeah I think in the future the way to fix this is to have well defined names for these auth routes. Also I'd like to at some point make the auth provider produce a blueprint similar to what you did in app.py. It'd be nice to not have that dynamic magic. Big fan of the blueprint you added there.

costrouc (approved):

PR #119, opened 2021-09-01
Add default username/pw to docker instructions

saulshanabrook:
From @peytondmurray

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -73,7 +73,7 @@ docker-compose.yaml for deployment. The example files required are in
 docker-compose up
 ```
 
-Then visit via your web browser http://localhost:5000. The default username is `admin` and password is `test`.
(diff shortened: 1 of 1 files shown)


docs/installation.md
```diff
@@ -73,7 +73,7 @@ docker-compose.yaml for deployment. The example files required are in
 docker-compose up
 ```
 
-Then visit via your web browser http://localhost:5000
+Then visit via your web browser http://localhost:5000. The default username is `admin` and password is `test`.
```
costrouc, line 76: Any username works the password though however has to be `test`. This is since we are using https://github.com/Quansight/conda-store/blob/main/tests/assets/jupyterhub_config.py#L4

saulshanabrook, line 76:
```suggestion
Then visit via your web browser http://localhost:5000. By default, you can log in with any username and use the password `test`, since [we are using](https://github.com/Quansight/conda-store/blob/a679e5c4d2f2fe7d992fd93c5d90c34b38c513ef/tests/assets/jupyterhub_config.py#L4) the [DummyAuthenticator](https://github.com/jupyterhub/jupyterhub/blob/4e7936056744cdad31d608388a349207196efa56/jupyterhub/auth.py#L1122)
```

saulshanabrook: Thanks for clarifying! I updated the text.

costrouc (approved):
LGTM. Thanks!

PR #125, opened 2021-09-10
Proposal to fix #122 : filter envs by name

### LINKED ISSUE #122: Environment and Packages Rest api endpoints should have filtering parameters
Adding a query parameter for filtering environments and builds.

pierrotsmnrd:
This is a first attempt at fixing issue #122 : filtering parameters on Environments (original issue includes packages as well, this PR only works for environments)

I have tested this manually.
- I have 2 envs, `data_engineering` and `data_visualization`
- `http://localhost:5000/api/v1/environment/` returns all envs
- `http://localhost:5000/api/v1/environment/?name=engine` returns `data_engineering`
- `http://localhost:5000/api/v1/environment/?name=_visualization` returns `data_visualization`
- `http://localhost:5000/api/v1/environment/?name=data_` returns both `data_engineering` and `data_visualization`

Let me know how to improve this. Hope this helps.

### DIFF
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -1,6 +1,7 @@
 from typing import List
 
 from sqlalchemy import func
+from sqlalchemy.sql.elements import or_
 
 from conda_store_server import orm
 from .conda import conda_platform
@@ -20,10 +21,19 @@ def get_namespace(db, name: str = None, id: int = None):
     return db.query(orm.Namespace).filter(*filters).first()
 
 
-def list_environments(db, namespace: str = None, search=None):
+def list_environments(db, namespace: str = None, search: str = None):
     filters = []
     if namespace:
         filters.append(orm.Namespace.name == namespace)
+
+    if search:
+        new_filter = or_(orm.Environment.name.like(f"{search}"),
+                        orm.Environment.name.like(f"%{search}"),
+                        orm.Environment.name.like(f"{search}%"),
+                        orm.Environment.name.like(f"%{search}%")
+        )
+
+        filters.append(new_filter)
 
     return db.query(orm.Environment).join(orm.Environment.namespace).filter(*filters)
 
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -27,14 +27,7 @@ def list_environments(db, namespace: str = None, search: str = None):
         filters.append(orm.Namespace.name == namespace)
 
     if search:
-        new_filter = or_(
-            orm.Environment.name.like(f"{search}"),
-            orm.Environment.name.like(f"%{search}"),
-            orm.Environment.name.like(f"{search}%"),
-            orm.Environment.name.like(f"%{search}%"),
-        )
-
-        filters.append(new_filter)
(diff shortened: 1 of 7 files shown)


conda-store-server/conda_store_server/api.py
```diff
@@ -20,11 +21,21 @@ def get_namespace(db, name: str = None, id: int = None):
...
     filters = []
     if namespace:
         filters.append(orm.Namespace.name == namespace)
 
+    if search:
+        new_filter = or_(
+            orm.Environment.name.like(f"{search}"),
+            orm.Environment.name.like(f"%{search}"),
+            orm.Environment.name.like(f"{search}%"),
+            orm.Environment.name.like(f"%{search}%"),
```
costrouc, line 34: Is there a reason that all four like queries must be run? Would the %search% not match all of the above statements? Could definitely bey misunderstanding of sql like queries

costrouc: I just checked and % is zero or more. While _ is one or more.

costrouc (changes requested):

conda-store-server/conda_store_server/api.py
pierrotsmnrd: When I tested with only `orm.Environment.name.like(f"%{search}%")` as filter, I had some wrong results in edge cases (like searching for "data_" and not retrieving "data_engineering", for instance). This led me to this `or_` thing. But I may have tested wrong as I realize, after testing again, that only `%search%` alone works perfectly.

this inspires me for another PR with solid unit tests over the API.

Anyway, I have commited a simpler version

costrouc: We'll need to sanitize the query. `_` and `%` are reserved for wildcard matches. Also would be nice if we could ensure that the query matches a regex so that we can't have sql injection. But this does look like a great PR!

pierrotsmnrd: I have pushed a much better version in my last commit.
Instead of using `.like()`, let's use `.contains()` with the `autoescape=True` parameter.

More details in the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/14/core/sqlelement.html#sqlalchemy.sql.expression.ColumnOperators.contains)

conda-store-server/conda_store_server/api.py
```diff
@@ -20,11 +20,14 @@ def get_namespace(db, name: str = None, id: int = None):
...
 
 
-def list_environments(db, namespace: str = None, search=None):
+def list_environments(db, namespace: str = None, search: str = None):
     filters = []
     if namespace:
         filters.append(orm.Namespace.name == namespace)
 
+    if search:
+        filters.append(orm.Environment.name.contains(search, autoescape=True))
```
costrouc, line 29: Wasn't aware of `autoescape` :smile:. Beautiful. I was concerned about this.

costrouc (approved):
Looks great. Thanks @pierrotsmnrd . Longer term I'd like to have some fuzzy matching logic but I think that would require more complex database queries e.g. full text search which practically any database these days supports. But I'll leave that issue alone since there would have to be a lot of database specific code written to support this.

PR #155, opened 2021-09-20
Fix #129 : sort environments (and packages)

### LINKED ISSUE #129: Sort the environments returned by `/api/v1/environment/`
Currently, paginated results are returned from a GET request targeting the `/api/v1/environment/` endpoint. Unlike the `/api/v1/package/` endpoint, the environments returned in the response are not sorted. To make the endpoints behave the same, it would be useful to have the environments sorted in the response.

pierrotsmnrd:
This pull request aims at fixing issue #129 : sort environments returned by the API.

- Results from `api/v1/environment/` can now be sorted by env's namespace, or env's name, or both : 
```
?sort_by=namespace                 -> sort by namespace
?sort_by=name                      -> sort by name
?sort_by=namespace&sort_by=name    -> sort by namespace and name
?sort_by=name&sort_by=namespace    -> sort by name and namespace
```

- Also, this PR contains commits to sort results from `api/v1/package/` as well : 
```
?sort_by=channel                 -> sort by channel name
?sort_by=name                    -> sort by package name
?sort_by=channel&sort_by=name    -> sort by channel name and package name
?sort_by=name&sort_by=channel    -> sort by package name and channel name
```
Regarding this point, @costrouc I'll ask for a specific review of [this commit ](https://github.com/Quansight/conda-store/commit/d43cab6d5470e6524424d1d111614365b86c9e0c#). In the core api, in function `list_conda_packages`, I had to add a join on `orm.CondaChannel` to be able to sort packages by channel's name. Let me know if you see any impact I might have missed here.

### DIFF
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -20,7 +20,7 @@ def get_namespace(db, name: str = None, id: int = None):
     return db.query(orm.Namespace).filter(*filters).first()
 
 
-def list_environments(db, namespace: str = None, search: str = None):
+def list_environments(db, namespace: str = None, search: str = None, sort_by:str=None, order:str='asc'):
     filters = []
     if namespace:
         filters.append(orm.Namespace.name == namespace)
@@ -28,7 +28,19 @@ def list_environments(db, namespace: str = None, search: str = None):
     if search:
         filters.append(orm.Environment.name.contains(search, autoescape=True))
 
-    return db.query(orm.Environment).join(orm.Environment.namespace).filter(*filters)
+    result = db.query(orm.Environment).join(orm.Environment.namespace).filter(*filters)
+
+    if sort_by == 'name':
+        field = orm.Environment.name
+        order = field.desc() if order=='desc' else field.asc()
+        result = result.order_by(order)
+
+    elif sort_by == 'namespace':
+        field = orm.Environment.namespace
+        order = field.desc() if order=='desc' else field.asc()
+        result = result.order_by(order)
+    
+    return result
 
 
 def get_environment(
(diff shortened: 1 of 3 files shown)


conda-store-server/conda_store_server/api.py
```diff
@@ -142,7 +146,7 @@ def list_conda_packages(db, search: str = None):
     filters = []
     if search:
         filters.append(orm.CondaPackage.name.contains(search, autoescape=True))
-    return db.query(orm.CondaPackage).filter(*filters)
+    return db.query(orm.CondaPackage).join(orm.CondaChannel).filter(*filters)
```
costrouc, line 149: Perfect. I believe (could be wrong) that sqlalchemy will optimize out the `join` if `CondaChannel` is not used. But either way I think this a great to add.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -58,11 +87,24 @@ def api_list_environments():
     search = request.args.get("search")
 
     limit, offset = get_paginated_args(request)
+
+    allowed_sorts = {
+        "namespace": orm.Environment.namespace,
```
costrouc, line 92: I think this should be `orm.Namespace.name`. Otherwise I believe this will sort by ID's for the namespace.

costrouc (changes requested):
Beside that one line change to the allowed_sorts this looks good to me. A nice to have would be a default sort supplied as well. Since I think that `namespace`, `name` is a sane default.

conda-store-server/conda_store_server/api.py
costrouc: done.

PR #164, opened 2021-09-23
Fix #163 Adding distinct_on query to package rest api endpoint

### LINKED ISSUE #163: Add `distinct_on` to package api list endpoint `/api/v1/package/`
The gator UI needs to know the number of packages that are going to be listed. The current `count` for the api only returns all the packages including all builds and versions. We need to know exactly how many unique (channel name, package name) there are.

costrouc:
Closes #163

This required removing the jsonify bits from `paginated_api_response`


costrouc:
Added the ability to perform distinct queries on certain columns. This should have the same effect.

PR #196, opened 2021-10-22
Include flask templates in built package

yuvipanda:
Without this, trying to access /home results in a
template not found error


costrouc:
Thanks @yuvipanda! I'll check out this PR on Monday. I appreciate you checking out conda store and I'll comment on the hackmd you wrote as well.

PR #205, opened 2021-11-11
Fix #198 : psycopg2-binary

### LINKED ISSUE #198: Make psycopg2 an optional dependency of conda-store-server
sqlalchemy should theoretically just work with sqlite, so it would be nice to not have psycopg2 as a required dependency. The dependency is also on the [pscopg2](https://pypi.org/project/psycopg2/) package, rather than the [psycopg2-binary](https://pypi.org/project/psycopg2-binary) package - so conda-store-server installation fails if a compilation environment and `libpq-dev` are not available (see https://www.psycopg.org/articles/2018/02/08/psycopg-274-released/).

pierrotsmnrd:
Replace `psycopg2` with `psycopg2-binary`.

References :
Issue open by @yuvipanda : https://github.com/Quansight/conda-store/issues/198
The reason behind using the `-binary` version : https://www.psycopg.org/articles/2018/02/08/psycopg-274-released/

### DIFF
--- a/conda-store-server/environment-dev.yaml
+++ b/conda-store-server/environment-dev.yaml
@@ -10,7 +10,7 @@ dependencies:
   - flower
   - redis-py
   - sqlalchemy
-  - psycopg2
+  - psycopg2-binary
   - requests
   - flask
   - flask-cors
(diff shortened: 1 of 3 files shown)


costrouc:
I've removed postgres as a dependency. Sorry @pierrotsmnrd I should have been more responsive on this! SQLAlchemy supports a lot of backends https://docs.sqlalchemy.org/en/14/dialects/ and I've written conda-store so that it doesn't depend on a specific database.

Issue #201, opened 2021-11-02
CORS policy issue when fetching at conda-store endpoint from JupyterLab


jess-x:
To reproduce:
1. set up conda-store sever, add `mytestenv` to `default` namespace
2. fetch DELETE at conda-store `api/v1/environment/default/mytestenv` endpoint from JupyterLab (i.e. `http://localhost:8888`)
3. see error
`Access to fetch at 'http://localhost:5000/api/v1/environment/default/mytestenv' from origin 'http://localhost:8888' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: Redirect is not allowed for a preflight request.`

pierrotsmnrd:
Feedback from @costrouc  : Let's use a blanket permission.

jess-x:
I looked into this
We can maybe add proper heading around
```
        app = Flask(__name__)
        cors_prefix = f"{self.url_prefix if self.url_prefix != '/' else ''}"
        CORS(
            app,
            resources={f"{cors_prefix}/api/v1/*": {"origins": "*"}},
        )
```
in conda_store_server app.py 
Also found some docs [here](https://flask-cors.readthedocs.io/en/latest/configuration.html?highlight=allow%20origin#configuration-options)  and [here](https://flask-cors.corydolphin.com/en/latest/api.html#using-cors-with-blueprints) should be helpful

jess-x:
Looked into this a bit
adding proper heading could maybe help?
```
        cors_prefix = f"{self.url_prefix if self.url_prefix != '/' else ''}"
        CORS(
            app,
            resources={f"{cors_prefix}/api/v1/*": {"origins": "*"}},
        )
```
Found some [docs ](https://flask-cors.corydolphin.com/en/latest/api.html#using-cors-with-blueprints)

costrouc:
@jess-x I believe that you may be using the wrong URL. Could you try `http://<host>/api/v1/environment/default/mytestenv/` notice the slash at the end which is important. Could you try this out and update the issue?

costrouc:
I closed the PR to make all URLS work since I'd prefer the current way that it works in that CORS is only available on the api.

jess-x:
adding the slash at the end works!! thanks @costrouc

PR #207, opened 2021-11-11
Make all CORS routes accessible

costrouc:
This is resolve the issue in #201 

Was able to reproduce this via

```javascript
fetch('http://localhost:5000/api/v1/environment/<namespace>/<name>', {method: 'DELETE'}).then(data => data.json()).then(data => console.log(data))
````

### CODE REVIEW

**Comment by costrouc** [comment]
The URL was incorrect. If you pass the correct URL there are no issues with CORS.

```javascript
fetch('http://localhost:5000/api/v1/environment/<namespace>/<name>/', {method: 'DELETE'}).then(data => data.json()).then(data => console.log(data))
```

PR #254, opened 2022-03-08
Add concurrency setting to example Docker Conda Store config file

gabalafou:
I think this got overlooked in #250


costrouc (approved):
Good add and will probably enhance the user experience. Thanks @gabalafou. I think that I included it in the docs but did not give an example of it.

gabalafou:
Thanks Chris. A little more context. I added this specifically because running `docker compose up` from examples/docker was erroring out. The error message suggested that something was expecting `concurrency` to be defined... just so you know in case that's unexpected behavior

PR #255, opened 2022-03-08
Docker compose add gator

gabalafou:
Before starting, make sure you have adequately provisioned Docker. For reference, my Docker resources look like this:
8 CPUs, 8 GB memory, 1 GB swap, 59.6 GB disk

git clone this repo, then:

```
cd conda-store/examples/docker
docker compose up --build 
```

This will take several minutes. Then open the following link in your browser:
http://localhost:8000

This will open JupyterHub. Login as test user:
user: test
password: test

Should take you to JupyterLab on port 8000. From there you can go to:
Settings > Conda Store Packages Manager

And play around with creating, deleting, cloning environments and adding, removing, updating packages from those environments.

### DIFF
--- a/conda-store/Dockerfile
+++ b/conda-store/Dockerfile
@@ -2,8 +2,8 @@ FROM --platform=linux/amd64 condaforge/mambaforge
 
 USER root
 
-ARG GATOR_GIT_URL=https://github.com/mamba-org/gator.git
-ARG GATOR_GIT_BRANCH=conda-store-integration
+ARG GATOR_GIT_URL=https://github.com/Quansight/gator.git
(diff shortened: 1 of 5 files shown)


costrouc:
Was able to run this with the modifications that I made in #256 and a few tweaks to the condarc and conda_store_config.py for the example. The only notable error I saw @gabalafou and you brought this up earlier was what conda-store does when there are no available environments. 

image

gabalafou:
Going to start documenting issues here in the comments.

Here's one I noticed. We need a better way to display the environment name in the JupyterLab Launcher:

In the JupyterLab Launcher, all you can see for each Conda Store environment name is "default-"

PR #264, opened 2022-03-15
Fix #263 : exact search

### LINKED ISSUE #263: Ability for exact package name match in search
Right now the `search` parameter in the `/api/v1/package` and `/api/v1/build/<build-id>/packages/`. This is a fuzzy match on the package names (with plans to also be searching the description as well).

The goal is to add a (`name`? not sure what sounds best) query parameter to the routes. This should match exactly on the name of the package.

My guess for the easiest way to add this is to modify:
 - https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/api.py#L113
 - https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/api.py#L188

With a new input of either `name` add as an input to those functions and pass in from the api routes.

pierrotsmnrd:
Here is the work to have an exact search in the api for the two packages endpoints : `/api/v1/package` and `/api/v1/build/<build-id>/packages/`.

As explained in the issue #263, I chose to use an `exact` parameter, complementary to `search`, instead of the alternative design (have `search` and `name` parameters, which would imply to choose arbitrarily between one of the two when both parameters are given)

I tested it manually with an env containing `pandas` and `pandas_schema` and querying the API accordingly.

The only solution I found with sqlalchemy to achieve exact match is to use the `like` operator, forcing me to replace occurences of `%`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ docs/_build
 
 # nix
 .direnv
+**/.DS_Store
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -110,11 +110,13 @@ def get_build(db, build_id: int):
     return db.query(orm.Build).filter(orm.Build.id == build_id).first()
 
 
-def get_build_packages(db, build_id: int, search: str = None, exact: bool = False, build: str = None):
+def get_build_packages(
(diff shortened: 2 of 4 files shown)


costrouc:
I mentioned some things in the main issue but I'll merge this since it is a blocker for the frontend development.

PR #271, opened 2022-03-23
Adding permissions view to user ui endpoint along with api endpoint

costrouc:
This PR aims to add an api endpoint that allows the rest api to give applications information about if it is authentication and what permissions it currently has. Note that a user/application does not have to be authenticated to have permissions.

Additionally the UI for the user displays the permissions that the user currently has along with information about what it means.

image


costrouc:
An example of returned information from the `/api/v1/permission/` endpoint is 

```json
{
  "data": {
    "authenticated": false, 
    "entity_permissions": {
      "*/*": [
        "environment::delete", 
        "environment::read", 
        "environment:create", 
        "namespace::read", 
        "environment::update", 
        "namespace::create", 
        "build::delete", 
        "namespace::delete"
      ]
    }, 
    "primary_namespace": "default"
  }, 
  "status": "ok"
}
```

This endpoint should always return a 200 status.

PR #295, opened 2022-04-29
Fix #294 : yaml spec file uses pip options

### LINKED ISSUE #294: Support for pypi options in Conda-Store via specification or other means
See https://stackoverflow.com/questions/65383467/can-conda-be-configured-to-use-a-private-pypi-repo where index url can be specified in the arguments.

For example

```yaml
name: asdf
dependencies:
 - numpy
 - pip:
   - --index-url https://.../../
   - flask
```

pierrotsmnrd:
Fix #294 : Yaml spec file uses pip options.

Tests : 

- First, I've tried a public pypi mirror : 

```yaml
channels:
- conda-forge
dependencies:
- python=3.6
- pandas
- ipykernel
- pip
- pip:
  - --index-url https://mirrors.sustech.edu.cn/pypi/simple
  - flask
name: test_env_3
prefix: null
```
Setting up a pypi mirror is cumbersome. either I use bandersnatch but it requires 250GB (which I haven't), or [this repo](https://github.com/montag451/pypi-mirror) but it doesn't provide wheels. So for the sake of progressing faster, I've used a public mirror. 
This yaml works and builds.

```yaml
channels:
- conda-forge
dependencies:
- python=3.6
- pandas
- ipykernel
- pip
- pip:
  - --index-url https://mirrors.sustech.edu.cn/pypi/simple
  - --test quansight
  - flask
name: test_env_3
prefix: null
``` 
This one doesn't work, as this is expected due to the unallowed pip parameter "--test".

### DIFF
--- a/conda-store-server/conda_store_server/environment.py
+++ b/conda-store-server/conda_store_server/environment.py
@@ -101,11 +101,11 @@ def validate_environment_pypi_packages(
     required_packages: List[str],
 ) -> schema.Specification:
     def _package_names(packages):
-        
+
         result = {}
         for p in packages:
-            if isinstance(p, str) and not p.startswith('--'):
-                result[ Requirement.parse(p).name ] = p
+            if isinstance(p, str) and not p.startswith("--"):
+                result[Requirement.parse(p).name] = p
         return result
 
     def _get_pip_packages(specification):
--- a/conda-store-server/conda_store_server/environment.py
+++ b/conda-store-server/conda_store_server/environment.py
@@ -101,7 +101,12 @@ def validate_environment_pypi_packages(
     required_packages: List[str],
 ) -> schema.Specification:
     def _package_names(packages):
-        return {Requirement.parse(_).name: _ for _ in packages if isinstance(_, str)}
+        
+        result = {}
+        for p in packages:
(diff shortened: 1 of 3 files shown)


costrouc:
From review 

```
Package Index Options:
  -i, --index-url <url>       Base URL of Python Package Index (default https://pypi.org/simple). This should point to a repository compliant with PEP 503 (the simple repository API) or a
                              local directory laid out in the same format.
  --extra-index-url <url>     Extra URLs of package indexes to use in addition to --index-url. Should follow the same rules as --index-url.
  --no-index                  Ignore package index (only looking at --find-links URLs instead).
  -f, --find-links <url>      If a url or path to an html file, then parse for links to archives. If a local path or file:// url that's a directory, then look for archives in the
                              directory listing.

```

Looks like `--extra-urls` needs to be changed to `--extra-index-url`

costrouc (approved):

costrouc:
@pierrotsmnrd I made some additions for better error handling on matches. I found that `.split(' ')` would give a non-intuitive error message to the user of expecting a tuple of size 2 if they give a back package name. Looks great to me and thanks for the work!

PR #300, opened 2022-05-13
Fix #167 : API performances via DB redesign

### LINKED ISSUE #167: REST API Performance on package requests
When requesting `/api/v1/package/?distinct_on=namespace&distinct_on=name&distinct_on=version` the query takes around 1-2 seconds to complete. As this project scales to more users this is untenable. Investigation is needed on how to speed up the query. To me two immediate solutions are apparent. Also both solutions are orthogonal so they could also both be good to have.

 - The `CondaPackage` table is not normalized well. Many packages have the same name and description. Realistically the table should be split into `CondaChannel`, `CondaPackage(channel_id, name)`, `CondaPackageBuild(package_id, version, build, ...)`.

 - A caching layer for database queries. The `/api/v1/package/` route is extremely cachable. The cache could easily be evicted for a given channel based on the last updated datetime in the database.

pierrotsmnrd:
As described in issue #167, the table `conda_package` needs to be split into 2 tables : 
- `conda_package` containing the data about the packages, regardless of the builds : name, license, version, summary ..
- `conda_package_build` containing the data about the builds of a conda package : build number, constrains, md5, size, timestamp ...

Impacts of splitting this into two tables : 
- each environment build (table `build`)  used to be linked to a `conda_package`. After this, they need to be associated to a `conda_package_build`.
- One build has N `conda_package_build`, and one `conda_package_build` can be linked to N builds : this results in a N-N association. the table `build_conda_package_build` is used for this purpose.

Check the new DB schema : 

output

I'm opening this PR as draft because I need to see how the API tests behave on GH actions, I struggle to make them work properly locally. I expect multiple changes to be needed.

Multiple questions for you @costrouc : 
- Question 1 : When I started this work, the tables `solve` and `solve_conda_package` were not there yet. I assume that `solve_conda_package` should point to a `conda_package_build` instead of a `conda_package`. Do you confirm ? If yes, I'll make the change.

- Question 2 : inside the table `conda_package`, multiple data columns are still duplicated. For instance, the license, the description, the summary, is always the same for each package, regardless of the channel. 
Putting these data into as `conda_package_metadata` would make the DB lighter and save space, and improve performances as we would request these metadata only when needed.
This can be done later in a separate PR, or now.

- Question 3 : check the conflict below between my branch and main, in `orm.py`, that I don't resolve on purpose.
The `batch_size` parameter was used to reduce the load on the DB when updating conda channels, right ? 
Now that we don't insert as much data as before, do you think it's still needed ? 

- Question 4 : do you have any metric relative to performances so I can compare ?  Can we imagine having tests that measure the performance and helps comparing the improvements ? it's hard to do so manually.

TODO : 
- [x] Finish API impacts
- [x] Make all tests work
- [x] update documentation : DB schema, API, ...


pierrotsmnrd:
@costrouc The relevant commits are : 
7a485ca3c09955146dff4f2e8ff86b0253247d39
7ecf00135bdf70e8694d4ed9ac28fc643e737146
e5d1495dba711a1ff914c96993e7eaf91901ae23
e955ef95b4fd3d2ae1d035c07bc43f86548a7245
dd290a0415e4a7c1b51834f7bff29e06524284db

pierrotsmnrd:
Meeting with Chris 2022-05-26 : 
- Question 1 : solve_conda_package should point to a conda_package_build instead of a conda_package
- Question 2 : we can split conda_package into conda_package + conda_package_metadata. To do in another PR and test migration with it.
- Question 3 : Multiple processes insert new rows into the conda package table. That leads to contraints errors when adding a package that already exists. SQLAlchemy doesn't seem to have a proper generic solution for this, we would need to do something specific for each DB engine. batch_size is a more convenient solution.
- Question 4 : test with `api/v1/package`, there was 310.000 packages and it was very slow, 1 to 1.5 seconds.


Next steps : 

- [ ] - Resolve question 1 with conda_package_build and solve
- [ ] - Resolve question 3 with batch_size + resolve the conflict in this PR
- [ ] - Open an migration-oriented issue, using question 2 as an example of migration to do
- [ ] - Resolve question 4 as described above
- [ ] -  Finish API impacts
- [ ] - Make all tests work
- [ ] - Update documentation : DB schema, API, ...

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -62,21 +62,54 @@ def task_update_storage_metrics(self):
...
+
+
+@current_app.task(base=WorkerTask, name="task_update_conda_channel", bind=True)
+def task_update_conda_channel(self, channel_name):
+    conda_store = self.worker.conda_store
+
+    task_key = f"lock_{self.name}_{channel_name}"
+
+    is_locked = False
+    lock = conda_store.redis.lock(task_key)
```
costrouc, line 98: @pierrotsmnrd redis was removed as a requirement (due to trying to allow for a standalone mode). Is it possible for us to make this a non-requirement for the locking? I know that redis the the correct way to handle this...

costrouc (changes requested):

conda-store-server/conda_store_server/worker/tasks.py
pierrotsmnrd: @costrouc  We absolutely need a lock of some sort, to avoid updating a channel twice at the same time to prevent from any error. The algorithm to populate the database is based on this assumption - which I guess is fine considering it's now very fast.

What other options do we have for a lock ?

costrouc:
@pierrotsmnrd performance is great love how you preserved the api. I don't see any conflicts except for requiring redis. It is completely fine if it is optional. It is possible to check for redis and if it does not exist fallback with to retrying+backoff if bulk inserts fail.

conda-store-server/conda_store_server/worker/tasks.py
costrouc: @pierrotsmnrd we could also use a filelock which for now will work fine. It will break down when we have multiple celery workers but in that case they should have redis anyways.

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -62,21 +63,87 @@ def task_update_storage_metrics(self):
...
+    }
+    channel_name_sanitized = channel_name
+    for k, v in sanitizing.items():
+        channel_name_sanitized = channel_name_sanitized.replace(k, v)
+
+    task_key = f"lock_{self.name}_{channel_name_sanitized}"
+
+    is_locked = False
+
+    if conda_store.redis_url is not None:
```
costrouc, line 117: An alternative to this pattern and using try/finally is using the https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack. This way you don't have to enter/exit manually.

costrouc (changes requested):
Two examples fail -- examples/docker and examples/standalone. The docker one is failing due to a permissions issue. This just needs a chmod in the https://github.com/Quansight/conda-store/blob/fix167_api_perf/conda-store-server/Dockerfile#L26-L32.

```shell
conda-store-worker_1  | Traceback (most recent call last):
conda-store-worker_1  |   File "/opt/conda/envs/conda-store-server/bin/conda-store-worker", line 8, in <module>
conda-store-worker_1  |     sys.exit(main())
conda-store-worker_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/traitlets/config/application.py", line 978, in launch_instance
conda-store-worker_1  |     app.start()
conda-store-worker_1  |   File "/opt/conda-store-server/conda_store_server/worker/app.py", line 81, in start
conda-store-worker_1  |     self.conda_store.ensure_directories()
conda-store-worker_1  |   File "/opt/conda-store-server/conda_store_server/app.py", line 391, in ensure_directories
conda-store-worker_1  |     os.makedirs(self.store_directory, exist_ok=True)
conda-store-worker_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/os.py", line 225, in makedirs
conda-store-worker_1  |     mkdir(name, mode)
conda-store-worker_1  | PermissionError: [Errno 13] Permission denied: '/opt/conda-store/conda-store'
````

The standalone one is not working due to sqlite not having the alter capability and this needs to be fixed. This can be accomplished with the batch alter seen here https://github.com/Quansight/conda-store/blob/fix167_api_perf/conda-store-server/conda_store_server/alembic/versions/48be4072fe58_initial_schema.py#L161-L169. This uses alter in the postgresql case and copy/drop in the sqlite case where alter is not supported.

```
age_build", type_="unique")
conda-store-server_1  |   File "<string>", line 8, in drop_constraint
conda-store-server_1  |   File "<string>", line 3, in drop_constraint
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/alembic/operations/ops.py", line 218, in drop_constraint
conda-store-server_1  |     return operations.invoke(op)
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/alembic/operations/base.py", line 399, in invoke
conda-store-server_1  |     return fn(self, operation)
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/alembic/operations/toimpl.py", line 184, in drop_constraint
conda-store-server_1  |     operations.impl.drop_constraint(
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/alembic/ddl/sqlite.py", line 90, in drop_constraint
conda-store-server_1  |     raise NotImplementedError(
conda-store-server_1  | NotImplementedError: No support for ALTER of constraints in SQLite dialect. Please refer to the batch mode feature which allows for SQLite migrations using a copy-and-move strategy.
conda-store-server_1  | Traceback (most recent call last):
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/bin/conda-store-server", line 8, in <module>
conda-store-server_1  |     sys.exit(main())
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/traitlets/config/application.py", line 977, in launch_instance
conda-store-server_1  |     app.initialize(argv)
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/site-packages/traitlets/config/application.py", line 110, in inner
conda-store-server_1  |     return method(app, *args, **kwargs)
conda-store-server_1  |   File "/opt/conda-store-server/conda_store_server/server/app.py", line 173, in initialize
conda-store-server_1  |     dbutil.upgrade(self.conda_store.database_url)
conda-store-server_1  |   File "/opt/conda-store-server/conda_store_server/dbutil.py", line 97, in upgrade
conda-store-server_1  |     check_call(["alembic", "-c", alembic_ini, "upgrade", revision])
conda-store-server_1  |   File "/opt/conda/envs/conda-store-server/lib/python3.10/subprocess.py", line 369, in check_call
conda-store-server_1  |     raise CalledProcessError(retcode, cmd)
conda-store-server_1  | subprocess.CalledProcessError: Command '['alembic', '-c', '/tmp/tmp1an46k_i/alembic.ini', 'upgrade', 'head']' returned non-zero exit status 1.
```

costrouc (approved):
Looks great to me. Thanks for all the work @pierrotsmnrd. I was able to run this locally and it was snappy.

PR #310, opened 2022-05-21
More pip fixes

jonzeper:
Weird thing, I swear I had seen an error about `'dict' object has no attribute 'extend'` before, indicating that the pip dependencies was a dict, but....

Actually it looks like the pip dependencies will be a `schema.CondaSpecificationPip`, not a dict.

This fixes an issue we were seeing when using `c.CondaStore.pypi_included_packages`:

- If a `pip:` section existed in the original yaml, a second `pip:` section would be
added with the 'included' packages rather than appening to the original `pip:` section.
For example, creating an environment with this yaml:
  ```
  dependencies:
    - pip:
      - requests
  ```
  And this config:
  ```
  c.CondaStore.pypi_included_packages = ["pandas"]
  ```
  Would result in this:
  ```
  dependencies:
    - pip:
      - requests
    - pip:
      - pandas
  ```

### DIFF
--- a/conda-store-server/conda_store_server/environment.py
+++ b/conda-store-server/conda_store_server/environment.py
@@ -112,16 +112,16 @@ def validate_environment_pypi_packages(
 
     def _get_pip_packages(specification):
         for package in specification.dependencies:
-            if isinstance(package, dict) and "pip" in package:
-                return package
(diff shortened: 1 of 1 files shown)


costrouc (approved):
Thanks for catching this @jonzeper!

PR #314, opened 2022-06-01
Adding alembic and first migration to build DB

pierrotsmnrd:
This PR adds [`alembic`](https://alembic.sqlalchemy.org/) to Conda Store to enable migrations.
It contains the first DB migration to create the DB schema as it already generated currently from the ORM.

Paradigm shift to pay attention to : 
- before this PR, the DB is generated via the ORM. `ORM ➞ DB`
- from this PR, the DB is generated via migrations, and the migrations are generated from the ORM. `ORM ➞ Migrations ➞ DB`

Generating the first migration required adding `use_alter` on `Environment.current_build_id` to allow it to work. It basically says that the Foreign key must be added doing an `ALTER TABLE` instead of adding it during the `CREATE TABLE`. Without it, there's a conflict because Environment needs Build to exist to be created, and vice-versa.

What remains to be done : 
- we now need a way to run `alembic upgrade head` before conda-store-server starts, and disable the auto creation of the DB from the ORM that is done in `orm.new_session_factory`
- We also need to add documentation on how to add a migration, from both model and data perspectives.


costrouc:
> we now need a way to run alembic upgrade head before conda-store-server starts, and disable the auto creation of the DB from the ORM that is done in orm.new_session_factory

Could we take inspiration from jupyterhub here? 
 - https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L146
 - https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L1789-L1790
 - https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L2809
 
This could be added to the `initialize` method in the CondaStoreServer app.py. Looks like we can largely keep the new_session_factory the same? I see they have a check in the method that the alembic version matches https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/orm.py#L1061-L1062.

pierrotsmnrd:
The 3 first points above are done, the 4th one and documentation remain to be done.

pierrotsmnrd:
Discussion with @costrouc  about the documentation : 
- In the "Contributing" page :  Create a "Database" section after "Architecture"
- Put all the stuff related to the DB in it : database model + database migrations (modify the ORM, alembic --autogenerate, review the migration script, migrate data ...)

pierrotsmnrd:
@costrouc ready for review :)

conda-store-server/conda_store_server/server/app.py
```diff
@@ -15,10 +15,13 @@
...
 from conda_store_server import __version__
 
+import conda_store_server.server.dbutil as dbutil
+
 
 class CondaStoreServer(Application):
     aliases = {
         "config": "CondaStoreServer.config_file",
+        "upgrade-db": "CondaStoreServer.upgrade_db",
     }
```
costrouc, line 26: Add a traitlets attribute for this similar to https://github.com/jupyterhub/jupyterhub/blob/main/jupyterhub/app.py#L1297.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -119,6 +122,9 @@ def initialize(self, *args, **kwargs):
         super().initialize(*args, **kwargs)
         self.load_config_file(self.config_file)
 
+        if self.config.CondaStore.upgrade_db:
+            dbutil.upgrade(self.config.CondaStore.database_url)
+
```
costrouc, line 127: Move to below the `self.conda_store = ...` bits and then use `if self.conda_store.upgrade_db: ....` instead. This ensures that CondaStore Configuration object is properly initialized.

costrouc (changes requested):
Small changes requested but otherwise this PR looks great!

conda-store-server/conda_store_server/server/app.py
pierrotsmnrd: Done. I realized I was using this in the config : 
```py
c.CondaStore.upgrade_db = True
```
So the `upgrade-db` traitlet property should go in `CondaStore` class, not in `CondaStoreServer`. That's fixed now.

conda-store-server/conda_store_server/server/app.py
pierrotsmnrd: Done.

costrouc (approved):
Thanks for the hard work on this @pierrotsmnrd!

PR #315, opened 2022-06-02
Change build id from integer to uuid representation

costrouc:
Several reasons for this PR:
 - build id as an integer gives non-user friendly ids
 - wanting to follow a model similar to uploading docker images/tags
 to docker hub this should allow making it easier to mirror
 - correlating task ids to celery tasks to monitor progress if lost
 
Merge after #314 and use alembic for the migration changing the build_id type.

### DIFF
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -1,5 +1,6 @@
 from typing import List
 import re
+import uuid
 
 from sqlalchemy import func, null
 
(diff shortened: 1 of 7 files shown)


costrouc:
This PR would break existing client's assumptions that the build id is an integer. Additionally this is why the tests are failing.

costrouc:
I'd like to close this issue for the time being. Certainly something that would be nice to have as we imagine conda-store to evolve into a docker hub like service.

PR #327, opened 2022-06-23
Conda-Store client

costrouc:
This is building an initial implementation of a python conda-store client. This is needed for some other work I have been doing on creating an nb_conda_kernels like interface for conda_store. For that work I need a good client library for interacting with Conda-Store. Additionally I've learned some exciting things about Conda-Pack which will make this client especially compelling giving it an immediate useful application.

More details to follow.


costrouc:
Images from some of the commands

image

costrouc:
A uri is either <namespace>/<name>, <name>, <namespace>/<name>:<build id>, or <build id> which all reference a build

Currently in the PR https://github.com/Quansight/conda-store/pull/327 I have added what I would call the uncontroversial bits:

    conda info
    conda run <uri> --command ....
    conda download <uri>

I'm looking for the discussion about the environment, build, namespace interaction. I've reduced the commands that I originally added in the PR.

PR #356, opened 2022-07-17
fix 343: add conda_default_allow_channels option

### LINKED ISSUE #343: Allow new channels to be added dynamically
Currently channels are restricted to only those listed in https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/app.py#L100-L107 and this is by design so that we can restrict users to only a subset of channels. However at time it is necessary just freely accept all channels.

I propose that `c.CondaStore.default_allow_channels` be created as a boolean trait. If true any new environment encountered in https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/app.py#L26-L31 and https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/build.py#L43-L48 should be instead added to the database via a sqlalchemy transaction.

Channels can be added via:
 - normalize the channel name https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/conda.py#L23-L28
 - and added via the orm CondaChannel class https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/orm.py#L280-L282

csadorf:
If set, automatically add new channels to the channels database.

Before merge:
- [ ] Implement tests
- [x] Add documentation

Closes #343.

### DIFF
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -42,9 +42,7 @@ def set_build_completed(conda_store, build, logs, packages):
 
         channel_id = api.get_conda_channel(conda_store.db, channel)
         if channel_id is None:
-            raise ValueError(
-                f"channel url={channel} not recognized in conda-store channel database"
-            )
+            channel_id = api.add_conda_channel(conda_store.db, channel)
         package["channel_id"] = channel_id.id
 
         _package = (
@@ -205,9 +203,7 @@ def solve_conda_environment(conda_store, solve):
 
             channel_id = api.get_conda_channel(conda_store.db, channel)
             if channel_id is None:
-                raise ValueError(
-                    f"channel url={channel} not recognized in conda-store channel database"
-                )
+                channel_id = api.add_conda_channel(conda_store.db, channel)
             package["channel_id"] = channel_id.id
 
             _package = (
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -42,7 +42,12 @@ def set_build_completed(conda_store, build, logs, packages):
 
         channel_id = api.get_conda_channel(conda_store.db, channel)
         if channel_id is None:
-            channel_id = api.add_conda_channel(conda_store.db, channel)
+            if conda_store.conda_default_allow_channels:
+                channel_id = api.add_conda_channel(conda_store.db, channel)
+            else:
+                raise ValueError(
+                    f"channel url={channel} not recognized in conda-store channel database"
+                )
         package["channel_id"] = channel_id.id
 
         _package = (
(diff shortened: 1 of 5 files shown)


csadorf:
The new option is set to "True" by default which breaks previous behavior.

csadorf:
@costrouc I think this is more or less ready, however I think we should at least consider whether maybe instead of adding an additional option, setting the `conda_allowed_channels` to an empty list would be more intuitive to users.

I also could use some guidance on the test implementation.

csadorf:
Manually tested with:

```yaml
channels:
- conda-forge
- bioconda
- defaults
dependencies:
- _libgcc_mutex=0.1
- _openmp_mutex=4.5
- bcftools=1.10.2
- blis=0.7.0
- pip:
  - nothing
- ipykernel
name: map-and-call
prefix: null
```

conda-store-server/conda_store_server/build.py
```diff
@@ -42,9 +42,7 @@ def set_build_completed(conda_store, build, logs, packages):
 
         channel_id = api.get_conda_channel(conda_store.db, channel)
         if channel_id is None:
-            raise ValueError(
-                f"channel url={channel} not recognized in conda-store channel database"
-            )
+            channel_id = api.add_conda_channel(conda_store.db, channel)
```
costrouc, line 45: Need a conditional check `if conda_store.default_allow_channels: ...` here because ValueError should be raised if not enabled.

conda-store-server/conda_store_server/build.py
```diff
@@ -205,9 +203,7 @@ def solve_conda_environment(conda_store, solve):
 
             channel_id = api.get_conda_channel(conda_store.db, channel)
             if channel_id is None:
-                raise ValueError(
-                    f"channel url={channel} not recognized in conda-store channel database"
-                )
+                channel_id = api.add_conda_channel(conda_store.db, channel)
```
costrouc, line 206: Need a conditional check `if conda_store.default_allow_channels: ...` here because ValueError should be raised if not enabled.

costrouc (changes requested):
I saw some Black formatting issues and just added two comments on a conditional check that needs to be made. Otherwise this PR looks great to me. Thanks for your work @csadorf

conda-store-server/conda_store_server/build.py
csadorf: Ok, I think I was under the assumption that since the channel list as part of the specification has already been validated, that this is no longer necessary.

csadorf: Realized that this make much sense. Done. ✅

csadorf:
@costrouc I'd suggest that you consider my suggestion concerning the overloading of the `conda_allowed_channels` trait, but otherwise this should be ready to merge.

costrouc:
Thanks @csadorf. Yeah I get what you are saying and that makes sense. I think it would be a good idea to write it like that especially since there is no case where you would want an empty list of channels in the first place. Making it so that we can overload that behavior!

Would you mind making that change? Sorry for the late response

PR #363, opened 2022-07-30
Fix 339 env description

pierrotsmnrd:
This PR adds the feature for environment description.

- The ORM model for Environment has a new `description` field and an alembic migration as been added
- the CondaSpecification model has a description field, allowing to add a description when creating an env
- the API endpoint to update the build of an env, `PUT /environment/{namespace}/{name}/`, used to take a JSON body as parameter, like : `{"build_id":1}`. This endpoint now also accepts a `description` key to update an env description :  `{"build_id":1, "description":"this is a very useful env"}`. Each key is optional, meaning we can update an env build, or an env description, or both.


conda-store-server/conda_store_server/app.py
```diff
@@ -446,9 +446,13 @@ def register_environment(
             environment = orm.Environment(
                 name=specification.name,
                 namespace_id=namespace.id,
+                description=specification.spec["description"],
```
costrouc, line 449: Does this assume that a specification has to have a "description" field? Otherwise this PR looks great to me and I'm ready to merge this. I do want to make sure though that the `description` field is optional within a spec that is submitted.

pierrotsmnrd: I confirm that the `description` field is optional.

I just made the test : 
- When submitting a yaml spec without the description, the env builds. 
- The rendering of the yaml **and** the content in the DB is : empty string 

### Rendering of the current UI : 

> Capture d’écran 2022-08-04 à 08 23 30


### Data in DB : 

> Capture d’écran 2022-08-04 à 08 28 23

costrouc: Awesome thanks!

costrouc (approved):

PR #365, opened 2022-08-01
Setup conda-store helm charts for automated build and publish

HarshCasper:


costrouc:
@HarshCasper please create a deploy key for https://github.com/Quansight/conda-store-helm-chart to deploy the helm chart. After this this look ready to me.

.github/workflows/release.yaml
```diff
@@ -103,3 +103,48 @@ jobs:
...
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
viniciusdc, line 139: Why do we need docker build here? is it to keep the helm resources up to date with the published docker tags?

resources/helm/conda-store/values.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+  service:
+    annotations: {}
+
+  auth:
+    type: simple
+
+    simple:
+      password:
+
```
viniciusdc, line 27: As this will work as the default value template conda-store will look into it when deployed and as a reference point. I would suggest including comments describing its field and its optional values. 

As a ref. https://github.com/PrefectHQ/server/blob/master/helm/prefect-server/values.yaml

resources/helm/tools/build-publish-helm-chart.sh
```diff
@@ -0,0 +1,42 @@
...
+    --platform=linux/amd64 \
+    --platform=linux/arm64
+"
+
+# chartpress needs to run next to resources/helm/chartpress.yaml
+cd resources/helm/
+
+# chartpress use git to push to our Helm chart repository.
+# Ensure that the permissions to the Docker registry are
+# already configured.
```
viniciusdc, line 20: Ah this answers my question

viniciusdc (commented):
@HarshCasper @costrouc 
LGTM already, I left a suggestion about the `values.yaml` which I think will be handy. Regardless, is there a way to encapsulate all work to the helm publish action instead of relying on that bash script?

.github/workflows/release.yaml
viniciusdc: disregard this one. I already got why :smile:

viniciusdc (approved):
LGTM! Thanks @HarshCasper

costrouc (approved):

PR #370, opened 2022-08-05
push/pull container images to/from additionall registries

costrouc:


conda-store-server/alembic/versions/5ad723de2abd_adding_container_registry_value_to_enum.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision = '5ad723de2abd'
+down_revision = '8d63a091aff8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
```
costrouc, line 18: @pierrotsmnrd this has to do with my limited knowledge of sqlalchemy. Is there a good way to modify this enum by adding a new value? Downgrade also sounds tricky since I'd have to drop all build artifacts which are of that enum type.

costrouc: Also this query is postgresql dependant and I'll need to make it cross database.

Is a better approach to rename the column, create a new enum, then delete the old one?

PR #377, opened 2022-08-14
Fix 'should store docker manifest?' check

jonzeper:


costrouc:
Thanks for catching this @jonzeper! I was also fixing this in https://github.com/Quansight/conda-store/pull/376 but I think that your PR is more targeted. Appreciate the help!

PR #381, opened 2022-08-17
Redirect on login and cookie handling

costrouc:


costrouc:
Still a WIP and wondering how we might have a list of approved redirects.

costrouc:
Significant testing to get this working on browsers. I'm still not 100% sure I'm implementing this correctly.

@pierrotsmnrd I could use some help to test this on your end.

1. Visit some page e.g. https://google.com

2. Spin up the `example/docker` this configuration has two important settings `cd example/docker; docker-compose up --build`

```
c.CondaStoreServer.cors_allow_origins = []
c.CondaStoreServer.cors_allow_origin_regex = ".*"
```

Check that from the `https://google.com` in the console the following works

```javascript
fetch("https://conda-store.localhost/conda-store/api/v1/permission/", {credentials: "include"}).then((r) => r.json()).then((d) => console.log(d.data.authenticated)) 
// should show false if haven't logged in
window.location = "https://conda-store.localhost/conda-store/login?next=https://google.com";
// should pass you through the login flow
fetch("https://conda-store.localhost/conda-store/api/v1/permission/", {credentials: "include"}).then((r) => r.json()).then((d) => console.log(d.data.authenticated)) 
// should now show true
```

Try the same with `docker/standalone` example. With the follow (only the urls have been changed).

```
fetch("http://localhost:5000/api/v1/permission/", {credentials: "include"}).then((r) => r.json()).then((d) => console.log(d.data.authenticated)) 
// should show false if haven't logged in
window.location = "http://localhost:5000/conda-store/login?next=https://google.com";
// should pass you through the login flow
fetch("http://localhost:5000/api/v1/permission/", {credentials: "include"}).then((r) => r.json()).then((d) => console.log(d.data.authenticated)) 
// should now show true
```

I've been able to verify that this works for firefox and chrome.

costrouc:
Well this is a bug in cypress:
 - https://github.com/cypress-io/cypress/issues/18690

The dilema is that I need to set `secure=True` on cookies otherwise chrome will not work (firefox will) due to https://developers.google.com/search/blog/2020/01/get-ready-for-new-samesitenone-secure.

Also the `secure=True` trips up aiohttp which enforces secure cookies a little too strictly. Which is worked around in this PR. 

All this is to say cypress doesn't look like it is going to handle this. But this PR should be treated as passing.

pierrotsmnrd:
Here are the results of my tests : 


With `examples/docker` stack : 
- I always had to visit `conda-store.localhost` first at least once, to bypass the certificate validation error. 
- Chrome : I confirm it works in normal mode. It doesn't work in Incognito mode, I get false.
- Firefox (v103.0.2) : the second fetch returns false :-/  Important note : when I run `window.location = ...` again, I get immediately redirected to the `next` url.
- Safari (v15.2): exactly like firefox

with the standalone stack :
- works on Chrome in normal mode, not in incognito mode
- Firefox : doesn't work, I get false
- Safari = the first fetch doesn't work, I get this error : 
```
> fetch("http://localhost:5000/api/v1/permission/", {credentials: "include"}).then((r) => r.json()).then((d) => console.log(d.data.authenticated)) 
[Warning] [blocked] The page at https://www.google.com/?client=safari was not allowed to display insecure content from http://localhost:5000/api/v1/permission/.

[Error] Not allowed to request resource
	Évaluation de la console (Évaluation de la console 1:3)
	evaluateWithScopeExtension
	(fonction anonyme)
	_wrapCall
[Error] Fetch API cannot load http://localhost:5000/api/v1/permission/ due to access control checks.
	Évaluation de la console (Évaluation de la console 1:3)
	evaluateWithScopeExtension
	(fonction anonyme)
	_wrapCall
< Promise {status: "pending"}

```

On Firefox, with the standalone app, here is the full request of the second fetch : 
Query header : 
```GET /api/v1/permission/ HTTP/1.1
Host: localhost:5000
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:103.0) Gecko/20100101 Firefox/103.0
Accept: */*
Accept-Language: fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3
Accept-Encoding: gzip, deflate, br
Referer: https://www.google.com/
Origin: https://www.google.com
DNT: 1
Connection: keep-alive
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: cors
Sec-Fetch-Site: cross-site
```

Response header : 
```HTTP/1.1 200 OK
date: Wed, 24 Aug 2022 07:50:48 GMT
server: uvicorn
content-length: 224
content-type: application/json
access-control-allow-credentials: true
access-control-allow-origin: https://www.google.com
vary: Origin```


I hope this confusing mess helps ...

costrouc:
Thanks for testing all this @pierrotsmnrd! So at best this is not a great solution... since we get unpredictable results. In fact other services like jupyterhub refuse to have crossorigin cookies. https://github.com/jupyterhub/jupyterhub/issues/1087. Basically saying that for anything cross origin tokens are the way to go. I'll need to think on this and how this might be done.

costrouc:
Dropping the cookie functionally and only supporting samesite=strict. This is really the only secure option and the only way to ensure compatibility on all browsers. @pierrotsmnrd as long as the sites have the same hostname (port does not matter) testing should work.

PR #409, opened 2022-11-13
Add new gui from `conda-store-ui` project on `/coatl` route

telamonian:
This adds a `/coatl` route (eg `localhost:5000/conda-store/coatl`) that serves a standalone html page that contains a single div that in turn renders the `App` React component that comprises the new gui's top level.

image

As for the `coatl` name, here's my thinking: the killer feature of the new gui (and of the corresponding `jupyterlab-conda-store` extension that exposes said gui in jlab) is that you can manage your cloud environments (among others) in the cloud. That led me to:

  cloud-conda -> sky-serpent -> coatl

pros:
  - it's very pithy, only 5 short letters (much better than `conda-store-ui` or `jupyterlab-conda-store`)
  - it's at least as inherently memorable as "gator"


conda-store-server/Dockerfile
```diff
@@ -23,8 +23,7 @@ RUN cd /opt/conda-store-server && \
 ENV PATH=/opt/conda/condabin:/opt/conda/envs/conda-store-server/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:${PATH}
 ENV TZ=America/New_York
 
-RUN chown -R 1000:1000 /opt/conda/pkgs && \
```
telamonian, line 26: Is this recursive command really needed here for something?. I found that removing the recursive `chown` of `/opt/conda/pkgs` speeds up image builds by literally ~60X (from 80s to 1.something), and doesn't seem to break the operation of the resulting `conda-store` dev build.

telamonian:
This is currently just an example of how this could be done by vending a static bundle of the `conda-store-ui` js lib. Before this PR can be considered complete, I at least need to make some much needed improvements upstream to the build of the `conda-store-ui` bundle.

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -68,6 +68,30 @@ def ui_list_environments(
     return templates.TemplateResponse("home.html", context)
 
 
+@router_ui.get("/coatl")
+def ui_list_environments(
```
costrouc, line 72: Change the name of this method. It will collide with the `/` route.

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -68,6 +68,30 @@ def ui_list_environments(
...
+    entity=Depends(dependencies.get_entity),
+):
+    orm_environments = auth.filter_environments(
+        entity,
+        api.list_environments(conda_store.db, search=search, show_soft_deleted=False),
+    )
+
+    context = {
+        "request": request,
+        "environments": orm_environments.all(),
```
costrouc, line 88: Isn't the react api hitting the api endpoints. Why does it need to know the environments?

costrouc (changes requested):
I'm totally fine with the name `coatl`. I also thing that long term this UI will replace the default flask/jinja2/bootstrap. If it were to replace the default UI eventually I also think that the boring `beta`/`alpha` would be applicable. 

I also see several css/js bundles. Could we have a minimal bash script and documentation on how to generate those

conda-store-server/conda_store_server/server/views/ui.py
telamonian: not knowing what's required where for fastapi, i literally just copy pasted. What should get cut? If you want you can also just suggest a more minimal/appropriate version of `ui_coatl` (ie what i should have named the enclosing function).

conda-store-server/conda_store_server/server/views/ui.py
telamonian: whoops. ill rename it to `ui_coatl`

conda-store-server/conda_store_server/server/views/ui.py
costrouc: You are just serving static files right? If so you need this function to return a https://fastapi.tiangolo.com/advanced/custom-response/#fileresponse with the path to the html file (assuming it doesn't need templating/auth. You can remove everything from the function since it is not needed. Then make sure that the html file references paths within the static folder. I would suggest using relative links so that you don't have to know the path prefix.

costrouc:
Closing in favor of #444

Issue #392, opened 2022-10-14
Parallel builds of environments


costrouc:
A conda environment install can be broken into several steps:

1. solve environment from environment.yaml
2. download and extract packages to package cache
3. install conda environment 

Currently all of this is done in https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/build.py#L97-L110 via `conda env create -f environment.yaml -p <prefix>` in one monolithic approach. As a first approach (which can be improved later) we need to break this function into several steps.

Replace `conda env create -f environment.yaml -p <prefix>` with

1. `conda-lock -f environment.yml  -p linux-64`
2. Python code to inspect the conda lock yaml file (example [here](https://github.com/conda-incubator/conda-lock/blob/main/example/pangeo/conda-lock.yml)) and download each package with `manager: conda` to the conda package cache directory and extract. This step must use locking using filelock https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/build.py#L135-L137.
3. `conda-lock install -p <prefix>`

costrouc:
You can get the package cache directory via the following. This will help determine where packages are downloaded an extracted. 

```
from conda.base.context import context
context.pkgs_dirs
```

There are two formats that I know of for packages that are downloaded and need to be handled.
 - .tar.gz (straightforward to handle)
 - .conda https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#conda-file-format

costrouc:
@aktech was there any progress in implementing this? Happy to meet and discuss further if this is still a priority.

costrouc:
Closing as implemented in #417

costrouc:
See notes form a matrix conversation https://matrix.to/#/!SgckMCYyQUhdyooKZe:gitter.im/$U5CYJdRgHFBF6pqWSmP_h2HbY8mHTTl7wX-F1Po1JkI?via=gitter.im&via=matrix.org&via=tum.de

PR #417, opened 2022-12-07
Separate conda build process into lock, download, build environment

anirrudh:
As Chris noted in #392, we need to solve this in order to support building environments in parallel.

This PR solves 1) and 3) as noted in that issue, which is to simply:

* 1) Build the environment solve
* 3) Install the environment

We do this utilizing conda-lock. 

The new methods that were introduced (corresponding to this problem) are solve_lock_environment and build_lock_environment, which are called in the larger build_conda_environment function call. 

cc: @costrouc


costrouc:
@anirrudh this looks good these two steps. I don't see anything I would request that needs changing. Next step would be to add the downloads of packages

anirrudh:
Updates On https://github.com/Quansight/conda-store/issues/392

I was able to get the packages downloading and extracting using filelock. For more information, please consult the changes to `build.py`.

Currently, the conda prefix is hard-coded inside the function. I looked around the conda module and wasn't able to find a way to get a prefix outside of the activated environment it was launching from while using a python instance, so I'm not sure if there's a better way to get this information. Perhaps I am missing something. Other than that, the function behaves as it should in accordance with the lockfile API. 

cc: @costrouc

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,85 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+            if filepath.exists():
+                conda_store.log.info(f"SKIPPING {filename} | FILE EXISTS")
+                packages_searched += 1
+
+            else:
+                conda_store.log.info(f"DOWNLOAD {filename} | {num_files_log}")
+                res = requests.get(url, stream=True)
+
+                with open(filepath, "wb") as tarball:
+                    tarball.write(res.raw.read())
```
costrouc, line 164: This is inefficient since it will read the entire file into memory. Use https://docs.python.org/3/library/shutil.html#shutil.copyfileobj. Which will copy to disk in chunks efficiently.

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,85 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+                    packages_searched += 1
+                    tarball.close()
+
+                if ".tar" in filepath.__str__():
+                    folderpath = filepath.with_suffix("")
+                    folderpath.mkdir(parents=True, exist_ok=False)
+                    conda_store.log.info(f"EXTRACT {filename} | PATH: {folderpath}")
+                    tf = tarfile.open(filepath)
```
costrouc, line 172: Use a context manager here https://docs.python.org/3/library/tarfile.html#tarfile-objects.

costrouc (changes requested):
Overall structure looks great only a few things I'd change for performance and reliability. Additionally after this we will need to implement the libarchive conda format.

anirrudh:
Cool! Seems like I just missed a couple things - will update and resolve!

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,85 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+    )
+
+
+def fetch_and_extract_packages(conda_store, lock_filename):
+    """Fetch links from a conda-locked build recipe and then
+    gets a filelock on the required folder.
+    """
+    import tarfile
+
+    prefix = pathlib.Path("/opt/conda/pkgs/")
```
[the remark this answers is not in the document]
anirrudh: Hey @costrouc, 

I tried the suggested solution, which then resulted in: `/opt/conda/envs/...` which is the same value I get when I do `conda.CONDA_SOURCE_ROOT` or `conda.CONDA_PACKAGE_ROOT`. The best I can do to not hardcode this path is to derive it from one of these attributes, so I will do that!

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,85 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+        )
+
+    spec = dict()
+
+    with open(pathlib.Path(lock_filename)) as f:
+        spec = yaml.safe_load(f)
+
+    packages_searched = 1
+
+    for p in spec["package"]:
```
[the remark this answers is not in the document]
anirrudh: Checking the manager attribute!

conda-store-server/conda_store_server/build.py
anirrudh: Switched to context manager

conda-store-server/conda_store_server/build.py
anirrudh: Done!

anirrudh:
Currently, the lock files are left even after exiting the context manager for filelock. I have a few ideas on how to resolve and remove them cleanly, including putting them into a tempdir while we are using this function. That way, they disappear when we exit context. 

https://github.com/tox-dev/py-filelock/issues/76 is the best example of this.

costrouc:
> Currently, the lock files are left even after exiting the context manager for filelock

Yeup this is expected. https://github.com/tox-dev/py-filelock/blob/main/src/filelock/_unix.py#L44-L46 says there are some race condition issues. I'm completely fine with us leaving these files behind.

anirrudh:
Awesome, are we good for merge then or should I also give implement the libarchive conda format as a part of this PR?

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,99 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+
+                        folderpath = pathlib.Path(
+                            str(filepath).replace(file_extension, "")
+                        )
+                        folderpath.mkdir(parents=True, exist_ok=False)
+
+                        conda_store.log.info(f"EXTRACT {filename} | PATH: {folderpath}")
+                        with tarfile.open(filepath, extensions[file_extension]) as tf:
+                            tf.extractall(folderpath)
+                            tf.close()
```
costrouc, line 189: No need to explicitely close using a contextmanager

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,99 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+            pass
+
+        else:
+            url = p["url"]
+
+            filename = pathlib.Path(url.split("/")[-1:][0])
+            filepath = prefix.joinpath(filename)
+            lockfile = filelock.FileLock(f"{filepath.__str__()}.lock")
+
+            num_files_log = f"{packages_searched} of {len(spec['package'])}"
```
costrouc, line 161: Could this just be directly added to the log statement below so we don't have to hunt down the line it updates?

costrouc (changes requested):
Minor changes. Now we need to implement the .conda format.

conda-store-server/conda_store_server/build.py
anirrudh: good catch!

conda-store-server/conda_store_server/build.py
anirrudh: done!

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,108 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+                        folderpath.mkdir(parents=True, exist_ok=False)
+
+                        conda_store.log.info(f"EXTRACT {filename} | PATH: {folderpath}")
+                        with tarfile.open(filepath, extensions[file_extension]) as tf:
+                            tf.extractall(folderpath)
+
+                    if ".conda" in filepath.__str__():
+                        libarchive.extract_file(filepath)
+                        for zip in filepath.glob('*.zst'):
+                            if "info" in str(zip):
```
costrouc, line 193: Could we use a more robust regex here? Or something more strict such as startswith/endswith etc.

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,108 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+                                p = filepath.joinpath('/info/')
```
costrouc, line 194: Can be written as `filepath / "info/"`

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,108 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+                        folderpath = pathlib.Path(
+                            str(filepath).replace(file_extension, "")
+                        )
+                        folderpath.mkdir(parents=True, exist_ok=False)
+
+                        conda_store.log.info(f"EXTRACT {filename} | PATH: {folderpath}")
+                        with tarfile.open(filepath, extensions[file_extension]) as tf:
+                            tf.extractall(folderpath)
+
+                    if ".conda" in filepath.__str__():
```
costrouc, line 190: Make stricter `.endswith('.conda')`

conda-store-server/conda_store_server/build.py
```diff
@@ -110,6 +111,108 @@ def build_environment(conda_command, environment_filename, conda_prefix):
...
+                        libarchive.extract_file(filepath)
+                        for zip in filepath.glob('*.zst'):
+                            if "info" in str(zip):
+                                p = filepath.joinpath('/info/')
+                                pathlib.Path.mkdir(p)
+                                libarchive.extract_file(zip, p)
+                        pass
```
costrouc, line 197: unneeded line

costrouc (changes requested):

costrouc (approved):

costrouc:
@anirrudh I made a few changes. Mainly making sure that we are leveraging the work that the conda folks have been doing with 
 - https://github.com/conda/conda-package-streaming
 - https://github.com/conda/conda-package-handling

PR #418, opened 2022-12-19
adding missing config file for standalone

pierrotsmnrd:
The docs show, under the "Installation" page, the following command : 

```sh
conda install conda-store-server>=0.4.10
conda-store-server --standalone --config tests/assets/conda_store_standalone_config.py
```

Unfortunately, the file was missing so far. This PR adds it.

@costrouc feel free to modify it if you spot anything wrong.


costrouc:
@pierrotsmnrd I remember removing this file since I wanted standalone to be able to run without a config file. Could you instead remove that mention in the docs to use `--config` or just say that a config file is optional?

pierrotsmnrd:
@costrouc I've removed the file and updated the docs. Let me know if that works for you. 
I have put any mention of the config files as this is the very first commands the users will try, so I assumed it's better to keep it simple.

costrouc (approved):
Perfect thanks!

Issue #413, opened 2022-11-23
Environment Variables parsed away


StefanBrand:
I tried to specify `variables` key as described in Conda docs: [Setting environment variables](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables):

> Further, you can declare environment variables in the environment.yml file as shown here:

```
name: env-name
channels:
  - conda-forge
  - defaults
dependencies:
  - python=3.7
  - codecov
variables:
  VAR1: valueA
  VAR2: valueB
```
After submission the new build does not include the `variables`. Activating an environment built in this way does not set environment variables.

dharhas:
Thats interesting, I wonder when the `variables` feature was added to conda, I was not aware of it. This should be a straightforward fix. @anirrudh can you take a look at this.

anirrudh:
Yep, sorry just got around to this notification while looking at issues.

anirrudh:
cc: @dharhas @costrouc 

Hey @StefanBrand, I'll be working on getting this branch merged, but we have environment variables now working - they are not parsed away! Thanks for catching this and bringing it to our attention.

We are working on merging some major changes, but the branch is called feature/add-variable-parsing. Feel free to use it and let us know if you run into anything -- I'll close this issue once the PR is merged.

anirrudh:
@StefanBrand this has been merged. Try running from the latest main. Closing Issue!

PR #424, opened 2023-01-24
Implement variable parsing for environments

anirrudh:
This solves issue #413 by updating the schema and the according unpacking of those variables. 

cc: @dharhas @costrouc

### DIFF
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -158,7 +158,7 @@ def build_environment(conda_command, environment_filename, conda_prefix):
     )
 
 
-def parse_conda_environment_variables(
+def set_conda_environment_variables(
     conda_prefix: pathlib.Path, environment_variables: Dict[str, Union[str, int]]
 ):
     """Takes an input of the conda prefix and the, variables defined in the environment yaml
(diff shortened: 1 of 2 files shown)


conda-store-server/conda_store_server/build.py
```diff
@@ -155,6 +158,29 @@ def build_environment(conda_command, environment_filename, conda_prefix):
     )
 
 
+def set_conda_environment_variables(
```
costrouc, line 161: Is this required? It looks like it is built into it https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables. I think all that is needed is https://github.com/Quansight/conda-store/pull/424/files#diff-127657f20e6c5c5301f08bded4596eca413c74ef25a9d0f5c1e12fada1bfe16cR191 for this PR.

anirrudh: Hmmm - I recall trying it without this function because I thought it would be handled by conda, but no environment variables were found or activated!

anirrudh: I'll try it and report back shortly!

costrouc: And if not we should be using this file instead https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#macos-and-linux

anirrudh: It is using those files :) trying it without this method, I wasn't able to get vars set.

anirrudh: It is using those files :) trying it without this method, I wasn't able to get vars set.

costrouc (approved):
LGTM. Thanks @anirrudh!

Issue #308, opened 2022-05-20
Builds should be in reverse chronological order.  


dharhas:
Current ordering is alphabetical (I think) and looks odd. 

We should switch to reverse cronological / descending order so the most recent build is at the top.

<img width="623" alt="image" src="https://user-images.githubusercontent.com/561301/169434964-85caa007-c8d7-4756-a3fc-83ec997dcc6b.png">

anirrudh:
I've fixed this in a branch alongside some other improvements I'm working on. I'll link the PR when I raise it.

kcpevey:
This is implemented in the conda-store-ui

Issue #306, opened 2022-05-20
Ability to cancel a build.


dharhas:
I made a mistake in my YAML spec and realized I needed to add some more packages. There is no way to stop a long running build.

costrouc:
This looks possible https://stackoverflow.com/questions/8920643/cancel-an-already-executing-task-with-celery

costrouc:
Linking https://github.com/conda-incubator/conda-store/pull/531. There is no progress to report from @nkaretnikov

nkaretnikov:
Related, but can be done separately: https://github.com/conda-incubator/conda-store/issues/547

nkaretnikov:
Status update:
- Reviewed the linked PR. It's almost ready.
- Still need to make changes, but I currently have no dedicated hours for this. Might happen during unallocated hours this week or in 1 week.

nkaretnikov:
Status update: this should be fixed by #531, it's in review.

PR #427, opened 2023-02-02
Implement build cancelling

anirrudh:
This PR adds the ability to cancel a build from the built-in UI. It upgrades the database, api, and rendered content accordingly. 

This PR solves #306 and #308 as a byproduct. 

I can revert the flake.nix to the old version, but needed an environment that included alembic - the old one still returns a 404, and I could address the poetry2nix in another PR. 

With some additional time, I could also include tests as a part of this PR. 

cc: @dharhas @costrouc


costrouc:
Going to close now. We will be implementing build canceling soon just need to think it through a bit more.

PR #445, opened 2023-04-05
Fix #439 : conda indexed channels

### LINKED ISSUE #439: Conda-Store trait for indexed channels
Currently the indexed conda channels are the same as the allowed channels. We should separate this logic so that we can control which channels have autocomplete even when we explicitly allow all channels.

These are the channels for which we know the packages that are available. This will fetch/prepopulate the available packages. I am completely open to a better trait name.

pierrotsmnrd:
This PR fixes issue #439 

The default indexed channels are `main` and `conda-forge`.

The issue suggested the name `conda_indexed_channels`. For clarity, I'd suggest to rename it as `conda_prefetched_channels` or something similar. 
The name `conda_indexed_channels` confused me at first, because "indexed channels" means to me "channels already indexed", and not "channels to index".

### DIFF
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -125,10 +125,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     conda_indexed_channels = List(
-        [
-            "main",
-            "conda-forge",
-        ],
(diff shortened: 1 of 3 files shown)


costrouc (approved):
I rebased the commits and ran locally. Works as expected. I checked by adding the `nvidia` to indexed channels.

Issue #359, opened 2022-07-28
Channel priority should be set to strict and conda info should be printed with conda-store


costrouc:
Right now some solves are pulling in multiple channels that were not specified.

costrouc:
I think that channel priority strict is already set. It would be nice to have the output of:
 - `conda info`
 - `conda config --show`

kcpevey:
I'd also like to confirm that we are setting channel priority to strict and ensure that its not pulling in things from defaults or anywhere else.

nkaretnikov:
Note: the block of code that Chris refers to above is actually this (no hash was specified in the link and it's been refactored since then):

`git show 5f7adfb5737edef307f0a70a42f7d04d66adce30:conda-store-server/conda_store_server/build.py`:
```py
145 def build_environment(conda_command, environment_filename, conda_prefix):
146
147     return subprocess.check_output(
148         [
149             conda_command,
150             "env",
151             "create",
152             "-p",
153             conda_prefix,
154             "-f",
155             str(environment_filename),
156         ],
157         stderr=subprocess.STDOUT,
158         encoding="utf-8",
159     )
```

nkaretnikov:
Status update: this should be fixed by #693.

Issue #435, opened 2023-03-21
Emit logs frequently for debugging


costrouc:
Instead of only emitting logs at the end of a job completing/finishing instead we should emit logs at a set frequency. This will require the build to potentially be an async process to be able to periodically update the logs on file/object storage.

costrouc:
This is an example of how logs are uploaded https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/build.py#L24-L31.

Right now this is only done on success/failure of the build. Instead we would like to be able to successively update/upload the logs.

https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/build.py#L147-L159 should be continually uploading logs at a set frequency that is configurable on `CondaStore.log_update_frequency = ... [s]`.

costrouc:
Redis would result long term in the best user experience. 

1. check that redis is enabled
2. Potentially use redis streams https://redis.io/docs/data-types/streams-tutorial/ (but worth investigating if this is the right approach).


If a build is not completed/failed we can stream the logs via websockets, otherwise redirect the user to the storage for the logs (current behavior).

costrouc:
Closing since frequent logs are currently emited. Please open another issue if more detailed logs are needed. It is hard to balance frequently writing logs and storing them on s3/object storage.

PR #456, opened 2023-04-11
log environment build configuration

leej3:
Towards resolving #359. Logs the output of `conda info` and `conda config --show` along with the build command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
  - conda-store and conda-store-server images are now deployed to quay.io/Quansight, which has support for podman and rkt. (#455)
  - Parallel builds of conda environments (#417)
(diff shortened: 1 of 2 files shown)


leej3:
@costrouc I tried to use io.StringIO here but unfortunately it has no fileno attribute which is required for the subprocess methods to believe it is indeed a file. I had a read around for implementing this in a non-blocking way.  I found [this answer](https://stackoverflow.com/a/18423003) useful (and indeed the whole series of answers). There are some things that start to make this tricky:

+ It seems like using subprocess.PIPE  can block/deadlock I believe this is using the os pipe buffer which sometimes won't be fast enough. There's a warning about this in subprocess.poll docs. 
+ Multiple pipes make things harder, 
+ buffering in the launched process can be a problem
+  there’s a different between “live” and reading line by line (likely not relevant in our case)

leej3:
I reduced the scope of the changes here. The asynchronous stuff can be addressed entirely in #435

Issue #460, opened 2023-04-13
Development ability to reload fastapi server with changes


costrouc:
Currently the code is copied into the container and an install is performed. Instead for development we should:
 - mount the code
 - install in development mode

Would like a hatch to do this

trallard:
Duplicate of #901

PR #464, opened 2023-04-14
WIP: enable faster server reload

leej3:
Towards resolving #460, this resolves issues with mounting the source into the containers.

This isn't working yet because `uvicorn.run(...,reload=True)` requires that the app argument is specified as a import string, something like "my_package.my_module.app" but currently the app is instantiated within the   `CondaStoreServer.start` method. Perhaps we could return the instantiated app from `CondaStoreServer.__call__`? I haven't worked with fastapi or CondaStoreServer to have a good sense of the best solution.

Also, the Hatch target has not been added yet.

### DIFF
--- a/conda-store-server/Dockerfile
+++ b/conda-store-server/Dockerfile
@@ -11,30 +11,26 @@ RUN apt-get update \
 # we do not want to tamper with the solve
 RUN printf 'channels: []\n' > /opt/conda/.condarc
 
-RUN chown -R 1000:1000 /opt/conda && \
-    mkdir -p /opt/conda-store/conda-store && \
-    chown 1000:1000 /opt/conda-store/conda-store && \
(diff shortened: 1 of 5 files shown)


trallard:
@costrouc this has been waiting for your review for a while, a quick check would be appreciated

costrouc:
> This isn't working yet because uvicorn.run(...,reload=True) requires that the app argument is specified as a import string, something like "my_package.my_module.app" 

Yeah this is my concern about this PR. I would prefer supporting server reload with the `--standalone` setting. Though I do not know how we could support this since we are launching `fastapi` via Traitlets. Basically I think a major rework is needed and there are several steps:
 - how to pass the configuration file/settings?
 - how to remove Traitlets and the `start` code

PR #474, opened 2023-04-27
push images for main

leej3:
I could add sha as a tag as well... it would make usage for downstream's more reproducible but would clutter the tags listed on the registries.


costrouc:
> I could add sha as a tag as well... it would make usage for downstream's more reproducible but would clutter the tags listed on the registries.

Yes please :smile: I would prefer this over just a main tag since that is not reproducible at all. Kubernetes and containers in general it is bad practive to use `latest`/`main`/`develop` etc. since you don't know what is in the image.

costrouc (changes requested):
Requesting the changes I mentioned above.

.github/workflows/release.yaml
```diff
@@ -2,6 +2,8 @@ name: "Release"
 
 on:
   push:
+    branches:
```
costrouc, line 5: This will also trigger the pypi releases. Could we separate this workflow into two workflows so we can be more explicit on the triggers?

costrouc (approved):

costrouc:
Thanks! Will be nice to be able to run main docker images frequently.

PR #477, opened 2023-04-28
Fix #476 : delete dangling build artifacts before migration

### LINKED ISSUE #476: Alembic migration failing in some edge case
We have identified a case when one alembic migration fails with some edge case data.

Migration with revision [`16f65805dc8f` ](https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/alembic/versions/16f65805dc8f_split_conda_package_into_conda_package_.py) fails [here](https://github.com/Quansight/conda-store/blob/ad41419909569d0ac9aa8973a4eb539c1a86da5d/conda-store-server/conda_store_server/alembic/versions/16f65805dc8f_split_conda_package_into_conda_package_.py#L217) in case of data inconsistency.

In the identified case, the reason is dangling build artifacts : rows in table `build_artifact` with a `build_id` that doesn't exists anymore. 
```sql
SELECT * FROM build_artifact
WHERE build_id NOT IN (SELECT DISTINCT(id) FROM build);
/* this returns non-empty result */
```

pierrotsmnrd:
Fix #476 : Delete dangling build artifacts before migration.

### DIFF
--- a/conda-store-server/conda_store_server/alembic/versions/16f65805dc8f_split_conda_package_into_conda_package_.py
+++ b/conda-store-server/conda_store_server/alembic/versions/16f65805dc8f_split_conda_package_into_conda_package_.py
@@ -225,6 +225,16 @@ def upgrade():
             "build_environment_id_fkey", "environment", ["environment_id"], ["id"]
         )
 
+    # One case of data inconsistency has been identified, that makes the migration fail :
+    #   Dangling build artifacts : rows in table `build_artifact` with a `build_id`` that doesn't exists anymore.
+    #   see issue https://github.com/Quansight/conda-store/issues/476
+    # Fix : delete dangling artifacts
+    op.execute(
+        """ DELETE FROM build_artifact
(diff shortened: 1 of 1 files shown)


costrouc (approved):
@pierrotsmnrd

costrouc:
@pierrotsmnrd going forward I was wondering what adjustments would be needed in the orm model?

pierrotsmnrd:
@costrouc my understanding is that nothing else is necessary.

[This line](https://github.com/Quansight/conda-store/blob/f8a0ba10e0aa3567f84d2bdf0363034975773e7e/conda-store-server/conda_store_server/orm.py#L147) ensures the on-delete cascade from build to build_artifacts and was added 2 years ago.

So :
- Either there's something we haven't spot yet in the codebase that leads to dangling build_artifacts (but what ?)
- Or the case we encountered is the result of an old install of conda-store, where builds has been deleted before the on-delete cascade was added. Considering this identified case is using an old Postgres 11.14, this hypothesis could make sense. Or maybe they tinkered with their data and we don't know.

costrouc (approved):

PR #480, opened 2023-04-28
Pin SQLAlchemy <=1 .4.47

pierrotsmnrd:
Pinning SQLAlchemy to avoid solving with incompatible version >= 2


costrouc (approved):
Thanks @pierrotsmnrd!

PR #481, opened 2023-05-03
[WIP] POC of on-delete cascade on build➞artifacts

pierrotsmnrd:
Now, deleting a build deletes its artifacts.

Additionally, ensures Alembic detects some changes in the ORM and generates a migration step for these changes ( compare_type and compare_server_default )

This PR needs to be completed to setup the on-delete policy correctly all over the DB.


pierrotsmnrd:
Some tests are failing because the migration doesn't work on SQLite. This will be fixed for a fully-scoped PR beyond the simple POC.

costrouc:
Rebased on `main`

costrouc:
@pierrotsmnrd is this still a WIP? I think due to recent changes you'll need to change the alembic parent/child version sha.

PR #500, opened 2023-07-12
Add trailing slash to avoid redirects

### LINKED ISSUE #499: aiohttp drops authorization header on redirect
https://github.com/aio-libs/aiohttp/issues/5943#issuecomment-897637093

Because of this the conda-store cli will be unable to make authenticated requests if the slash is not appended, as it will not handle the redirect.

ClaytonAstrom:
Resolves #499

### DIFF
--- a/conda-store/tests/authenticated-token-tests.sh
+++ b/conda-store/tests/authenticated-token-tests.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export CONDA_STORE_URL=http://localhost:5000/conda-store
+export CONDA_STORE_AUTH=basic
+export CONDA_STORE_USERNAME=username
+export CONDA_STORE_PASSWORD=password
+
+conda-store --version
+conda-store info
+
+export CONDA_STORE_TOKEN=$(conda-store token)
+export CONDA_STORE_AUTH=token
+unset CONDA_STORE_USERNAME
+unset CONDA_STORE_PASSWORD
+echo $CONDA_STORE_TOKEN
+
+conda-store info
+conda-store list build
+echo "waiting for build 1 to finish"
+conda-store wait 1
+conda-store wait filesystem/python-flask-env
+conda-store wait filesystem/python-flask-env:1
+conda-store wait filesystem/python-flask-env:1 --artifact=archive
+conda-store list namespace
+conda-store list build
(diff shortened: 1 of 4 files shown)


ClaytonAstrom:
Newer versions (>=1.9.0) of yarl don't seem to agree with `/ ""`.  If you prefer the later versions of yarl I can change the syntax to `str(yarl.URL("foo") / "bar") + "/"`.  But I felt that was breaking convention too much

trallard:
Thanks for your contribution @ClaytonAstrom 🙌🏽 I will request a review from @costrouc to keep this moving

conda-store/tests/authenticated-token-tests.sh
costrouc, line 1: @ClaytonAstrom was this file meant to be checked in?

conda-store/pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 dependencies = [
    "rich",
    "click",
-   "yarl",
+   "yarl<1.9.0",
```
costrouc, line 31: I'd prefer not to have to pin to a lower version. Is the `str(...) + '/'` the only viable alternative?

costrouc (changes requested):

conda-store/pyproject.toml
costrouc: I will say I've tried several approaches and it's been pretty disappointing :frowning_face: . I'd prefer if we used the str(...) + '/' method just to avoid having to pin yarl. Could you create a function to wrap these like `def _ensure_slash(...)` or something along those lines.

ClaytonAstrom: Ok, updated @costrouc - possibly not the most elegant solution but let me know if I can change anything to suit this better.

conda-store/tests/authenticated-token-tests.sh
costrouc, line 1: Does this file need to be checked in?

ClaytonAstrom: It didn't look like there were any tests using the token explicitly, so I copied the basic auth and unset user/password after generating a token.  If this test isn't helpful I'll ditch it

costrouc (approved):
LGTM. Thanks for the contribution @ClaytonAstrom!

Issue #491, opened 2023-07-01
Improvements to the namespace Role Mappings


costrouc:
# Motivation

For authentication currently the user to role and permissions has been managed in traitlets. Via a few important attributes

```
c.RBACAuthorizationBackend.role_mappings: Dict[str, set[str]] = {
            "viewer": {
                schema.Permissions.ENVIRONMENT_READ,
                schema.Permissions.NAMESPACE_READ,
             }
 }       
c.RBACAuthorizationBackend.unauthenticated_role_bindings: Dict[str, set[str]] =  {
        "default/*": {"viewer"},
 }
c.RBACAuthorizationBackend.authenticated_role_bindings: Dict[str, set[str]] = {
            "default/*": {"viewer"},
            "filesystem/*": {"viewer"},
}

class MyAuthentication(Authentication):
       async def authenticate(self, request: Request):
            return schema.AuthenticationToken(
                primary_namespace=username,
                role_bindings={
                    "*/*": ["admin"],
                },
            )

c.CondaStoreServer.authentication_class = MyAuthentication
```

Developers were responsible for creating a function `authenticate(...)` which handled the user to role mapping. With this configuration it was possible to conda-store to not know anything about the users.

# Proposal

I propose adding one new database table to conda-store. We leverage the fact that the username will match a namespace name which is created on the backend.

```python
# orm.py

# update namespace table
class Namespace(Base):
     ...
     metadata = Column(JSON) # would be nice to rely on metadata so that we don't enforce the attributes that are on a user

class NamespaceRoleMapping(Base):
    id = Column(Integer, primary_key=True)
    namespace_id = ...
    entity = Column(Unicode(255), nullable=False)     # arn e.g. <namespace>/<name> like `quansight-*/*` or `quansight-devops/*`
    role = Column(Unicode(255), nullable=False)  # e.g. viewer
```

Then add a method to `RBACAuthorizationBackend` 

```python
def get_namespace_role_bindings(namespace: str):
      # query database for all entity role mappings
      return {entity: roles} # will match form in `schema.AuthenticationToken.role_bindings`
```

Additionally there should be rest api methods and permissions for CRUD. I am least sure about this part.
    - need to add a `schema.Permissions` `namespace::update` permission which will allow updating the metadata
    - namespace update rest api method that uses the namespace update 
[...]

PR #508, opened 2023-07-24
API : update a namespace's metadata and role mappings

pierrotsmnrd:
This PR adds a new API endpoint to update a namespace's role mappings and metadata.

the `PUT /conda-store/api/v1/namespace/{namespace}/` endpoint takes a JSON payload to update the metadata, the role mappings, or both.

- update the metadata : 
```json
{
  "metadata": {"my_key":"my_value"}
}
``` 

- update the role mappings : 
```json
{
  "role_mappings": {
    "admin/admin": ["admin", "viewer", "developer"],
    "admin/*": ["viewer"],
    "demo/*": ["developer"]
  }
}
```

- update both :
```json
{
  "metadata": {"my_key":"my_value"},
  "role_mappings": {
    "admin/admin": ["admin"],
    "admin/*": ["viewer"],
    "demo/*": ["developer"]
  }
}
```

@costrouc tell me if that works for you or if you'd see another approach or pattern.

According to the original issue #491 : 
> - namespace update rest api method that uses the namespace update
> - create, list, delete for adding namespace role mappings. I think you should only be able to edit namespace role mappings if you have namespace::update permissions. If you have namespace::read it should be sufficient to view members.

We now need to add an endpoint to list the role mappings.

### DIFF
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Any, Optional
+from typing import List, Dict, Any, Optional, Union
 import re
 
 from sqlalchemy import func, null, or_, distinct
@@ -50,8 +50,8 @@ def create_namespace(db, name: str):
 def update_namespace(
     db,
     name: str,
-    metadata_: Optional[Dict] | None = None,
-    role_mappings: Optional[Dict[str, List[str]]] | None = None,
+    metadata_: Union[Optional[Dict], None] = None,
+    role_mappings: Union[Optional[Dict[str, List[str]]], None] = None,
 ):
 
     namespace = get_namespace(db, name)
--- a/conda-store-server/conda_store_server/api.py
+++ b/conda-store-server/conda_store_server/api.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Optional
 import re
 
 from sqlalchemy import func, null, or_, distinct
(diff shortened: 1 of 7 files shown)


conda-store-server/conda_store_server/schema.py
```diff
@@ -34,6 +34,7 @@ class Permissions(enum.Enum):
     BUILD_DELETE = "build::delete"
     NAMESPACE_CREATE = "namespace::create"
     NAMESPACE_READ = "namespace::read"
+    NAMESPACE_UPDATE = "namespace::update"
```
costrouc, line 37: @pierrotsmnrd this PR looks great to me. No real issues. My questions is on the granularity of permissions:

Should we have one permission `namespace::update` or should we have two? To me having the ability to change a namespaces metadata and giving users permissions to a namespace are at much different permission levels.

So for example:
  - `namespace::update`
  - `namespace-role-mapping::create`
  - `namespace-role-mapping::delete`

I'm not entirely sure on the naming.

conda-store-server/conda_store_server/api.py
```diff
@@ -45,6 +47,47 @@ def create_namespace(db, name: str):
     return namespace
 
 
+def update_namespace(
+    db,
+    name: str,
+    metadata_: Optional[Dict] | None = None,
```
costrouc, line 53: Let's not use `|` since this isn't backwards compatible with 3.9. Ue `typing.Union`

costrouc (approved):
LGTM. Thanks @pierrotsmnrd

Issue #514, opened 2023-08-01
TASK - Post org migration tasks


trallard:
Now that we have migrated the repository to the conda-incubator organisation there are several follow-up tasks to complete:

```[tasklist]
### Tasks
- [ ] https://github.com/conda-incubator/conda-store/issues/532
- [x] Update `.github/workflows` as needed - https://github.com/conda-incubator/conda-store/pull/525
- [x] Update issue and PR templates in the repository - https://github.com/conda-incubator/conda-store/pull/521
- [x] Ensure the repo points to the conda CoC - https://github.com/conda-incubator/conda-store/pull/518
- [x] Update repository URL in the docs and README - https://github.com/conda-incubator/conda-store/pull/518
- [x] Fix broken links in README - https://github.com/conda-incubator/conda-store/pull/518
- [ ] https://github.com/conda-incubator/conda-store/issues/519
```

PR #518, opened 2023-08-02
Update repository URL

steff456:
This PR is part of https://github.com/conda-incubator/conda-store/issues/514

This PR,

- [x] Changes url that where pointing to `Quansight` org and replace it with `conda-incubator` org.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,12 +57,12 @@ Learn more, including how to install, use, and contribute to conda-store in our 
 
 ## Related repositories
 
-- We are working on a new UI for conda-store at: [`Quansight/conda-store-ui`](https://github.com/Quansight/conda-store-ui), and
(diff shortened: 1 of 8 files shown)


costrouc (changes requested):
@steff456 thanks for the PR. Should we also update the links in the `CHANGELOG.md`? Also I see another file https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/alembic/versions/16f65805dc8f_split_conda_package_into_conda_package_.py#L230 which mentions the Quansight repo.

cc @trallard what are your thoughts about changing bits in the CHANGELOG.md?

steff456:
I wasn't sure if we were going to modify the changelog, so I'll wait for a reply before doing changes in that file

costrouc (approved):
I would prefer that the changelog doesn't change since it can help to show the history of the project. @trallard we can apply these changes later if needed. LGTM. I'll merge.

Issue #507, opened 2023-07-20
META - Support for Windows and MacOS (local)


trallard:
## Background

Currently, `conda-store` is limited to Linux-based systems use.
To better serve our community, we'd need to ensure it is Windows and MacOS compatible.

## Tasks

```[tasklist]
### General
- [ ] https://github.com/conda-incubator/conda-store/issues/648
```

```[tasklist]
### Windows
- [ ] https://github.com/conda-incubator/conda-store/pull/640
- [ ] https://github.com/conda-incubator/conda-store/issues/611
- [ ] https://github.com/conda-incubator/conda-store/issues/588
```

```[tasklist]
### macOS
- [ ] https://github.com/conda-incubator/conda-store/pull/537
- [ ] https://github.com/conda-incubator/conda-store/pull/550
```

costrouc:
@asmeurer releated to https://github.com/conda-incubator/conda-store/issues/513. And should be included with that work.

asmeurer:
I was able to run conda-store-server with `--standalone` on macos. You have to comment out `conda-docker` from the environment-dev.yml and add `python-docker` to build the development environment.

When I tried creating an environment, I get

```pytb
Traceback (most recent call last):
  File "/Users/aaronmeurer/Documents/conda-store/conda-store-server/conda_store_server/build.py", line 278, in build_conda_environment
    fetch_and_extract_packages(conda_store, pathlib.Path(tmp_lock_filename))
  File "/Users/aaronmeurer/Documents/conda-store/conda-store-server/conda_store_server/build.py", line 197, in fetch_and_extract_packages
    with filelock.FileLock(f"{str(filepath)}.lock"):
  File "/Users/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/filelock/_api.py", line 255, in __enter__
    self.acquire()
  File "/Users/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/filelock/_api.py", line 213, in acquire
    self._acquire()
  File "/Users/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/filelock/_unix.py", line 37, in _acquire
    fd = os.open(self.lock_file, open_flags, self._context.mode)
[...]

PR #524, opened 2023-08-04
Add python-docker as an explicit dependency in environment.yml

asmeurer:
It is also an implicit dependency of conda-docker, but conda-docker may become an optional dependency for Mac (#507).


costrouc (approved):
Yeah agreed this should be included since it is directly imported.

jaimergp:
I think this is currently missing in `pyproject.toml` too.

PR #529, opened 2023-08-07
Using scoped session to ensure single connection per thread

costrouc:
Using scoped sessions avoids creating excess connections.


costrouc:
Closing since this clearly isn't the correct fix.

PR #531, opened 2023-08-08
Adding build canceling (only works with compatible celery brokers redis and rabbitmq)

costrouc:
Closes https://github.com/conda-incubator/conda-store/issues/306

Build canceling exposed at `PUT /api/v1/build/<build-id>/cancel/` and added a new permissions `build::cancel` which is available by default for `admin` and `developer` roles.

### DIFF
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -958,7 +958,10 @@ async def api_put_build_cancel(
     )
 
     if conda_store.celery_app.control.inspect().ping() is None:
-        raise HTTPException(status_code=409, detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation")
+        raise HTTPException(
+            status_code=409,
+            detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation",
+        )
 
     conda_store.celery_app.control.revoke(
         [
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -943,50 +943,50 @@ async def api_put_build_cancel(
     build_id: int,
     request: Request,
     conda_store=Depends(dependencies.get_conda_store),
-    db: Session = Depends(dependencies.get_db),
     auth=Depends(dependencies.get_auth),
 ):
-    build = api.get_build(db, build_id)
-    if build is None:
-        raise HTTPException(status_code=404, detail="build id does not exist")
+    with conda_store.get_db() as db:
+        build = api.get_build(db, build_id)
+        if build is None:
+            raise HTTPException(status_code=404, detail="build id does not exist")
 
-    auth.authorize_request(
-        request,
-        f"{build.environment.namespace.name}/{build.environment.name}",
-        {Permissions.BUILD_CANCEL},
-        require=True,
-    )
-
-    if conda_store.celery_app.control.inspect().ping() is None:
-        raise HTTPException(
-            status_code=409,
-            detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation",
+        auth.authorize_request(
+            request,
+            f"{build.environment.namespace.name}/{build.environment.name}",
+            {Permissions.BUILD_CANCEL},
+            require=True,
         )
 
-    conda_store.celery_app.control.revoke(
-        [
-            f"build-{build_id}-conda-env-export",
-            f"build-{build_id}-conda-pack",
-            f"build-{build_id}-conda-docker",
-            f"build-{build_id}-environment",
-        ],
-        terminate=True,
-        signal="SIGTERM",
-    )
+        if conda_store.celery_app.control.inspect().ping() is None:
+            raise HTTPException(
+                status_code=409,
+                detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation",
+            )
 
-    from conda_store_server.worker import tasks
+        conda_store.celery_app.control.revoke(
+            [
+                f"build-{build_id}-conda-env-export",
+                f"build-{build_id}-conda-pack",
+                f"build-{build_id}-conda-docker",
+                f"build-{build_id}-environment",
+            ],
+            terminate=True,
+            signal="SIGTERM",
+        )
 
-    tasks.task_cleanup_builds.si(
-        build_ids=[build_id],
-        reason=f"""
-build {build_id} marked as FAILED due to being canceled from the REST API
-""",
-    ).apply_async(countdown=5)
+        from conda_store_server.worker import tasks
 
-    return {
-        "status": "ok",
-        "message": f"build {build_id} canceled",
-    }
+        tasks.task_cleanup_builds.si(
+            build_ids=[build_id],
+            reason=f"""
+    build {build_id} marked as FAILED due to being canceled from the REST API
+    """,
+        ).apply_async(countdown=5)
+
+        return {
+            "status": "ok",
+            "message": f"build {build_id} canceled",
+        }
 
 
 @router_api.delete(
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -957,6 +957,9 @@ async def api_put_build_cancel(
         require=True,
     )
 
+    if conda_store.celery_app.control.inspect().ping() is None:
+        raise HTTPException(status_code=409, detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation")
+
     conda_store.celery_app.control.revoke(
         [
             f"build-{build_id}-conda-env-export",
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -967,7 +967,7 @@ async def api_put_build_cancel(
             [
                 f"build-{build_id}-conda-env-export",
                 f"build-{build_id}-conda-pack",
-                f"build-{build_id}-conda-docker",
+                f"build-{build_id}-docker",
                 f"build-{build_id}-environment",
             ],
             terminate=True,
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -974,6 +974,12 @@ async def api_put_build_cancel(
         signal="SIGTERM",
     )
 
+    from conda_store_server.worker import tasks
+
+    tasks.task_cleanup_builds.si(build_ids=[build_id], reason=f"""
+build {build_id} marked as FAILED due to being canceled from the REST API
+""").apply_async(countdown=5)
+
     return {
         "status": "ok",
         "message": f"build {build_id} canceled",
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -980,6 +980,8 @@ async def api_put_build_cancel(
 
         from conda_store_server.worker import tasks
 
+        # Waits 5 seconds to ensure enough time for the task to actually be
+        # canceled
         tasks.task_cleanup_builds.si(
             build_ids=[build_id],
             reason=f"""
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -935,6 +935,45 @@ async def api_put_build(
         }
 
 
+@router_api.put(
+    "/build/{build_id}/cancel/",
+    response_model=schema.APIAckResponse,
+)
+async def api_put_build_cancel(
+    build_id: int,
+    request: Request,
+    conda_store=Depends(dependencies.get_conda_store),
+    db: Session = Depends(dependencies.get_db),
+    auth=Depends(dependencies.get_auth),
+):
+    build = api.get_build(db, build_id)
+    if build is None:
+        raise HTTPException(status_code=404, detail="build id does not exist")
+
+    auth.authorize_request(
+        request,
+        f"{build.environment.namespace.name}/{build.environment.name}",
+        {Permissions.BUILD_CANCEL},
+        require=True,
+    )
+
+    conda_store.celery_app.control.revoke(
+        [
+            f"build-{build_id}-conda-env-export",
+            f"build-{build_id}-conda-pack",
+            f"build-{build_id}-conda-docker",
+            f"build-{build_id}-environment",
+        ],
+        terminate=True,
+        signal="SIGTERM",
+    )
+
+    return {
+        "status": "ok",
+        "message": f"build {build_id} canceled",
+    }
+
+
 @router_api.delete(
     "/build/{build_id}/",
     response_model=schema.APIAckResponse,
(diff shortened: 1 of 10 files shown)


costrouc:
TODO: this PR requires a test:
 -  that demonstrates starting a build 
 - canceling via the api endpoint
 - seeing that the status of the build -> `FAILED`
 - and the reason in logs of the FAILURE is due to build being canceled

trallard:
@costrouc  are you adding the test to this PR or do you need someone to take over this?

costrouc:
@trallard I could definitely use some help and I think this would be a good experience for writing tests in conda-store.

trallard:
Ok will assign @anirrudh  to wrap up the rest of the PR - had a look and indeed seems we are missing tests (overall) so need to prioritise that

costrouc:
> had a look and indeed seems we are missing tests (overall) so need to prioritise that

Overview of the tests that we currently have:
  - unit tests over authentication, authorization, all build actions, celery worker tasks, traitlets, db actions `conda-store-server/tests/test_*.py`
  - integration tests (create env, update, delete etc), (and full api stability tests) `tests/test_*.py`
  - playwright browser test full flow `tests/test_playwright.py`
  - integration tests for conda-store client  `conda-store/tests`

anirrudh:
@costrouc in that case, I will be writing tests based on the comment that you made earlier in this thread.

anirrudh:
@costrouc Are `build_id` values still identifiers of a certain build of an environment but global (i.e. the build ids for a given env can be 1 and 3, skipping 2 as another env in the same namespace is being built but has been assigned the build_id value 2)? IIRC this is how this works.

anirrudh:
@costrouc Are `build_id` values still identifiers of a certain build of an environment but global (i.e. the build ids for a given env can be 1 and 3, skipping 2 as another env in the same namespace is being built but has been assigned the build_id value 2)? IIRC this is how this works.

tests/test_api.py
```diff
@@ -730,3 +732,45 @@ def test_delete_build_auth(testclient):
...
+    testclient.login()
+    response = testclient.put(f"api/v1/build/{build_id}")
+    r = schema.APIPostSpecification.parse_obj(response.json())
+    assert r.status == schema.APIStatus.OK
+
+    new_build_id = r.data.build_id
+
+    # delay to ensure the build kicks off
+    building = False
+    while building is False:
```
costrouc, line 749: This while loop will never end is some cases. Need to have this as a timeout with check interval.

tests/test_api.py
```diff
@@ -730,3 +732,45 @@ def test_delete_build_auth(testclient):
...
+    # The new build should have kicked off, so now we will request to cancel it
+    response = testclient.put(f"api/v1/build/{new_build_id}/cancel")
+
+    r = schema.APIResponse.parse_obj(response.json())
+
+    assert r.status == schema.APIStatus.OK
+    assert "canceled" in r.message
+
+    # delay to ensure the build is marked as failed
+    time.sleep(120)
```
costrouc, line 767: Should take less than 10s

costrouc (commented):
I can't request changes since I started the PR :smile:. But just two tweaks in my mind. Also good to see tests are passing! So this tells me build canceling is indeed working.

tests/test_api.py
anirrudh: set max to 10

anirrudh: Tests fail when set to something low :(

tests/test_api.py
anirrudh: Done, using `time`.  Good catch.

costrouc:
LGTM. Thanks @anirrudh !

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -688,6 +688,60 @@ async def api_put_build(
...
+            detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation",
+        )
+
+    conda_store.celery_app.control.revoke(
+        [
+            f"build-{build_id}-conda-env-export",
+            f"build-{build_id}-conda-pack",
+            f"build-{build_id}-conda-docker",
+            f"build-{build_id}-environment",
+        ],
```
pierrotsmnrd, line 725: I see what's done here. We're cancelling all the possibly existing tasks for the given build_id. 

There's a problem with `f"build-{build_id}-conda-docker"` : The name of the task is actually `f"build-{build.id}-docker"`, based on [app.py](https://github.com/conda-incubator/conda-store/blob/67364dcd65f2b50fbea3be5be7dc4fbaadf2b7ab/conda-store-server/conda_store_server/app.py#L674)

Overall, I think we could provide a more resilient approach, where we don't have to manually maintain a list of tasks here in `api.py`, to match with `app.py`

Short term solution : fix the wrong task name above
Longer term solution : Use constants or something so the list above is generated automatically and identical to the strings used when creating the tasks, removing the need to maintain it.

pierrotsmnrd (changes requested):
At least one requested change to properly cancel "docker" tasks. The rest is improvements / suggestions.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -688,6 +688,60 @@ async def api_put_build(
...
+        build_ids=[build_id],
+        reason=f"""
+build {build_id} marked as FAILED due to being canceled from the REST API
+""",
+    ).apply_async(countdown=5)
+
+    return {
+        "status": "ok",
+        "message": f"build {build_id} canceled",
+    }
```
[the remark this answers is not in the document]
trallard: Agree. There needs to be a clear status distinction between builds that were cancelled by the user and failed ones.

conda-store-server/conda_store_server/server/views/api.py
costrouc: Totally. This makes sense. 

I'm wondering where the best place to store this is. Currently the most natural place feels like `orm.py`. 

Also related to this I am not sure if all those task names needed to be included? This is something that confuses me in celery. The "root" task name is `build-{build_id}-environment`

conda-store-server/conda_store_server/server/views/api.py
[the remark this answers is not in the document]
costrouc: I agree with this and checking celery and redis queues they both have a concept of a canceled task (revoked and canceled).

This will require a lot more file/changes. Since CANCELED is adding a new final state. Also will require an alembic migration to account for this new state.

Should we wrap this into this PR?

conda-store-server/conda_store_server/server/views/api.py
costrouc: @pierrotsmnrd I don't think that celery considers the children subtasks to "exist" and be cancelable/revokable. This is what needs to be checked.

costrouc: @anirrudh can you check this?

anirrudh: Yeah, I will check this in our case. I'll have some thoughts in a bit.

anirrudh: @costrouc I've experimented a bit, and this seems to be the case. Celery does not list these as part of tasks that are collected in the overall task id. I tried, and we would have to use something like `celery.chord` at task orchestration time, and then pull this information out later from the object.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -688,6 +688,60 @@ async def api_put_build(
...
 
+@router_api.put(
+    "/build/{build_id}/cancel/",
+    response_model=schema.APIAckResponse,
+)
+async def api_put_build_cancel(
+    build_id: int,
+    request: Request,
+    conda_store=Depends(dependencies.get_conda_store),
+    db: Session = Depends(dependencies.get_db),
```
nkaretnikov, line 699: Will need to be fixed after rebase. We don't use deps for db anymore.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -688,6 +688,60 @@ async def api_put_build(
...
+        f"{build.environment.namespace.name}/{build.environment.name}",
+        {Permissions.BUILD_CANCEL},
+        require=True,
+    )
+
+    if conda_store.celery_app.control.inspect().ping() is None:
+        raise HTTPException(
+            status_code=409,
+            detail="conda-store celery broker does not support task cancelation. Use redis or rabbitmq message queues. See docs for a more detailed explanation",
+        )
```
nkaretnikov, line 717: The said docs are here (took me a bit to find): https://docs.celeryq.dev/en/latest/userguide/workers.html#worker-persistent-revokes
> Note that remote control commands must be working for revokes to work. Remote control commands are only supported by the RabbitMQ (amqp) and Redis at this point.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -688,6 +688,60 @@ async def api_put_build(
...
+            f"build-{build_id}-conda-env-export",
+            f"build-{build_id}-conda-pack",
+            f"build-{build_id}-conda-docker",
+            f"build-{build_id}-environment",
+        ],
+        terminate=True,
+        signal="SIGTERM",
+    )
+
+    from conda_store_server.worker import tasks
```
nkaretnikov, line 730: Move to the start of the file unless this conflicts with something.

conda-store-server/conda_store_server/server/views/api.py
nkaretnikov: - I agree that task ids shouldn't be hardcoded. I'll make this change.
- I'll see if it's possible to cancel subtasks using `build-{build.id}` and/or whether this task id should be canceled as well.

conda-store-server/conda_store_server/server/views/api.py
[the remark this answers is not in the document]
nkaretnikov: I was confused by the above thread a bit since the message already says "canceled". The requested change here is to pass a yet-to-be-added CANCELED (made up name) marker/param to `task_cleanup_builds`, to make use of it in `build_cleanup`, which currently just calls `set_build_failed` regardless of the reason.

Will be done in a separate PR.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
 
     # r = schema.APIResponse.parse_obj(response.json())
     # assert r.status == schema.APIStatus.ERROR
+
+
+def test_api_cancel_build(testclient):
```
nkaretnikov, line 737: This test needs an unauth version to check that auth is working, see `test_delete_build_unauth`.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
...
+    start = time.time()
+    while time.time() - start < build_timeout:
+        response = testclient.get(f"api/v1/build/{new_build_id}")
+        response.raise_for_status()
+
+        r = schema.APIGetBuild.parse_obj(response.json())
+        if r.data.status == schema.BuildStatus.BUILDING.value:
+            building = True
+            break
+        time.sleep(30)
```
nkaretnikov, line 759: This should probably be lowered to 5. No need to wait half a minute on failure.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
...
+
+    # delay to ensure the build kicks off
+    build_timeout = 180
+    building = False
+    start = time.time()
+    while time.time() - start < build_timeout:
+        response = testclient.get(f"api/v1/build/{new_build_id}")
+        response.raise_for_status()
+
+        r = schema.APIGetBuild.parse_obj(response.json())
```
nkaretnikov, line 755: This should also have the standard `assert r.status == schema.APIStatus.OK` to fail early if there are problems. It will fail due to building not being True, but that'd be harder to debug later.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
...
+
+    assert building is True
+
+    # The new build should have kicked off, so now we will request to cancel it
+    response = testclient.put(f"api/v1/build/{new_build_id}/cancel")
+
+    r = schema.APIResponse.parse_obj(response.json())
+
+    assert r.status == schema.APIStatus.OK
+    assert "canceled" in r.message
```
nkaretnikov, line 769: Formatting: remove newlines here, would be easier to read near `r`.
Check the full error message in the assert.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
...
+        r = schema.APIGetBuild.parse_obj(response.json())
+        if r.data.status == schema.BuildStatus.BUILDING.value:
+            building = True
+            break
+        time.sleep(30)
+
+    assert building is True
+
+    # The new build should have kicked off, so now we will request to cancel it
+    response = testclient.put(f"api/v1/build/{new_build_id}/cancel")
```
nkaretnikov, line 764: `response.raise_for_status()` missing after this, like everywhere else.

tests/test_api.py
```diff
@@ -730,3 +732,50 @@ def test_delete_build_auth(testclient):
...
+    time.sleep(30)
+
+    # Ensure status is Failed
+    response = testclient.get(f"api/v1/build/{new_build_id}")
+    response.raise_for_status()
+
+    r = schema.APIGetBuild.parse_obj(response.json())
+    assert r.status == schema.APIStatus.OK
+    assert r.data.id == new_build_id
+    assert r.data.status == schema.BuildStatus.FAILED.value
```
nkaretnikov, line 781: Note to self: when another PR adding `BuildStatus.CANCELED` is done, this should fail and will need to be updated.

nkaretnikov:
I'm taking over this PR.

Reviewed the code (without running it) as of this commit. Looks fine overall. Will fix things outlined in the comments.

nkaretnikov:
Note to self: rebased on a CI fix from #689, might need to rebase later in case this causes problems

tests/test_api.py
```diff
@@ -983,7 +983,7 @@ def test_api_cancel_build(testclient):
         if r.data.status == schema.BuildStatus.BUILDING.value:
             building = True
             break
-        time.sleep(30)
+        time.sleep(5)
```
nkaretnikov, line 986: I've seen the build status being set to COMPLETED because this was first checking too early and then too late.

nkaretnikov:
This is ready for review.  I've tested manually as well.

Note that this commit might need to be removed: https://github.com/conda-incubator/conda-store/pull/531#issuecomment-1848796250. Waiting for someone to approve, then will merge myself.

nkaretnikov:
@costrouc PTAL, I cannot assign you as a reviewer since it's a PR started by you.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -935,6 +935,64 @@ async def api_put_build(
...
+        )
+
+        from conda_store_server.worker import tasks
+
+        tasks.task_cleanup_builds.si(
+            build_ids=[build_id],
+            reason=f"""
+    build {build_id} marked as FAILED due to being canceled from the REST API
+    """,
+        ).apply_async(countdown=5)
```
costrouc, line 988: Could maybe use a comment to explain that we have to wait 5 seconds to ensure enough time for the task to actually be canceled.

costrouc:
Only one "comment" about making it clear why we set a countdown of 5 seconds otherwise looks great to me! I can't approve this since I was the one that started the PR

trallard (approved):
Approving and once that comment is added you should be able to merge @nkaretnikov

PR #539, opened 2023-08-16
Don't run conda-docker on Mac

asmeurer:
conda-docker doesn't work on Mac, and making it work is not straightforward, so for now we just disable it when we aren't on Linux.

With this and #537, basic Mac support seems to work with --standalone (#507). I'm not sure what else I should be testing, but I can create environments and they appear to be functional.

### DIFF
--- a/conda-store-server/conda_store_server/action/generate_conda_docker.py
+++ b/conda-store-server/conda_store_server/action/generate_conda_docker.py
@@ -1,12 +1,5 @@
+import sys
 import pathlib
-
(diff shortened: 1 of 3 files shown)


costrouc:
@asmeurer awesome! Glad to hear that OSX looks to be working. 

Two things I think that are next:
 - adding tests to the CI that we could run just to ensure this functionality is working using osx runners
 - working on ensuring that windows is supported

costrouc (approved):
LGTM

PR #541, opened 2023-08-17
Delete subdomain keys on logout

### LINKED ISSUE #512: logout method does not clear subdomain cookie
Currently when a user logs out when `Authentication.cookie_domain` is set. The cookie is not properly deleted.

To fix the https://github.com/conda-incubator/conda-store/blob/b9063dd003e04b892b6fb0c57c6e31bd4e773f62/conda-store-server/conda_store_server/server/auth.py#L452 the `domain` key needs to be set.

anirrudh:
Fixes #512

## Description
Deletes cookies properly on logout. Code has been formatted with Black. 

This pull request:

- a
- b
- c

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/server/auth.py
+++ b/conda-store-server/conda_store_server/server/auth.py
@@ -445,7 +445,7 @@ form.addEventListener('submit', loginHandler);
     def post_logout_method(self, request: Request, next: Optional[str] = None):
         redirect_url = next or request.url_for("ui_list_environments")
         response = RedirectResponse(redirect_url, status_code=303)
-        response.set_cookie(self.cookie_name, "", domain="", expires=0)
+        response.set_cookie(self.cookie_name, "", domain=self.cookie_domain, expires=0)
         return response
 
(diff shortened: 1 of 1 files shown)


conda-store-server/conda_store_server/server/auth.py
```diff
@@ -447,7 +445,7 @@ async def post_login_method(
     def post_logout_method(self, request: Request, next: Optional[str] = None):
         redirect_url = next or request.url_for("ui_list_environments")
         response = RedirectResponse(redirect_url, status_code=303)
-        response.set_cookie(self.cookie_name, "", expires=0)
+        response.set_cookie(self.cookie_name, "", domain="", expires=0)
```
costrouc, line 448: Sorry I gave the wrong advice it should be `domain=self.cookie_domain`

costrouc (changes requested):
Gave the wrong advice in the issue. Need to set the domain to the traitlets variable.

costrouc (approved):

PR #546, opened 2023-08-25
Better handling of pydantic error msgs

### LINKED ISSUE #535: Improve pydantic validation check error reporting on yaml specification
### Describe the issue

I have a conda yaml that passes a yaml linter for valid yaml spec, but when I try to use conda-store-ui to create an env with it, I get an error about validation. 

```
channels:
  - conda-forge
  - pytorch
dependencies:
  - python >= 3.8
  - typing-extensions >= 4.5,!= 4.6
  - numpy >= 1.17

  - pillow >= 9.4.0
  - pytorch >= 1.7,!=1.12.0
  - torchvision >= 0.14.1
  - datasets >= 2.7.1
  - transformers >= 4.25.1
  - timm >= 0.6.12
  - torchmetrics >= 0.11,< 1.0
  - tqdm >= 4.62.1

  - # test
  - jupytext >= 1.14.0
  - numpydoc >= 1.5.0
  - pyright >= 1.1.280
  - pytest >= 7.0.0
  - hypothesis >= 6.61.0
  - pytest-mock >= 3.10.0

  - pip:
      - huggingface-hub >= 0.11.1
      - torcheval >= 0.0.6
```

The error is:
```
1 validation error for CondaSpecification dependencies -> 11 none is not an allowed value (type=type_error.none.not_allowed)
```

### Expected behavior

I expect conda-store to give me an explanation of the validation error - 
* What part of my yaml spec failed?
* What line number (of my original spec) failed?
* Or was it something in the description that failed?

Copy and paste the above into a new environment yaml spec on nebari conda store ui. 

### Output

image

_No response_

### Anything else?

_No response_
```[tasklist]
### Tasks
- [ ] Pydantic schema - Identify improvements needed
- [ ] Improve thrown error message from PyDantic validation
```

pierrotsmnrd:
Fixes #535 (work in progress)

## Description
The purpose of this PR is to present more user-friendly error messages, in case of invalid YAML specification when trying to create an env.

## Pull request checklist

- [ ] ~Did you update the documentaion (if required)?~

## Additional information
I consider this PR as a work in progress.
- the unique error case currently handled, might have its human-friendly message improved
- we might handle there other cases than the one described in the original issues
- Does the "translation"  `pydantic error ➞ human` need to be done somewhere else ? 
- Also : unit tests


costrouc:
Personally there are two things that have bugged me about pydantic error messages:
 - context
 - human readable messages

For context it would be nice to see the offending lines/location

For human readable messages I imagine it might be worth putting some work into the schema `CondaSpecification` and adding some additional validation to ensure that the messages are generally better.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -565,7 +565,15 @@ async def api_post_specification(
...
-        raise HTTPException(status_code=400, detail=str(e))
+
+        error_type = e.errors()[0]["type"]
+        human_readable_error = str(e)
+        if error_type == "type_error.none.not_allowed":
+            human_readable_error = (
+                "Invalid YAML. A forbidden `None` value has been encountered. "
+            )
+
+        raise HTTPException(status_code=400, detail=human_readable_error)
```
asmeurer, line 576: A more scalable way to handle this would be to have a subclass of the HTTPException class that handles the translation in the constructor.

pierrotsmnrd:
I have followed @costrouc 's recommendations.

Now, the validation is done in the `CondaSpecification` class, and **all the validation errors are raised** (we use to display only one)

For instance, for a spec where the name of the env is missing, and an empty dependency exists, we have the two errors like this : 
Capture d’écran 2023-09-08 à 08 12 48

We need to add a fix in the UI to handle the `\n` in between the two errors.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -564,6 +564,8 @@ async def api_post_specification(
         specification = schema.CondaSpecification.parse_obj(specification)
     except yaml.error.YAMLError:
         raise HTTPException(status_code=400, detail="Unable to parse. Invalid YAML")
+    except ValueError as e:
```
asmeurer, line 567: You should use a custom exception here. Something else could raise a ValueError which wouldn't have a list in its args, and that probably shouldn't be converted to an HTTPException anyway.

pierrotsmnrd: Done.

trallard (commented):
This looks good to me, @asmeurer  since you had reviewed this before, can you have another look at this and approve if applicable?

asmeurer:
This looks fine. We'll probably want to refactor this if we start translating a lot more error messages, but with just a couple the way it's written now is fine.

Issue #513, opened 2023-07-31
Revisit the implementation of `--standalone` as well as the assumptions made


trallard:
One of our goals is to improve the local experience or story of the project. To do so, we need to:

```[tasklist]
### Tasks
- [x] Revisit the current `--standalone` flag and its assumptions
- [x] Test the current implementation
- [ ] Identify improvements or fixes needed
- [ ] Make improvements / add features as needed
- [ ] Document this feature
```

trallard:
More than likely it is using SQLite database for the local implementation

asmeurer:
I was able to use --standalone to create an environment, but if I try to create a subsequent environment I get the error

```pytb
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 408, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/uvicorn/middleware/message_logger.py", line 84, in __call__
    raise exc from None
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/uvicorn/middleware/message_logger.py", line 80, in __call__
    await self.app(scope, inner_receive, inner_send)
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/fastapi/applications.py", line 289, in __call__
    await super().__call__(scope, receive, send)
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/aaronmeurer/anaconda3/envs/conda-store-server-dev/lib/python3.10/site-packages/starlette/middleware/base.py", line 108, in __call__
    response = await self.dispatch_func(request, call_next)
[...]

PR #549, opened 2023-08-25
Make tests use a temporary directory for conda environments

asmeurer:
Previously they used the current conda prefix, but this is dangerous (there's a risk a broken test could break the dev environment), and the current conda prefix could contain things that break the tests.

This currently works by parameterizing in the fixture itself (currently just for two basic envs, one without pip packages and one with). But in the future we can do this parameterization on a per-test level with indirect parameterization. See
https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization

Currently, the test_generate_conda_export() test fails here. The reason is that it creates an export which cannot actually be read by the CondaSpecification object.

The reason for this is that the "name" of the environment that gets generated is just the same as the "prefix". This fails a validation check, because the name is guarded by a regular expression that disallows paths.

However, this appears to be a bug in conda-store, or maybe even an architectural issue. The "name" of a conda environment, as defined by "conda env export" at least, is only really defined if that environment is in the envs directory. I checked, and even env export yamls generated by conda-store itself (in standalone mode) end up having the prefix for the name, rather than the supposed actual environment name.

At best, it means that this test is just wrong and it shouldn't actually check the generated YAML file by trying to load it into a CondaSpecification.

I don't know if conda actually stores the environment "name" anywhere in the environment. It seems to me that this data should be stored separately by conda-store, especially since the environment path for environments created by conda-store is not actually the same as the environment name (the path contains a hash).

This is related to issue https://github.com/conda-incubator/conda-store/issues/513.

## Description

This pull request:

- a
- b
- c

## Pull request checklist

### DIFF
--- a/conda-store-server/tests/conftest.py
+++ b/conda-store-server/tests/conftest.py
@@ -6,7 +6,7 @@ import pytest
 import yaml
 from fastapi.testclient import TestClient
 
-from conda_store_server import app, schema, dbutil, utils, testing, api
+from conda_store_server import action, app, schema, dbutil, utils, testing, api
 from conda_store_server.server import app as server_app
 
 
@@ -24,7 +24,11 @@ def conda_store_config(tmp_path):
     filename = pathlib.Path(tmp_path) / "database.sqlite"
 
     with utils.chdir(tmp_path):
-        yield Config(CondaStore=dict(database_url=f"sqlite:///{filename}?check_same_thread=False"))
+        yield Config(
+            CondaStore=dict(
+                database_url=f"sqlite:///{filename}?check_same_thread=False"
+            )
+        )
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


conda-store-server/tests/conftest.py
```diff
@@ -173,6 +177,29 @@ def simple_conda_lock():
         return yaml.safe_load(f)
 
 
-@pytest.fixture
-def current_prefix():
-    return pathlib.Path(os.environ["CONDA_PREFIX"])
+@pytest.fixture(
```
costrouc, line 180: Love it. This is a great way I think to be able to test more complete environment in the future.

costrouc (approved):
> I don't know if conda actually stores the environment "name" anywhere in the environment. It seems to me that this data should be stored separately by conda-store, especially since the environment path for environments created by conda-store is not actually the same as the environment name (the path contains a hash).

Yeah like you said conda doesn't store the `name` anywhere.

It takes the `envs_dir` director and any folder in there is assumed to be the `name`.  It is possible for there to be collisions in the environment names. 

I chose that the `CondaSpecification` does not store the `name` since I felt this was redundant (already stored in the db) and did not influence the solve.

I'm fine with this PR as is

asmeurer:
Just to be clear, `test_generate_conda_export` fails in its current state. Should I just [remove the line](https://github.com/conda-incubator/conda-store/pull/549/files#diff-8a31c7c4d818471c7260ef5e11763e8c56ce3bf4a1302308041c8b9361ed40bfR101) that tries to load the environment file into a CondaSpecification? I could also to other workarounds like hotfixing the `name` before loading it, which would still test that everything else loads fine.

costrouc:
> At best, it means that this test is just wrong and it shouldn't actually check the generated YAML file by trying to load it into a CondaSpecification.

@asmeurer I think you are completely right.

> I could also to other workarounds like hotfixing the name before loading it, which would still test that everything else loads fine.

Yeah I'd go this route just so that we can be sure that export is still generating valid environment yamls.

asmeurer:
OK, I've made that change.

costrouc (approved):
LGTM!

PR #554, opened 2023-08-30
Store the state and database files in ~/.conda-store by default

asmeurer:
Still some issues that need to be worked out here. Also will need #549 to be merged and rebased first. 

This work is part of #513.

## Description

This pull request:

- a
- b
- c

## Pull request checklist


costrouc:
Much saner default considering before how it would just write to your current directory. Agree this is needed.

asmeurer:
Also need to double check that there aren't any places that were relying on the default values without configuring it (e.g., in the Docker setup).

asmeurer:
Hmm, previously I was having issues where the database file was not being initialized properly, but now I'm not. I'm not sure if something else changed in the interim that changed this, or if I'm just not doing the same steps I was before. It would be helpful if someone else can test this (just run `conda-store-server --standalone`, or alternatively run `conda-store-server` and `conda-store-worker` separately, and see if it appears to function and that there aren't any errors printed to the console).

trallard:
@nkaretnikov to test

conda-store-server/conda_store_server/app.py
```diff
@@ -200,7 +201,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     database_url = Unicode(
-        "sqlite:///conda-store.sqlite",
+        "sqlite:///" + str(PosixPath.home()) + "/.conda-store/conda-store.sqlite",
```
nkaretnikov, line 204: Need to remove `Posix`, it doesn't work on Windows:

```diff
-        "sqlite:///" + str(PosixPath.home()) + "/.conda-store/conda-store.sqlite",
+        "sqlite:///" + str(Path.home()) + "/.conda-store/conda-store.sqlite",
```

Tested on Windows: after rebasing on https://github.com/conda-incubator/conda-store/pull/559 and fixing the Posix thing, the server started (`python -m conda_store_server.server`) and stored state files in the home dir.
Screen Shot 2023-10-01 at 19 42 39

conda-store-server/conda_store_server/app.py
```diff
@@ -87,7 +88,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     store_directory = Unicode(
-        "conda-store-state",
+        str(Path.home() / ".conda-store" / "conda-store-state"),
```
nkaretnikov, line 91: Add a variable for `Path.home() / ".conda-store"` and use it everywhere. For example: `CONDA_STORE_DIR`.

nkaretnikov: The location of `CONDA_STORE_DIR` needs to be printed when the server is started.

Ideally: you can even print something like "Found existing conda-store.sqlite in `<dirname>`" or "Creating conda-store.sqlite in `<dirname>`". Same for the rest of top-level state files/dirs.

Otherwise, we'll run into problems when reproducing issues. People will have no idea that some state was created somewhere. Up until now, it hasn't been a problem because files are more visible in the current project dir, since it's tracked by git.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -179,6 +179,8 @@ def initialize(self, *args, **kwargs):
 
         self.conda_store = CondaStore(parent=self, log=self.log)
 
+        self.conda_store.ensure_directories()
```
nkaretnikov, line 182: Note for other reviewers: this runs:
```py
os.makedirs(self.store_directory, exist_ok=True)
```
because store_directory might not exist.

conda-store-server/tests/conftest.py
```diff
@@ -24,9 +24,13 @@ def conda_store_config(tmp_path, request):
 
     filename = pathlib.Path(tmp_path) / "database.sqlite"
```
nkaretnikov, line 25: Not essential, but needs `".conda-store"` here, to be consistent with how we run this outside of tests.

conda-store-server/conda_store_server/app.py
nkaretnikov: Tested on Linux: works without changes.

```
% ls -1
conda-store.sqlite
conda-store-state
% pwd
/home/nkaretnikov/.conda-store
```

nkaretnikov: Verified that this env builds on Linux and is available after restart (with the same state dir):

```
channels:
- conda-forge
dependencies:
- python
- pip:
  - nothing
- ipykernel
- pytest
- requests
description: ''
name: test-env
prefix: null
variables: null
```

Server was started via `python -m conda_store_server.server --standalone`.

asmeurer: Are you sure about that? PositPath is used for URLs because URLs always use forward slashes, even on Windows.

asmeurer: See for instance https://github.com/conda-incubator/conda-store/pull/559/commits/80d9ea863aa0acbc4f3420449ee97ad1e1f65254

nkaretnikov: > Are you sure about that? PosixPath is used for URLs because URLs always use forward slashes, even on Windows.

Have you tested on Windows? When I ran it on Windows, it raised an error.

asmeurer:
Adding another note here: I think it would be a good idea to rename `~/.conda-store/conda-store-state` to just `~/.conda-store/state`, which is not only cleaner and less redundant, but should help with #588 as well.

conda-store-server/conda_store_server/app.py
asmeurer: Apparently on Windows you have to use pathlib.PurePosixPath instead of pathlib.PosixPath

asmeurer:
Tested this on Windows by manually merged it with #559

asmeurer:
I've addressed the review comments. 

One thing worth being aware of here is that if anyone is running conda-store without configuring database_url or store_directory but just relying on the defaults, their setup will break. All the example configs do configure it, so hopefully that isn't happening.

nkaretnikov:
> One thing worth being aware of here is that if anyone is running conda-store without configuring database_url or store_directory but just relying on the defaults, their setup will break. All the example configs do configure it, so hopefully that isn't happening.

Wait, I think it wasn't an issue before, no? Why is this required now? Like, I'm pretty sure I used to be able to just run with `--standalone` without passing the config option.

asmeurer:
The point is it will break *existing* installations if they were relying on the default configuration, because the defaults have changed.

asmeurer:
Hmm, the test failure is because tests conda_store object from the fixture isn't being used for some reason.

nkaretnikov:
@asmeurer The test failure looks legit. You also need to rebase.

```
________________ test_conda_store_register_environment_workflow ________________

...

ERROR    celery.app.trace:trace.py:270 Task task_update_storage_metrics[43b3f586-1076-4ec6-8185-6b17c6e0bda6] raised unexpected: FileNotFoundError(2, 'No such file or directory')
  File "/usr/share/miniconda/envs/conda-store-server-dev/lib/python3.10/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/runner/work/conda-store/conda-store/conda-store-server/conda_store_server/worker/tasks.py", line 75, in task_update_storage_metrics
    conda_store.configuration(db).update_storage_metrics(
  File "/home/runner/work/conda-store/conda-store/conda-store-server/conda_store_server/orm.py", line 663, in update_storage_metrics
    disk_usage = shutil.disk_usage(store_directory)
  File "/usr/share/miniconda/envs/conda-store-server-dev/lib/python3.10/shutil.py", line 1280, in disk_usage
    st = os.statvfs(path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/runner/.conda-store/state'
```

nkaretnikov:
Taking over this one. See https://github.com/conda-incubator/conda-store/pull/639.

PR #559, opened 2023-09-05
Windows support

asmeurer:
See https://github.com/conda-incubator/conda-store/issues/507

Nothing is working yet, but I plan to push changes up here as I make them. 

## Description

This pull request:

- a
- b
- c

## Pull request checklist


asmeurer:
As far as I can tell, with the latest changes, the server process runs just fine on Windows. However, the worker process does not (https://github.com/conda-incubator/conda-store/issues/507#issuecomment-1709185995), so nothing substantial actually works. However, I am able to login and to click on the various UI elements, and everything that doesn't require a worker seems to function (there's no actual indication that there are no workers, but this is a separate issue that has been mentioned elsewhere).

asmeurer:
One issue I am running into is that installs are failing with errors like

```
```

This is a known issue in conda. See https://github.com/conda/conda/issues/7203. Windows, by default, only allows paths up to 260 characters. See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

It is possible to remove this limitation in Windows 10 and later if you set a registry value https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later. However, this (presumably) requires administrator privileges so I don't know if it's a viable thing to require. 

The exact length of the path depends on the specific conda package. However, it's pretty obvious that the issue is the long UUID generated by conda-store: `99108419ad0fd922fdeff9bbc434b58d41f68e3f923a83f6a7ab19568463bc82-20230915-201619-848782-1-test`. That's 90 characters (plus the 4 character environment name). Shortening or removing this UUID from the path would solve the issue for this specific package. One could also store the environment closer to the root (see also #554). However, I'm not really sure what the longest "real world" path in a conda package is. The conda issue mentions that this has been fixed upstream in many conda-forge packages with longer paths, but I don't know just how long of a path is considered "OK" for a Windows conda package, but it's still something that could pop up even if we fix both of those things. At that point, that would probably be considered an upstream issue with conda and/or the specific package, but it's worth being aware of this problem.

asmeurer:
At any rate, I think I am going to work on finishing up #554 as that should at least allow me to install some test packages so I can work on the remaining Windows issues (I guess I could also just set that registry value on my system, but that PR needs to be finished anyway).

asmeurer:
Based on this comment, https://github.com/conda/conda/issues/7203#issuecomment-1722123864, if we want to try to make the path no longer than a typical conda install default, the part of the path between `C:\Users\Username` and the individual package files (like `\Lib\site-packages\...`) should be no longer than `\AppData\Local\conda\conda\pkgs`, which is 31 characters. But `\.conda-store\conda-store-state`, as proposed in #554, is already 31 characters. 

Of course, most packages don't have really long paths, but some do.

trallard:
@nkaretnikov to test

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,11 +50,48 @@ def chdir(directory: pathlib.Path):
...
+            have.add(st.st_ino)
+            apparent_total_bytes += st.st_size
+        for d in dirnames:
+            dp = os.path.join(dirpath, d)
+            if os.path.islink(dp):
+                apparent_total_bytes += os.lstat(dp).st_size
+
+    # Round up
+    n_blocks = (apparent_total_bytes + 511) // 512
+    return n_blocks
```
nkaretnikov, line 85: Will do more tests tomorrow. So far: tested on Linux against `du -sb`, returns wrong results for `/bin`. For `/bin/ls` it seems to work. I'll write a test.

nkaretnikov:
@trallard How urgent is this needed? Personally, I'd rather get Windows working on CI first, see: https://github.com/conda-incubator/conda-store/issues/603

@asmeurer Do conda-store-server tests pass for you with `pytest` locally? I'm testing on Win11 ARM, for me it hangs at around 50%. Could be an issue with my setup, though:

Screen Shot 2023-09-30 at 18 34 02

@asmeurer What's known to work? What doesn't work?

I'll take a closer look tomorrow. For now, I can say that running the main UI at `/` is not found for me. `/admin` loads.

Screen Shot 2023-09-30 at 18 42 36

asmeurer:
Quite a few tests fail for me too. 

I forgot to mention that 0.0.0.0 doesn't work, but 127.0.0.1 or localhost does. I think it has something to do with the default way IPs are configured on Windows.

conda-store-server/conda_store_server/utils.py
nkaretnikov: OK, this is a bit tricky.

First, we incorrectly calculate file size on macOS with `du`. There's no built-in set of options that can give us the equivalent of `-sb` AFAICT. We could install `du` from GNU coreutils, which is what is used on Linux, then we'd get the same size as our script. _But I'd suggest we just use our Python code to calculate the size on all systems._ However, it would still not match what Finder is showing, but I think it's close enough to simply ignore. Different tools calculate this differently and it's not hugely important to users unless we're in the same ballpark.

I also had to remove the round up calculation from the original script. Also, simplified it.

I also haven't tested this script with symlink/filesystem loops. Is this a safe assumption to make that we won't run into this where this is supposed to be used?

My updated code + some tests: https://gist.github.com/nkaretnikov/1a66b90a74fa805f1022e90252e54c87
Note: uncomment the TemporaryDirectory line again and remove two lines after it, which were added for testing on Windows:
```
    # with tempfile.TemporaryDirectory() as dir:
        dir = "c:\\tmpdir"
        os.makedirs(dir)
```

On Windows, I've also attempted to test against [du from SysInternals](https://learn.microsoft.com/en-us/sysinternals/downloads/du), but it also shows different info. It hasn't been updated in a while, so I just ignored that.

In general, I suggest we rely on what OS built-in "File info" tools return to debug this.

On Linux, the updated script matches the du command we're using.

Windows (updated Python du matches native File info size, Sysinternals du shows different info):
Screen Shot 2023-10-01 at 12 55 55

macOS (updated Python du roughly matches native File info, built-in du returns different size):
Screen Shot 2023-10-01 at 13 12 57

macOS (updated Python du matches du from GNU coreutils, installed via brew):
Screen Shot 2023-10-01 at 13 19 23

Note: creating symlinks on Windows requires having Developer Mode on:
```
Settings > Privacy & Security > For Developers and turn Developer Mode to on
```

conda-store-server/conda_store_server/worker/app.py
```diff
@@ -71,9 +71,18 @@ def start(self):
         argv = [
             "worker",
             "--loglevel=INFO",
-            "--beat",
         ]
 
+        # The default Celery pool requires this on Windows. See
+        # https://stackoverflow.com/questions/37255548/how-to-run-celery-on-windows
+        if sys.platform == "win32":
+            os.environ.setdefault('FORKED_BY_MULTIPROCESSING', '1')
```
nkaretnikov, line 79: It makes me worried that we rely on a library that has no official Windows support.

It's worth discussing the level of Windows support we expect to provide ourselves and possibly consider alternatives to celery.

I did attempt to test this by running my concurrency test from https://github.com/conda-incubator/conda-store/pull/567/commits/3fc0e1425fae17f1bae8a1f35421e9c77f553059. I also parameterized it by all pools from `get_available_pool_names`. All of them failed. This could be for a number of reasons, but it's not worth investigating at the moment due to 50% of the testsuite failing for me. Also: pytest celery fixture might require additional tweaks to get working.

Before we start talking about celery or concurrency, I suggest we get the rest of the testsuite working on Windows.

nkaretnikov:
Just tried building a simple environment, also didn't work.

Screen Shot 2023-10-01 at 19 10 29

Env:
```
channels:
  - conda-forge
dependencies:
  - python
  - pip:
    - nothing
  - ipykernel
  # tests
  - pytest

  - requests  # for app

# other fields
description: ''
name: test-env    # required
prefix: null
variables: null
```

nkaretnikov:
@asmeurer @trallard This concludes my review for now, here's the summary (you can skip the rest of the comments if you just want a high-level overview):
- 50% of tests fail for me (could be an issue on my side, but it's hard to tell without CI)
- couldn't build a simple env
- celery has no official Windows support and my test for it didn't work, but it could be for other reasons
- Python du implementation needs fixing (see gist in a thread above), ideally: we use it on all OSes to be consistent
- I suggest to enable Windows on CI and start fixing the testsuite to establish a working baseline.

nkaretnikov:
Realized I was running the server incorrectly to test env building, so a quick update.

Building a simple env failed again, but for a different reason:
```
 (ERROR:root:[WinError 206] The filename or extension is too long: 'C:\\dev\\conda-store\\conda-store-server\\conda-store-state\\default\\a54148c1d1d435985989ea232bc982f0eb877991f6c70c4de1390dd76e7ef564-20231001-183326-961615-1-test-env\\Lib\\site-packages\\debugpy\\_vendored\\pydevd\\_pydevd_bundle\\pydevd_concurrency_analyser')
```

Env:
```
channels:
- conda-forge
dependencies:
- python
- pip:
  - nothing
- ipykernel
- pytest
- requests
description: ''
name: test-env
prefix: null
variables: null
```

Backtrace in Build log (via web UI):

```
ERROR:root:
ERROR:root:[WinError 206] The filename or extension is too long: 'C:\\dev\\conda-store\\conda-store-server\\conda-store-state\\default\\a54148c1d1d435985989ea232bc982f0eb877991f6c70c4de1390dd76e7ef564-20231001-183326-961615-1-test-env\\Lib\\site-packages\\debugpy\\_vendored\\pydevd\\_pydevd_bundle\\pydevd_concurrency_analyser'
ERROR:root:()
ERROR:root:
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\conda_lock\__main__.py", line 4, in <module>
    main()
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\click\decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\conda_lock\conda_lock.py", line 1489, in install
    install_func(file=lockfile)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\conda_lock\conda_lock.py", line 225, in do_conda_install
    _conda(
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\conda_lock\invoke_conda.py", line 129, in _invoke_conda
    raise subprocess.CalledProcessError(
conda_lock._vendor.poetry.utils._compat.CalledProcessError: Command '['C:\\dev\\miniconda3\\Scripts\\conda.exe', 'create', '--file', 'C:\\Users\\nikita\\AppData\\Local\\Temp\\tmpwi3q46m3', '--yes', '--prefix', 'C:\\dev\\conda-store\\conda-store-server\\conda-store-state\\default\\a54148c1d1d435985989ea232bc982f0eb877991f6c70c4de1390dd76e7ef564-20231001-183326-961615-1-test-env']' returned non-zero exit status 1.
```

Command to start the server: `python -m conda_store_server.server --standalone`.

conda-store-server/conda_store_server/utils.py
asmeurer: When I tested it on my Mac, du and this function gave the same numbers. I'm not sure if the subtle differences of blocks matter much (it matters if there are sparse or compressed files, but I doubt those would show up). The main thing is that we treat hard links correctly. 

Note that in my tests, this function is significantly slower than `du`, so we should consider whether it's worth using it over du.

asmeurer:
The "filename or extension is too long" error is described here https://github.com/conda-incubator/conda-store/pull/559#issuecomment-1721915623. For now, you'll have to set the registry key to enable long file paths to test this on Windows. This is tracked by #588

asmeurer:
@costrouc it turns out the `total_seconds` fix *does* fix all those 403 test failures we were seeing on Windows [`877f816` (#559)](https://github.com/conda-incubator/conda-store/pull/559/commits/877f8167b4251fae48caa814668c9ef74b7f9f18). It wasn't working the other day because apparently calling `int()` on `total_seconds()` is also important (I guess a float value for the `max_age` still messes it up somehow).

conda-store-server/conda_store_server/worker/app.py
asmeurer: Most of the other pools fail because they aren't actually concurrent on Windows, and one of the tasks blocks all the others (I think the watch task, but I'm not completely sure). But even if that weren't the case, conda-store would be very slow without concurrent tasks due to some slow tasks like updating channels.

conda-store-server/conda_store_server/utils.py
asmeurer: I tested it on my conda-store-state with a couple of environments:

```
$ du -sAB1 conda-store-state/
1111657	conda-store-state/
```

```
>>> from conda_store_server.utils import du
>>> du('conda-store-state')
1111657
```

That's on my Mac. I also can confirm that some of the environments do share files via hard-links:

```
$ ls -i conda-store-state/admin/b82cde5b5489ceffd8a8589ebd73f20a9f4836260b18295f49e057c441b235dc-20231005-213031-558981-2-test2/lib/libsqlite3.0.dylib
256048028 conda-store-state/admin/b82cde5b5489ceffd8a8589ebd73f20a9f4836260b18295f49e057c441b235dc-20231005-213031-558981-2-test2/lib/libsqlite3.0.dylib
$ ls -i conda-store-state/admin/99108419ad0fd922fdeff9bbc434b58d41f68e3f923a83f6a7ab19568463bc82-20231005-211948-613237-1-test/lib/libsqlite3.0.dylib
256048028 conda-store-state/admin/99108419ad0fd922fdeff9bbc434b58d41f68e3f923a83f6a7ab19568463bc82-20231005-211948-613237-1-test/lib/libsqlite3.0.dylib
```

I don't remember if I tested it on Linux, so it's possible there's a discrepancy there.

asmeurer: OK, I just checked on Linux, and apparently the `du -sb` command we are using is 512 times bigger than the macOS `du -sAB1` (despite the man pages implying that they would be the same) and this `du()` function. I guess we should fix them both to account for that, since the code is already written assuming bytes instead of blocks.

asmeurer: I pushed a fix for this. Note that the numbers will be off a little bit (by less than 512) on Mac because `du` rounds up to the nearest multiple of 512 when converting bytes to blocks.

However, we still need to figure out something for this `du()` function. I just tested it against my main `~/anaconda` directory and I had to exit out of it after several minutes. I think it might be accidentally quadratic. 

There is a Windows `du` command someone at https://learn.microsoft.com/en-us/sysinternals/downloads/du. Maybe we should just package that as a conda package so we can just use it directly (or maybe it already is packaged?).

asmeurer: Actually even regular `du -sAB1 ~/anaconda` is slow on my machine. Maybe it's just too hard to deal with that many hard links. 

But I just noticed that the way conda-store is using this, it gathers the stats for each prefix as it is created. I need to double check it, but I think it actually isn't accounting for hard links across environments correctly anyway.

asmeurer: Honestly, I'd appreciate some guidance from @costrouc or someone else on how disk_usage is actually used in conda-store before I feel comfortable with the du stuff here.

nkaretnikov: > There is a Windows du command someone at https://learn.microsoft.com/en-us/sysinternals/downloads/du. Maybe we should just package that as a conda package so we can just use it directly (or maybe it already is packaged?).

I've already tried it. See above where I post my du test results. Sysinternals du printed wrong results for me, compared to Windows file explorer. The output also differs from Linux/macOS du.

asmeurer: Oh I didn't notice you tested it already.

Regarding what the Finder returns, I would be curious to know exactly why the discrepancy exists. You can access it programmatically with AppleScript

```
osascript -e 'tell application "Finder" to get physical size of folder "Macintosh HD:Users:aaronmeurer:Documents:conda-store:conda-store-server:conda-store-state"'
6.6859008E+8
```

(I highly recommend using an LLM to help you write AppleScript)

asmeurer:
What's causing this error on CI?

```pytb
Run hatch env run -e dev lint
Creating environment: dev
Installing project in development mode
  File "C:\Miniconda\envs\conda-store-server-dev\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Miniconda\envs\conda-store-server-dev\lib\runpy.py", line [8](https://github.com/conda-incubator/conda-store/actions/runs/6462286985/job/17543617411?pr=559#step:6:9)6, in _run_code
    exec(code, run_globals)
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\__main__.py", line 22, in <module>
    from pip._internal.cli.main import main as _main
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\cli\main.py", line 10, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\cli\autocompletion.py", line 10, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\cli\main_parser.py", line [9](https://github.com/conda-incubator/conda-store/actions/runs/6462286985/job/17543617411?pr=559#step:6:10), in <module>
    from pip._internal.build_env import get_runnable_pip
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\build_env.py", line 19, in <module>
    from pip._internal.cli.spinners import open_spinner
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\cli\spinners.py", line 9, in <module>
    from pip._internal.utils.logging import get_indentation
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\utils\logging.py", line 29, in <module>
    from pip._internal.utils.misc import ensure_dir
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\utils\misc.py", line 41, in <module>
    from pip._internal.locations import get_major_minor_version
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\locations\__init__.py", line 14, in <module>
    from . import _sysconfig
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\locations\_sysconfig.py", line [11](https://github.com/conda-incubator/conda-store/actions/runs/6462286985/job/17543617411?pr=559#step:6:12), in <module>
    from .base import change_root, get_major_minor_version, is_osx_framework
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\locations\base.py", line 9, in <module>
    from pip._internal.utils import appdirs
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_internal\utils\appdirs.py", line [13](https://github.com/conda-incubator/conda-store/actions/runs/6462286985/job/17543617411?pr=559#step:6:14), in <module>
    from pip._vendor import platformdirs as _appdirs
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_vendor\platformdirs\__init__.py", line 46, in <module>
    PlatformDirs = _set_platform_dir_class()  #: Currently active platform
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_vendor\platformdirs\__init__.py", line [26](https://github.com/conda-incubator/conda-store/actions/runs/6462286985/job/17543617411?pr=559#step:6:27), in _set_platform_dir_class
    from pip._vendor.platformdirs.windows import Windows as Result
  File "C:\Users\runneradmin\AppData\Local\hatch\env\virtual\conda-store-server\6yWfAnoa\dev\lib\site-packages\pip\_vendor\platformdirs\windows.py", line 4, in <module>
    import ctypes
  File "C:\Miniconda\envs\conda-store-server-dev\lib\ctypes\__init__.py", line 8, in <module>
    from _ctypes import Union, Structure, Array
ImportError: DLL load failed while importing _ctypes: The specified module could not be found.
```

asmeurer:
This test failure:

```
self = <json.decoder.JSONDecoder object at 0x000001662FA408B0>
s = 'menuinst called from non-root env C:\\Miniconda\\envs\\conda-store-server-dev\n{\n  "channels": [\n    "main",\n    "...\AppData\\\\Local\\\\Temp\\\\pytest-of-runneradmin\\\\pytest-0\\\\test_generate_conda_export_con1\\\\test-prefix"\n}\n'
idx = 0

    def raw_decode(self, s, idx=0):
        """Decode a JSON document from ``s`` (a ``str`` beginning with
        a JSON document) and return a 2-tuple of the Python
        representation and the index in ``s`` where the document ended.
    
        This can be used to decode a JSON document from a string that may
        have extraneous data at the end.
    
        """
        try:
            obj, end = self.scan_once(s, idx)
        except StopIteration as err:
>           raise JSONDecodeError("Expecting value", s, err.value) from None
E           json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Comes from penuinst printing stuff at import time, which is pretty annoying https://github.com/conda/menuinst/blob/97d578806d04596e7a2391bea84088621cd97d9c/menuinst/_legacy/win32.py#L87-L88 (CC @jaimergp).

But is there a good reason why `context.run` pipes stderr to stdout (https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/action/base.py#L52)? That would necessarily break any `--json` command if anything is printed to stderr.

asmeurer:
OK, CI is passing now. This is ready for another review. I do still want to

- [ ] Add some tests for the `du()` function. Based on the discussion at the standup, the disk usage isn't that important so I don't think we need to stress too hard about it being super accurate for now. All other changes here should already be covered by the existing tests, but if anyone sees something else that a test should be added for let me know.
- [ ] Update documentation. Particularly about the long paths issue, but also some dev instructions probably need some slight tweaks (e.g., replacing `0.0.0.0` with `127.0.0.1`).

asmeurer:
> e.g., replacing 0.0.0.0 with 127.0.0.1

Actually 0.0.0.0 doesn't seem to be documented anywhere. It's used in the configuration, which is [correct](https://www.uvicorn.org/settings/#socket-binding), but I must have picked up using it in the browser to actually connect to conda-store from the someone else, not the docs. Everywhere in the docs says "localhost:5000".

At any rate, just to be clear, 0.0.0.0 does not redirect to localhost on Windows. This is a Linux/BSD thing, which is kind of a hack on their part, since 0.0.0.0 is not actually supposed to be meaningful as a destination address. See https://superuser.com/questions/1755409/is-the-definition-of-0-0-0-0-consistent-across-os and https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0 for more details. It would be a good habit to stop using http://0.0.0.0:5000 to connect to conda-store and always use http://localhost:5000 instead, even on Linux/macOS.

For now, I've just made conda-store-server --standalone print

```
Starting standalone conda-store server at http://localhost:5000
```

I still need to check if any other documentation needs updating for anything else here.

nkaretnikov:
My 2c: the reason 0.0.0.0 is used is to be able to connect to it from any address. From your link: "In the context of servers, 0.0.0.0 means "all IPv4 addresses on the local machine". If a host has two IP addresses, 192.168.1.1 and 10.1.2.1, and a server running on the host listens on 0.0.0.0, it will be reachable at both of those IPs.".

asmeurer:
Yes, 0.0.0.0 is correct as the host address for this reason. But it is not correct to use it as the address to connect to conda-store itself, i.e., you should not type "0.0.0.0" into your web browser address bar.

asmeurer:
This is ready for final review.

asmeurer:
I guess uvicorn itself prints:

```
INFO:     Uvicorn running on http://0.0.0.0:5000 (Press CTRL+C to quit)
```

(although sometimes it doesn't. I'm not sure why).

I don't know if there's an easy way to change that. We would probably want to upstream a fix for it to uvicorn.

nkaretnikov:
> I don't know if there's an easy way to change that. We would probably want to upstream a fix for it to uvicorn.

I don't think there's anything here to fix. It informs how the server was started, which is on 0.0.0.0. It has nothing to do with how the users should connect to it.

.github/workflows/tests.yaml
```diff
@@ -42,6 +40,24 @@ jobs:
           environment-file: conda-store-server/environment-macos-dev.yaml
           auto-activate-base: false
 
+      - name: Set up Python (Windows)
+        if: matrix.os == 'windows-latest'
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+          activate-environment: conda-store-server-dev
+          environment-file: conda-store-server/environment-windows-dev.yaml
+          auto-activate-base: false
```
nkaretnikov, line 49: Heads up: https://github.com/conda-incubator/conda-store/pull/634 switches to miniforge. This will need to be updated to match the rest of the jobs once that PR lands.

.github/workflows/tests.yaml
```diff
@@ -52,7 +68,7 @@ jobs:
 
       - name: Unit Tests
         run: |
-          pytest tests
+          pytest -vvv tests
```
nkaretnikov, line 71: Does this actually provide useful information? Or can it be removed?

conda-store-server/conda_store_server/action/base.py
```diff
@@ -38,20 +38,23 @@ class ActionContext:
...
             *args,
             **kwargs,
             stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
+            stderr=subprocess.STDOUT if redirect_stderr else subprocess.PIPE,
             encoding="utf-8",
         )
         self.stdout.write(result.stdout)
+        if not redirect_stderr:
+            self.stderr.write(result.stderr)
```
nkaretnikov, line 59: Why are these changes being introduced as part of this PR?

conda-store-server/conda_store_server/schema.py
```diff
@@ -9,6 +9,8 @@
 from conda_store_server import conda_utils, utils
 from pydantic import BaseModel, Field, ValidationError, constr, validator
 
+ON_WIN = sys.platform.startswith("win")
```
nkaretnikov, line 12: This is already declared in another file. Can we move that definition somewhere where it makes the most sense and import everywhere? Ideally, this should be done in a way that minimizes potential import cycles.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -335,6 +336,10 @@ def start(self):
 
         # start worker if in standalone mode
         if self.standalone:
+            address = "localhost" if self.address == "0.0.0.0" else self.address
+            print(
+                f"Starting standalone conda-store server at http://{address}:{self.port}"
+            )
```
nkaretnikov, line 342: I don't like this. If we're using `self.address` we should print `self.address`. I understand this is trying to be helpful, but it'll only make issues harder to debug later. I'd suggest to remove this entirely.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -436,7 +436,9 @@ async def post_login_method(
             samesite="strict",
             domain=self.cookie_domain,
             # set cookie to expire at same time as jwt
-            max_age=(authentication_token.exp - datetime.datetime.utcnow()).seconds,
+            max_age=int(
+                (authentication_token.exp - datetime.datetime.utcnow()).total_seconds()
+            ),
```
nkaretnikov, line 441: Why was this change necessary?

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,52 @@ def chdir(directory: pathlib.Path):
...
         cmd = ["du", "-sb", str(path)]
+    else:
+        return str(du(path))
 
-    return subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    output = subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    if sys.platform == "darwin":
+        # mac du does not have the -b option to return bytes
+        output = str(int(output) * 512)
+    return output
```
nkaretnikov, line 98: This "darwin" part needs to be moved to a different PR since it has nothing to do with Windows.

conda-store-server/environment-windows-dev.yaml
```diff
@@ -0,0 +1,56 @@
+name: conda-store-server-dev
+channels:
+  - conda-forge
+  - Microsoft
```
nkaretnikov, line 4: This PR https://github.com/conda-incubator/conda-store/pull/634 made changes to yaml files, please update the Windows one to match those changes once 634 lands.

conda-store-server/hatch_build.py
```diff
@@ -61,10 +61,12 @@ def initialize(self, version: str, build_data: Dict[str, Any]) -> None:
             # main.js to enable easy configuration see
             # conda_store_server/server/templates/conda-store-ui.html
             # for global variable set
-            with (source_directory / "main.js").open("r") as source_f:
+            with (source_directory / "main.js").open("r", encoding="utf-8") as source_f:
```
nkaretnikov, line 64: Why was the encoding change necessary?

conda-store-server/tests/test_actions.py
```diff
@@ -155,6 +164,7 @@ def test_remove_conda_prefix(tmp_path, simple_conda_lock):
     assert not conda_prefix.exists()
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="permissions are not supported on Windows")
```
nkaretnikov, line 167: Can we use the newly defined ON_WIN here?

docs/contributing.md
```diff
@@ -66,18 +66,20 @@ Install the following dependencies before developing on conda-store.
...
 Running `conda-store`. `--standalone` mode launched celery as a
-subprocess of the web server.
+subprocess of the web server. Run
 
-python -m conda_store_server.server --standalone tests/assets/conda_store_standalone_config.py
-
-````
+```
+python -m conda_store_server.server --standalone
+```
```
nkaretnikov, line 80: The old command was wrong because it didn't pass the config option. But why does the new example not tell how to run with a config? See https://github.com/conda-incubator/conda-store/issues/594

docs/contributing.md
```diff
@@ -99,6 +101,8 @@ To build the documentation install the development environment via
 Conda.
 
 ```shell
+# replace this with environment-macos-dev.yaml or environment-windows-dev.yaml
```
nkaretnikov, line 104: I'd prefer this to just talk about Windows. And do all mac changes in another PR.

docs/contributing.md
asmeurer: The idea behind --standalone is that the config defaults are suitable for standalone mode. So most developers shouldn't need to pass a config, unless they specifically need to change a given config value.

conda-store-server/hatch_build.py
asmeurer: It was using Windows cp1252 encoding for me, which was causing errors because these Javascript files contain emojis.

conda-store-server/conda_store_server/utils.py
asmeurer: Why?

conda-store-server/conda_store_server/server/auth.py
asmeurer: I explained this in the commit message [`877f816` (#559)](https://github.com/conda-incubator/conda-store/pull/559/commits/877f8167b4251fae48caa814668c9ef74b7f9f18)

conda-store-server/conda_store_server/action/base.py
asmeurer: https://github.com/conda-incubator/conda-store/pull/559#issuecomment-1756298903

conda-store-server/conda_store_server/schema.py
```diff
@@ -194,20 +196,20 @@ class Settings(BaseModel):
         metadata={"global": True},
     )
 
-    default_uid: int = Field(
-        os.getuid(),
+    default_uid: int | None = Field(
```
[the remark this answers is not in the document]
asmeurer: Actually this should be fine, but we need to add `from __future__ import annotations`. I guess we aren't testing old Python versions anywhere. What's the oldest version conda-store should support.

.github/workflows/tests.yaml
```diff
@@ -42,6 +40,24 @@ jobs:
...
+        with:
+          activate-environment: conda-store-server-dev
+          environment-file: conda-store-server/environment-windows-dev.yaml
+          auto-activate-base: false
+
+      # This fixes a "DLL not found" issue importing ctypes from the hatch env
+      - name: Reinstall Python 3.10 on Windows runner
+        uses: nick-fields/retry@v2.8.3
+        with:
+          timeout_minutes: 9999
```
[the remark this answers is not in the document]
asmeurer: FYI this was copied from somewhere else https://github.com/linz/emergency-management-tools/commit/d1d00931544f4a71ec1f7865041b16fd13d907a6

.github/workflows/tests.yaml
asmeurer: I just did this for debugging and forgot to remove it. The benefit is it forces pytest to show the full context in error messages.

.github/workflows/tests.yaml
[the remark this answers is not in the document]
nkaretnikov: Here are the docs https://github.com/nick-fields/retry. I guess the idea here is that it fails after doing max_attempts rather than timeouting, which might be flaky. I think it's fine to keep.

conda-store-server/conda_store_server/schema.py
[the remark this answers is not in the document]
nkaretnikov: I don't know what the minimal version is. The point I was trying to make: we don't use it anywhere else. I don't see why we should introduce this and immediately bump our lowest supported version to 3.10+ or require an extra import. Let's just use Optional, like everywhere else.

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: Oh, good catch! This can actually fail to consider some timedelta data, too:

```py
In [10]: import datetime
    ...:
    ...: t = datetime.timedelta(hours=24, minutes=2)

In [11]:

In [11]: t.seconds  # only considers minutes after 24 hours
Out[11]: 120

In [12]: int(t.total_seconds())  # considers total seconds in timedelta
Out[12]: 86520
```

conda-store-server/conda_store_server/storage.py
```diff
@@ -223,7 +224,7 @@ def get(self, key):
             return f.read()
 
     def get_url(self, key):
-        return os.path.join(self.storage_url, key)
+        return posixpath.join(self.storage_url, key)
```
nkaretnikov, line 227: Do you remember what failed without this? The effect of this change will be converting Windows paths to posixpaths:

```py
>>> import posixpath
>>>
>>> posixpath.join("foo", "bar")
'foo/bar'
>>>
>>> import os
>>> os.path.join("foo","bar")
'foo\\bar'
```

conda-store-server/conda_store_server/utils.py
nkaretnikov: I mean, it can stay if you want. I just dislike having unrelated changes in PRs. This PR is called "Windows support", this part is for macOS. Here's why:

- If you want to keep it, this now requires me to test this part on macOS before we merge (both on x86_64 and ARM64).
- If there are problems with this part on macOS, this will block the PR from being merged even if the Windows part is fine.
- If problems are discovered later, we cannot just use `git-revert` to rollback and need to manually create a PR, which takes longer.

conda-store-server/environment-windows-dev.yaml
```diff
@@ -0,0 +1,56 @@
+name: conda-store-server-dev
```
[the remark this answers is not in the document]
nkaretnikov: OK to ignore this.

docs/contributing.md
nkaretnikov: OK, the option is available in help anyway. Let's keep it simple.

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,52 @@ def chdir(directory: pathlib.Path):
...
+            if st.st_ino in have:
+                continue
+            have.add(st.st_ino)
+            apparent_total_bytes += st.st_size
+        for d in dirnames:
+            dp = os.path.join(dirpath, d)
+            if os.path.islink(dp):
+                apparent_total_bytes += os.lstat(dp).st_size
+
+    return apparent_total_bytes
```
nkaretnikov, line 83: Have you seen an implementation I shared here? https://github.com/conda-incubator/conda-store/pull/559#discussion_r1342122743

It fixes some style issues and the code is shorter. I also find it more readable because variables are shorter. The first two ifs can be written as one.

Direct link to gist: https://gist.github.com/nkaretnikov/1a66b90a74fa805f1022e90252e54c87

There are also tests there that you could re-use, like the file/directory creation logic. Ignore the part where I check against tools, it should just check against hardcoded numbers. But it's not a blocker anyway.

conda-store-server/tests/test_actions.py
```diff
@@ -15,9 +15,17 @@ def test_action_decorator():
     def test_function(context):
         print("stdout")
         print("stderr", file=sys.stderr)
-        context.run(["echo", "subprocess"])
-        context.run("echo subprocess_stdout", shell=True)
-        context.run("echo subprocess_stderr 1>&2", shell=True)
+        if sys.platform == "win32":
```
nkaretnikov, line 18: Can we use the newly defined ON_WIN here as well?

conda-store-server/tests/test_usage.py
```diff
@@ -0,0 +1,32 @@
...
+    dir_size = du(test_dir)
+    assert abs(dir_size - int(disk_usage(test_dir))) <= 1000
+
+    test_file = test_dir / "test_file"
+    test_file.write_text("a"*1000)
+    test_file2 = test_dir / "test_file2"
+    test_file2.write_text("b"*1000)
+    # Test hard links
+    test_file_hardlink = test_dir / "test_file_hardlink"
+    test_file_hardlink.hardlink_to(test_file)
```
nkaretnikov, line 19: This is not tested, right? Also symlinks are not tested.

conda-store-server/tests/test_usage.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # The exact values depend on the block size and other details. Just check
+    # that it is in the right range. Note that disk_usage uses the du command
+    # on Mac and Linux but uses the Python du() function on Windows.
+    val = disk_usage(test_file)
+    assert isinstance(val, str)
+    assert 1000 <= int(val) <= 1500
+    assert 1000 <= du(test_file) <= 1500
```
nkaretnikov, line 27: Our test file is 1000 bytes. A bit concerning that our margin is 50%. On multiple files, it'll add up quickly. Perhaps having a tighter range would be possible if we have an if-then-else for each platform.

nkaretnikov:
@asmeurer I've run this on Windows on 2216f47c5839c52d0116e6c7fd5987e013b27100. It built a simple env in standalone mode. I had to set 1 in the registry to allow for longer paths, but that's expected.

I've also seen this error, not sure whether it's something we should worry about for now:

```
[2023-10-18 20:55:02,917: ERROR/SpawnPoolWorker-3] Chord '73b673b3-172c-4544-8275-d285b626fd47' raised: FileNotFoundError(2, 'The system cannot find the file specified', None, 2, None)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\celery\app\builtins.py", line 78, in unlock_chord
    ret = j(
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\celery\result.py", line 774, in join
    value = result.get(
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\celery\result.py", line 246, in get
    self.maybe_throw(callback=callback)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\celery\result.py", line 364, in maybe_throw
    self.throw(value, self._to_remote_traceback(tb))
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\celery\result.py", line 357, in throw
    self.on_ready.throw(*args, **kwargs)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\vine\promises.py", line 234, in throw
    reraise(type(exc), exc, tb)
  File "C:\dev\miniconda3\envs\conda-store-server-dev\lib\site-packages\vine\utils.py", line 30, in reraise
    raise value
FileNotFoundError: [WinError 2] The system cannot find the file specified
[2023-10-18 20:55:02,964: INFO/SpawnPoolWorker-3] Task celery.chord_unlock[b87d024f-15ef-4f03-9761-56bc78d632d4] succeeded in 0.04699999999979809s: None
```

I'm waiting for you to rebase and address some of my comments. But overall this looks good, this is initial Windows support. We can fix other things in follow-ups. Thanks!

asmeurer:
Regarding the CI, hopefully the other changes don't break things here. I had to do quite a few workarounds to get stuff working on CI, due to issues with defaults/conda-forge and also some issues with mamba. On the other hand, it's also possible the new miniforge stuff will make some of my workaround here no longer necessary, so we should definitely try removing them.

nkaretnikov:
Taking over. See #640.

conda-store-server/conda_store_server/storage.py
asmeurer: Windows backslash paths are not correct for file URLs. Without this change, some of the buttons in the UI don't work because they use `%5C` (URL encoded `\`) instead of `/`.

conda-store-server/conda_store_server/utils.py
asmeurer: The reason I did this here is because you requested it in your previous review https://github.com/conda-incubator/conda-store/pull/559#discussion_r1341985604. If you want to refactor it or split it out, that's fine. Based on the discussion the standup last week, my impression was that it isn't actually that important that this be super accurate.

conda-store-server/tests/test_usage.py
asmeurer: The margin would be less if the files were bigger. It has to do with block rounding and the size of a directory. The exact value depends on the platform and filesystem.

asmeurer:
> I've also seen this error, not sure whether it's something we should worry about for now:

You'd need to debug to see the real source of the error.

PR #567, opened 2023-09-09
Add a test for concurrent builds

### LINKED ISSUE #548: [ENH] - Test for parallel builds
### Feature description

We need the ability to build environments in parallel. This functionality was added in https://github.com/conda-incubator/conda-store/issues/392 but there was issues and the work needed to be reverted. 

We need to re-implement this feature. 

### Value and/or benefit

Building all environments in series is not efficient and once there are a number of people on the platform, wait times for builds grows to an unacceptable length of time.

### Anything else?

_No response_

nkaretnikov:
Fixes #548.

### DIFF
--- a/conda-store-server/tests/test_server.py
+++ b/conda-store-server/tests/test_server.py
@@ -1,5 +1,9 @@
+import datetime
+import inspect
 import json
+import time
 
+import celery
 import pytest
 import yaml
 from conda_store_server import __version__, schema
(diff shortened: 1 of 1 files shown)


nkaretnikov:
Summary:
- Adds a test checking that `create_build` tasks run concurrently. This function is called when the user clicks on the rebuild button. I deliberately overwrite inner _functions_ rather than _tasks_ to have less influence on celery-specific bits
- This test ignores potential concurrency issues in the DB layer or celery misconfiguration (broker_url, result_backend). E.g., tasks can be concurrent, but if things wait on each other somewhere else (DB, channels), it'll affect the system
- Default celery concurrency parameter values are used, but they are not compared to our app (some refactoring might be needed, e.g., we read the concurrency value via `--concurrency`, which is why I haven't done it)
- Potential issues due to `celery_worker` not being the same worker our app uses.

trallard:
@nkaretnikov can you please resolve the merge conflicts?

nkaretnikov:
@trallard Resolved and rebased on main.

nkaretnikov:
@trallard A gentle reminder to take a look at this.

nkaretnikov:
> is this just testing that you can trigger multiple builds from the api and the will run concurently? I see you have mocked out the actual building.

Yes, I wanted to start with this. This PR is itself useful. For instance, I've used it to test concurrency support on Windows.

> The building of the conda environments concurently is what needs to be tested. This is due to in the past conda would have a shared package cache which would conflict. We need to check that two builds can trigger at the same time, install the same packages, and both successfully complete.

My plan was to add/discuss this after this PR was reviewed.

nkaretnikov:
We had a discussion about "testing the DB" during today's sync.

AFAICT, the testsuite only tests with sqlite, see `conda_store_config`. SQLite is not super concurrent-friendly and can lock under certain circumstances. SQLite is also not what people run in prod (postgres). We should create tests representing real-world prod configs. There are existing deployments, so we can ask what people are using.

We have integration tests that require spinning up the server (in the toplevel dir). Perhaps we should just try hitting APIs and simulating typical parallel workflows. But it would also be not super pure. People reported an issue to me today where the queue gets filled up, we could try somehow triggering that via the API tests, as a start.

nkaretnikov:
OK. Let's not merge this one. It's complicated and doesn't test prod workflow. New plan is here: https://github.com/conda-incubator/conda-store/issues/598#issuecomment-1747223058

PR #583, opened 2023-09-18
Make conda-docker a dependency in Linux only in pyproject.toml

### LINKED ISSUE #578: [BUG] - Can't install conda-store-server on M1 MacOS
### Describe the bug

Running the installation command results in the following:

```bash
~/dev/research
➜ m create -n env-research -c main -c conda-forge "conda-store-server>=0.4.10"
conda-forge/osx-arm64                                       Using cache
conda-forge/noarch                                          Using cache
main/noarch                                                   No change
main/osx-arm64                                                No change
error    libmamba Could not solve for environment specs
    The following package could not be installed
    └─ conda-store-server >=0.4.10  is not installable because it requires
       └─ conda-docker, which requires
          └─ fakechroot, which does not exist (perhaps a missing channel).
critical libmamba Could not solve for environment specs
```

I'm using `micromamba` here but the same issue also manifests when attempting installation via `conda`.

### Expected behavior

I'd expect the package to install.

Run `conda create -n env-research -c main -c conda-forge "conda-store-server>=0.4.10"` in a terminal running on an M1 Mac.

### Output

See above.

```bash
~/dev/research took 32.5s
➜ conda --version
conda 23.7.4

~/dev/research
➜ m --version
1.5.0
```

### Anything else?

Many thanks for building this tool and supporting the community!

asmeurer:
Fixes #578 .

Note, we will also need to make sure to fix this in the conda-forge feedstock.

## Description

This pull request:

- a
- b
- c

## Pull request checklist

### DIFF
--- a/conda-store-server/pyproject.toml
+++ b/conda-store-server/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
   "Programming Language :: Python :: 3 :: Only",
 ]
 dependencies = [
-  "conda-docker",
+  "conda-docker; sys_platform == 'linux'",
   "conda-pack",
   "celery",
   "sqlalchemy",


asmeurer:
Note that conda-docker is actually perfectly pip installable on all platforms, it just won't work when the platform isn't Linux (but it can be imported).

asmeurer:
The reason is that the Linux-only dependency, fakechroot, is not a Python package. It's an executable that gets called at runtime. It is, however, a dependency in the conda package which cannot be installed on Linux.

trallard (approved):
Thank you, this LGTM 🚀  merging now!

Issue #595, opened 2023-09-23
Do not break public interfaces, have a policy


nkaretnikov:
### Context

We need to have a backward compatibility policy.

This policy needs to specify what our public interfaces are. Example interfaces:
- conda-store is used as a library via Python
- conda-store web API is used via HTTP.

This policy needs to document how we make changes to those interfaces. Examples:
- a warning is raised for N conda-store releases before removing functionality
- an error is raised after that.

As an example, here's BC/FC policy for PyTorch: https://github.com/pytorch/pytorch/wiki/PyTorch's-Python-Frontend-Backward-and-Forward-Compatibility-Policy

An example where a Python API change broke nebari: https://github.com/conda-incubator/conda-store/issues/593#issuecomment-1732244137.

### Value and/or benefit

We don't break other software that depends on conda-store, like nebari.
When we introduce changes, things are easier to debug because an error/warning is raised.

nkaretnikov:
With this policy, the goal should be to have an answer to this immediately: "Is this a public API or not?"

pavithraes:
xref: #399

kcpevey:
This has been discussed with the team. @nkaretnikov will add notes from the discussion here.

nkaretnikov:
Proposed conda-store backward- and forward-compatibility policy

### Making changes

Incompatible changes should be avoided if possible. Note that this only applies to client-facing code, such as HTTP routes or public Python APIs.

For example, changes to a database table can be done without versioning if data can be migrated in a backward- and forward-compatible way. Everything in the `tests` directory can also be changed as we want.

When it's not possible to make a compatible change, client disruption needs to be minimized via versioning, see below.

### HTTP routes

Route handlers need to be versioned.

As of writing this, we already have the `v1` API.

New incompatible routes or changes to existing routes should be part of the new API.
Do not add more things to `v1` if they are incompatible.
New features that are compatible can be added, though, like completely new functionality exposed through a new route.

Incompatible changes need to be declared as part of the `experimental` namespace.

Experimental routes have no guarantees attached to them, we can remove them at any moment. This allows us to add features without committing to support, and test it with clients in real-world scenarios.

Once we determine the routes are stable and want to support them, they need to move to the versioned API, like `v2` or later.

Note: one a new API version is introduced, the rest of routes, which are compatible, need to add a new declaration, saying they are compatible with this new API.

[...]

PR #607, opened 2023-10-03
Change API to be able to assign roles to namespaces

### LINKED ISSUE #491: Improvements to the namespace Role Mappings

nkaretnikov:
Fixes #491.


nkaretnikov:
**This is WIP. Do not review.**

conda-store-server/conda_store_server/api.py
```diff
@@ -58,6 +57,20 @@ def update_namespace(
...
 
+    db.commit()
+
+    return namespace
+
+
+def update_namespace_role(
+    db,
+    name: str,
+    role_mappings: Dict[str, List[str]] = None,
```
nkaretnikov, line 68: This is interesting. Because this is typed `List[str]` instead of `Any`, like metadata, this now gets rejected by the FastAPI router with 422. In order to fix it, I had to make this change to the route signature:

```py
    role_mappings: Dict[str, List[str]] = None,  # doesn't work
    role_mappings: Dict[str, Any] = None,  # works
```

Printing `request.body()` before the route is being processed, I see the same value in both cases. So it must be an internal issue related to serialization, like this one: https://github.com/tiangolo/fastapi/issues/3373

Code to print request body (add it to `conda-store-server/conda_store_server/server/app.py`):

```py
        if self.enable_api:
            from fastapi import Depends
            async def log_json(request: Request):
                print("XXXXXXXXXXXXXXXXX", await request.body())
            app.include_router(
                views.router_api,
                prefix=trim_slash(self.url_prefix),
                dependencies=[Depends(log_json)],
            )
```

nkaretnikov: Ah, I see. After splitting the route into two, it's no longer necessary to include the keyword matching the route parameter. So no need to pass `role_mappings: dict`. This can be just the inner dict directly.

tests/test_api.py
```diff
@@ -10,6 +10,7 @@
 """
```
nkaretnikov, line 10: View this file locally with the following, it produces a much better diff:

```
[diff]
  algorithm = patience
```

conda-store-server/conda_store_server/orm.py
```diff
@@ -54,39 +54,36 @@ class Namespace(Base):
 
     metadata_ = Column(JSON, default=dict, nullable=True)
 
-    role_mappings = relationship("NamespaceRoleMapping", back_populates="namespace")
```
nkaretnikov, line 57: Not sure it's possible to have this field since I now use two foreign keys in the NamespaceRoleMapping table.
In any case, it doesn't seem to be necessary.
So I just removed it.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -233,31 +236,33 @@ def is_subset_entity_permissions(self, entity, new_entity):
...
+            # Used in tests to check default permissions without populating the
+            # DB, which raises an exception since namespace is not found.
+            try:
+                roles = api.get_other_namespace_roles(db, name=entity.primary_namespace)
+            except Exception:
+                return res
+
+            for x in roles:
+                res[self.convert_namespace_to_entity_arn(x.namespace)].add(x.role)
+            return res
```
nkaretnikov, line 265: I still need to look at our current auth tests and see if this is sufficiently tested.
Since this is used for authorization, it's worth looking carefully.

conda-store-server/conda_store_server/alembic/versions/46bdf428642d_change_namespacerolemapping_and_.py
```diff
@@ -0,0 +1,59 @@
...
+# revision identifiers, used by Alembic.
+revision = "46bdf428642d"
+down_revision = "b387747ca9b7"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column(
```
nkaretnikov, line 20: This fails with SQLite because it doesn't have ALTER COLUMN.

I think I'll just drop/create the entire table since any data potentially stored there would be inconsistent anyway.

conda-store-server/conda_store_server/alembic/versions/46bdf428642d_change_namespacerolemapping_and_.py
```diff
@@ -0,0 +1,47 @@
...
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('namespace_id', sa.Integer(), nullable=False),
+        sa.Column('entity', sa.Unicode(length=255), nullable=False),
+        sa.Column('role', sa.Unicode(length=255), nullable=False),
+        sa.ForeignKeyConstraint(['namespace_id'], ['namespace.id'], ),
+        sa.PrimaryKeyConstraint('id')
+    )
+    # Note: data is NOT copied before dropping the old table
+    op.drop_table("namespace_role_mapping")
+    op.rename_table("namespace_role_mapping_new", "namespace_role_mapping")
```
nkaretnikov, line 47: Migration seems to work with Postgres and SQLite. But it breaks new API tests, which used to pass, probably because I'm not copying data, so namespace URLs 404:

```
FAILED tests/test_api.py::test_update_namespace_metadata[True] - requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost:5000/conda-store/api/v1/namespace/...
FAILED tests/test_api.py::test_update_namespace_metadata[False] - assert 404 == 403
FAILED tests/test_api.py::test_crud_namespace_roles[True] - requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://localhost:5000/conda-store/api/v1/namespace/...
FAILED tests/test_api.py::test_crud_namespace_roles[False] - assert 404 == 403
```

nkaretnikov: Hmm, it passed on CI. But `test_namespace_role_mapping` needs to be updated to use new schema anyway.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+    assert roles[0].other_namespace == 'pytest-other-namespace1'
+    assert roles[0].role == 'admin'
+
+    assert roles[1].id == 3
+    assert roles[1].namespace == 'pytest-namespace'
+    assert roles[1].other_namespace == 'pytest-other-namespace3'
+    assert roles[1].role == 'viewer'
+
+    # Deletes all role mappings
+    api.delete_namespace_roles(db, name='pytest-namespace')
```
nkaretnikov, line 133: Use a variable for name.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+    api.create_namespace_role(db, name=namespace_name, other=other_namespace_name2, role="admin")
+    api.create_namespace_role(db, name=namespace_name, other=other_namespace_name3, role="viewer")
+    db.commit()
 
-    # create a Role Mapping, with a failing entity
-    with pytest.raises(Exception):
-        NamespaceRoleMapping(
-            namespace=namespace, namespace_id=namespace.id, entity="invalid_entity_name"
-        )
+    # Attempts to create a role mapping, with an invalid role
```
nkaretnikov, line 65: Remove comma.

conda-store-server/conda_store_server/api.py
```diff
@@ -58,28 +58,154 @@ def update_namespace(
...
+def get_namespace_roles(
+    db,
+    name: str,
+):
+    """Which namespaces can access namespace 'name'?"""
+    namespace = get_namespace(db, name)
+    if namespace is None:
+        raise ValueError(f"Namespace='{name}' not found")
 
-    db.commit()
```
nkaretnikov, line 80: This used to have a `db.commit`, which I removed to allow for bulk updates, without committing after every update. That's also what other methods do.

conda-store-server/conda_store_server/api.py
```diff
@@ -58,28 +58,154 @@ def update_namespace(
...
+    q = (
+        db.query(nrm.id, this.name, other.name, nrm.role)
+        .filter(nrm.namespace_id == namespace.id)
+        .filter(nrm.other_namespace_id == other_namespace.id)
+        .filter(nrm.namespace_id == this.id)
+        .filter(nrm.other_namespace_id == other.id)
+        .first()
+    )
+    if q is None:
+        return None
```
nkaretnikov, line 145: This is necessary because `from_list` fails on None.

conda-store-server/conda_store_server/schema.py
```diff
@@ -615,6 +620,28 @@ class APIGetNamespace(APIResponse):
     data: Namespace
 
 
+# GET /api/v1/namespace/{name}/role
```
nkaretnikov, line 623: These will need to be updated to v2 when I make these part of the new API.

conda-store-server/conda_store_server/schema.py
```diff
@@ -615,6 +620,28 @@ class APIGetNamespace(APIResponse):
...
+class APIGetNamespaceRole(BaseModel):
+    other_namespace: str
```
nkaretnikov, line 625: It's a bit strange to pass JSON for a GET, but I wanted to have similar interfaces everywhere. And I don't think it's forbidden by REST. I also didn't want to have too many parameters in the URL itself. I'm happy to reconsider if people feel strongly about this and can provide arguments against this.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -188,14 +188,17 @@ def convert_roles_to_permissions(self, roles):
             permissions = permissions | self.role_mappings[role]
         return permissions
 
-    def get_entity_binding_permissions(self, entity):
+    def convert_namespace_to_entity_arn(self, namespace):
+        return f"{namespace}/*"
```
nkaretnikov, line 192: I didn't want to change the auth part too much, so I just convert to a compatible format here. Namespace role mappings use concrete roles (like, "viewer") rather than this ARN (regex) format. You can think of a namespace as a username.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -188,14 +188,17 @@ def convert_roles_to_permissions(self, roles):
...
+
+    def get_entity_binding_permissions(self, entity: schema.AuthenticationToken):
```
nkaretnikov, line 194: The name entity is not very descriptive, so I add types in related functions.

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: The try-except part works in tests without any DB data because we have default permissions hardcoded in this Python file. All permissions are merged in `get_entity_bindings`.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -233,31 +236,33 @@ def is_subset_entity_permissions(self, entity, new_entity):
...
+            # Must have the same format as authenticated_role_bindings:
+            # {
+            #     "default/*": {"viewer"},
+            #     "filesystem/*": {"viewer"},
+            # }
+            res = defaultdict(set)
+
+            # FIXME: Remove try-except.
+            # Used in tests to check default permissions without populating the
+            # DB, which raises an exception since namespace is not found.
```
nkaretnikov, line 257: This FIXME is for later, won't be addressed as part of this PR.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -233,31 +236,33 @@ def is_subset_entity_permissions(self, entity, new_entity):
...
+            try:
+                roles = api.get_other_namespace_roles(db, name=entity.primary_namespace)
```
nkaretnikov, line 259: We use `other` here because we're trying to find which namespaces we have access to. The non-other one is the opposite - which namespaces have access to us.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -310,17 +310,14 @@ async def api_create_namespace(
...
     namespace: str,
     request: Request,
-    metadata: Dict[str, Any] = None,
-    role_mappings: Dict[str, List[str]] = None,
-    auth=Depends(dependencies.get_auth),
-    conda_store=Depends(dependencies.get_conda_store),
+    auth,
+    conda_store,
+    func,
+    **kwargs,
```
nkaretnikov, line 320: All these routes are pretty similar, so I define a common function to make it easier to write and review.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -339,11 +336,171 @@ async def api_update_namespace(
...
+):
+    _api_namespace_common(
+        namespace=namespace,
+        request=request,
+        auth=auth,
+        conda_store=conda_store,
+        func=api.update_namespace_metadata,
+        metadata_=metadata,
+    )
+    return {"status": "ok"}
```
nkaretnikov, line 365: The original method was split into two: metadata and roles. Chris requested this because the old method was doing too much.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
-    # check that only one namespace exists
-    assert len(api.list_namespaces(db).all()) == 1
+    # Checks that all namespaces exist
+    assert len(api.list_namespaces(db).all()) == 4
+
+    # Creates role mappings
+    api.create_namespace_role(db, name=namespace_name, other=other_namespace_name1, role="admin")
+    api.create_namespace_role(db, name=namespace_name, other=other_namespace_name2, role="admin")
+    api.create_namespace_role(db, name=namespace_name, other=other_namespace_name3, role="viewer")
+    db.commit()
```
nkaretnikov, line 63: admin is repeated twice on purpose - I overwrite it in this test later. This also makes sure we don't somehow not allow duplicate roles without considering other keys.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
 
-    # create a Role Mapping, with a failing entity
-    with pytest.raises(Exception):
-        NamespaceRoleMapping(
-            namespace=namespace, namespace_id=namespace.id, entity="invalid_entity_name"
-        )
+    # Attempts to create a role mapping with an invalid role
+    with pytest.raises(ValueError, match=r"invalid role=invalid-role"):
+        api.create_namespace_role(db, name=namespace_name, other=other_namespace_name3, role="invalid-role")
+        db.commit()
```
nkaretnikov, line 68: This tests that the checking function that we have on the ORM side works.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+    # Attempts to create a role mapping violating the uniqueness constraint
+    with pytest.raises(
+            Exception,
+            match=(r"UNIQUE constraint failed: "
+                   r"namespace_role_mapping.namespace_id, "
+                   r"namespace_role_mapping.other_namespace_id")):
+        # Runs in a nested transaction since a constraint violation will cause a rollback
+        with db.begin_nested():
+            api.create_namespace_role(db, name=namespace_name, other=other_namespace_name2, role="developer")
+            db.commit()
```
nkaretnikov, line 79: Here's why admin above is important. I attempt to crate a new mapping but with developer. role should not be considered since we want to have one mapping for a pair of ids. So this works correctly.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+
+    # Updates a role mapping
+    api.update_namespace_role(db, name=namespace_name, other=other_namespace_name2, role="developer")
+    db.commit()
```
nkaretnikov, line 83: Here, a similar command, but with update, runs fine - as it should.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
-    )
+    assert roles[1].id == 2
+    assert roles[1].namespace == namespace_name
+    assert roles[1].other_namespace == other_namespace_name2
+    assert roles[1].role == 'developer'
+
+    assert roles[2].id == 3
+    assert roles[2].namespace == namespace_name
+    assert roles[2].other_namespace == other_namespace_name3
+    assert roles[2].role == 'viewer'
```
nkaretnikov, line 103: All three mappings are inserted now, they also have different roles to make it easier to distinguish.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
 
-    NamespaceRoleMapping(namespace=namespace, namespace_id=namespace.id, entity="*/*")
+    # Gets other role mappings
+    roles = api.get_other_namespace_roles(db, other_namespace_name1)
+    db.commit()
+    assert len(roles) == 1
+    roles = api.get_other_namespace_roles(db, namespace_name)
+    db.commit()
+    assert len(roles) == 0
```
nkaretnikov, line 111: We also test the `other` variant of the get function, which is used in authentication. One namespace is part of the original three mappings, which we've just added, so 1 entry is returned as expected (it has access to one namespace). The namespace that we give access to we didn't insert anywhere, so it has no access to any other namespaces.

nkaretnikov: Perhaps I should check the other values here, as above, not just the lenghts.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+
+    assert roles[0].id == 1
+    assert roles[0].namespace == namespace_name
+    assert roles[0].other_namespace == other_namespace_name1
+    assert roles[0].role == 'admin'
+
+    assert roles[1].id == 3
+    assert roles[1].namespace == namespace_name
+    assert roles[1].other_namespace == other_namespace_name3
+    assert roles[1].role == 'viewer'
```
nkaretnikov, line 130: Delete one mapping and see how it's reflected in the DB. Namespace2 is deleted as it should.

conda-store-server/tests/test_db_api.py
```diff
@@ -39,35 +39,104 @@ def test_namespace_crud(db):
...
+
+    # Deletes all role mappings
+    api.delete_namespace_roles(db, name=namespace_name)
+    db.commit()
+
+    # Checks that roles were deleted
+    roles = api.get_namespace_roles(db, name=namespace_name)
+    db.commit()
+    assert len(roles) == 0
```
nkaretnikov, line 139: Check the bulk deletion works.

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+
+    r = schema.APIResponse.parse_obj(response.json())
+    if auth:
+        assert r.status == schema.APIStatus.OK
+        if data_pred is None:
+            assert r.data is None
+        else:
+            assert data_pred(r.data) is True
+    else:
+        assert r.status == schema.APIStatus.ERROR
```
nkaretnikov, line 597: All these tests are pretty similar, so I define a helper. data_pred = data predicate - a function returning a boolean.

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+@pytest.mark.parametrize("auth", [True, False])
+def test_update_namespace_metadata(testclient, auth):
+    namespace = f"filesystem"
+    make_request = partial(_crud_common, testclient=testclient, auth=auth)
+
+    make_request(
+        method=testclient.put,
+        route=f"api/v1/namespace/{namespace}/metadata",
+        json={"test_key1": "test_value1", "test_key2": "test_value2"},
+    )
```
nkaretnikov, line 609: Because I removed one parameter from the metadata handler (since I had to split it up), this doesn't require passing the "metadata" key in JSON as before.

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+
+    # Reads created role
+    make_request(
+        method=testclient.get,
+        route=f"api/v1/namespace/{namespace}/role",
+        json={
+            "other_namespace": other_namespace,
+        },
+        data_pred=lambda data: (
+            data['namespace'] == 'filesystem' and
```
nkaretnikov, line 648: use `namespace` here instead of hardcoding

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+            "role": "admin"
+        },
+    )
+
+    # Reads updated roles
+    make_request(
+        method=testclient.get,
+        route=f"api/v1/namespace/{namespace}/roles",
+        data_pred=lambda data: (
+            data[0]['namespace'] == 'filesystem' and
```
nkaretnikov, line 669: use `namespace` here instead of hardcoding

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+            data[0]['other_namespace'] == other_namespace and
+            data[0]['role'] == 'admin' and
+            len(data) == 1
+        ),
+    )
```
nkaretnikov, line 674: Role is admin now, only 1 entry in the table.

tests/test_api.py
```diff
@@ -660,6 +572,124 @@ def test_create_get_delete_namespace_auth(testclient):
...
+            "other_namespace": other_namespace,
+        },
+    )
+
+    # Reads roles to check if deleted
+    make_request(
+        method=testclient.get,
+        route=f"api/v1/namespace/{namespace}/roles",
+        data_pred=lambda data: data == [],
+    )
```
nkaretnikov, line 690: I structured this test to test all new API methods on role and roles.

nkaretnikov:
@costrouc @dcmcand **NOT READY FOR MERGE (SEE BELOW)**, but PTAL. This might have minor issues here and there, I'll do another review before this gets merged.

The design I'm trying to implement is here https://github.com/conda-incubator/conda-store/issues/491#issuecomment-1751773261. I had to deviate from that in some places, but it's mostly the same.

As requested by Tania and Chuck, I need to keep the old APIs and make the new APIs v2. On the ORM side, old tables will also stay until the deprecation period is over. For old APIs that I'm replacing, we'll communicate that those are deprecated (by updating documentation/release notes and also using HTTP). If you have opinions on how we should sunset/deprecate APIs on the HTTP side, let me know.

conda-store-server/conda_store_server/schema.py
nkaretnikov: I guess one argument might be is that it'd be easier to see the data in a browser if you don't have to pass external params.

conda-store-server/conda_store_server/orm.py
```diff
@@ -54,39 +54,37 @@ class Namespace(Base):
...
     role = Column(Unicode(255), nullable=False)
 
-    @validates("entity")
-    def validate_entity(self, key, entity):
-        if not ARN_ALLOWED_REGEX.match(entity):
-            raise ValueError(f"invalid entity={entity}")
-
-        return entity
-
     @validates("role")
```
costrouc, line 75: I'd be fine with us removing this validator since there can be more than admin, viewer, developer. Eventually we should validate this role properly using the database.

conda-store-server/conda_store_server/orm.py
```diff
@@ -54,39 +54,37 @@ class Namespace(Base):
...
-            raise ValueError(f"invalid entity={role}")
-
+            raise ValueError(f"invalid role={role}")
         return role
 
+    __table_args__ = (
+        # Ensures no duplicates can be added with this combination of fields.
+        # Note: this doesn't add role because role needs to be unique for each
+        # pair of ids.
+        UniqueConstraint("namespace_id", "other_namespace_id", name="_uc"),
```
costrouc, line 85: Wouldn't it be possible to assign `(namespace1, namespace2, 'viewer')` and `(namespace1, namespace2, 'admin')`. Roles are not always subsets of one another. So this should be  `UniqueConstraint("namespace_id", "other_namespace_id", "role', name="_uc"),`

conda-store-server/conda_store_server/schema.py
costrouc: It's common to use query parameters for GET requests. Could this be used? other_namespace is optional. And depending on if it is specified the scope of `GetNamespaceRoles` will return fewer results.

conda-store-server/conda_store_server/server/auth.py
costrouc: Looks great to me.

conda-store-server/conda_store_server/server/auth.py
costrouc, line 259: Like the much cleaner approach to use the api module to hide the sql logic.

costrouc (changes requested):

conda-store-server/conda_store_server/orm.py
nkaretnikov: For other reviewers: this has been discussed during the call. The current plan is to make the roles distinct since it's easier to reason about roles this way.

conda-store-server/conda_store_server/orm.py
nkaretnikov: For other reviewers: this has been discussed during the call. The current plan is to keep this as is.

conda-store-server/conda_store_server/schema.py
nkaretnikov: For other reviewers: I'll try using query params here. Pros: easier to test in a browser + caching.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -9,11 +9,37 @@
 from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request
 from fastapi.responses import PlainTextResponse, RedirectResponse
 
-router_api = APIRouter(
```
nkaretnikov, line 12: Renamed to `router_api_v1`. Based on the proposed [BC policy](https://github.com/conda-incubator/conda-store/issues/595#issuecomment-1783843666), I shouldn't have touched this, but I think there's zero chance this can be used by anyone except this application, since it's only for defining routes. So I renamed it to make it easier to see which API version is used.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -9,11 +9,37 @@
...
+router_api_v2 = APIRouter(
     tags=["api"],
-    prefix="/api/v1",
+    prefix=_API_PREFIX_V2,
 )
 
+router_api_experimental = APIRouter(
+    tags=["api"],
+    prefix=_API_PREFIX_EXPERIMENTAL,
+)
```
nkaretnikov, line 29: The experimental API is not currently used, but I've added it while I'm making changes to routing, since it's part of the BC policy proposal.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -9,11 +9,37 @@
...
+def _is_api_v1(request: Request) -> bool:
+    return request.scope["route"].path.startswith(_API_PREFIX_V1)
+
+
+def _is_api_v2(request: Request) -> bool:
+    return request.scope["route"].path.startswith(_API_PREFIX_V2)
+
+
+def _is_api_experimental(request: Request) -> bool:
+    return request.scope["route"].path.startswith(_API_PREFIX_EXPERIMENTAL)
```
nkaretnikov, line 41: These functions are not currently used, but I've added them while I'm making changes to routing, since they are useful. They allow determining, from the request handler function body, which API version the client is calling. This can be used to implement API-version-specific logic.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -109,18 +135,14 @@ def paginated_api_response(
     }
 
 
-@router_api.get(
-    "/",
-    response_model=schema.APIGetStatus,
-)
+@router_api_v1.get("/", response_model=schema.APIGetStatus)  # fmt: skip
+@router_api_v2.get("/", response_model=schema.APIGetStatus)  # fmt: skip
```
nkaretnikov, line 139: Here and elsewhere: since I needed to introduce the v2 API (and potentially other versions later), I decided to use the existing mechanism for defining versioned routes, as opposed to using a third-party library or something like that, because I don't want to make too many changes.

However, it's quite verbose, so I make sure routes are oneliners. This makes it easier to see if there are any changes between API versions, and takes less space line-wise. Because these oneliners can be long, I add a comment for the formatter, so they are not changed later.

nkaretnikov: Note: I'm not 100% sold on the oneliners idea because it requires changing existing routes, which might potentially introduce bugs if I'm not careful and requires reviewing those routes. However, we do have some endpoints, like `api_get_build_lockfile`, where the route declarations would take 40 LOC otherwise, so this does help.

nkaretnikov: I haven't looked into whether it's possible, but it might be also cool to configure the formatter to skip formatting lines starting with `@router`, to avoid having to insert skips everywhere, which is error-prone and verbose.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -310,10 +321,7 @@ async def api_create_namespace(
         return {"status": "ok"}
 
 
-@router_api.put(
-    "/namespace/{namespace}/",
-    response_model=schema.APIAckResponse,
-)
+@router_api_v1.put("/namespace/{namespace}/", response_model=schema.APIAckResponse)  # fmt: skip
```
nkaretnikov, line 324: This is a v1-specific route. v2 routes (defined below) replace it.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -346,7 +354,234 @@ async def api_update_namespace(
         return {"status": "ok"}
 
 
-@router_api.delete("/namespace/{namespace}/", response_model=schema.APIAckResponse)
```
nkaretnikov, line 349: This route is not removed. It's below this new block.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -346,7 +354,234 @@ async def api_update_namespace(
...
+@router_api_v2.put("/namespace/{namespace}/metadata", response_model=schema.APIAckResponse)  # fmt: skip
```
nkaretnikov, line 357: This section starts v2-only routes.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -346,7 +354,234 @@ async def api_update_namespace(
...
+):
+    with conda_store.get_db() as db:
+        auth.authorize_request(
+            request,
+            namespace,
+            {
+                Permissions.NAMESPACE_UPDATE,
+            },
+            require=True,
+        )
```
nkaretnikov, line 373: The previous version of this PR had a bug, where the required permissions were not matching the purpose of the handler because I reused the handler body in a common function. In this version of the PR, each handler is separate.

nkaretnikov: I also wonder if we need to have `NAMESPACE_READ` here and in other ones because we call `get_namespace` first.

nkaretnikov:
@costrouc PTAL.

New routes are part of the v2 API now. Old v1 routes (that are compatible) are marked as v2 as well. There's one incompatible v1 route.

This touches auth code. What I do is check the v2 roles first, then check v1.

I think we should later port the tests so that all v2 routes are tested as well, to ensure compatibility between v1 and v2. This PR only tests new v2 routes because it's very big already.

I've also added experimental route support while I'm changing this.

I'm not 100% sold on routes being oneliners, LMK what you think. I now feel like changing the oneliners back, just to keep the diff less verbose.

I'm also not sure whether having `NAMESPACE_ROLE_MAPPING_READ` permission implies we don't need to check for `NAMESPACE_READ`. I think we should check for both since there's no downside to do that, but there's a lot of downside if we introduce a mistake in permission-checking logic.

The commit we looked at previously is saved [here](https://github.com/nkaretnikov/conda-store/commit/9ac936bbb7c0eff6c338267ca25142618d4468c1) as part of [this branch](https://github.com/nkaretnikov/conda-store/tree/origin/role-mappings-api-491-old).

nkaretnikov:
TODO: Add namespace-read everywhere we use it.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -233,12 +237,24 @@ def is_subset_entity_permissions(self, entity, new_entity):
...
-    def database_role_bindings(self, entity):
+    def database_role_bindings(self, entity: schema.AuthenticationToken):
+        # This method can be reached from the router_ui via filter_environments.
+        # Since the UI routes are not versioned, we don't know which API version
+        # the client might be using. So we check all API versions, starting from
+        # the most recent one.
+        v2 = self._database_role_bindings_v2(entity)
+        if v2:
+            return v2
+        return self._database_role_bindings_v1(entity)
```
nkaretnikov, line 255: I don't like this solution anymore. Once there's any data in v2 (even if added by mistake), v1 will never be checked. I'll add a CLI parameter to check a particular table here. The other one will be ignored.

nkaretnikov: Resolved by adding an overridable field to the RBAC class: `RBACAuthorizationBackend.role_mappings_version`.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -58,6 +70,74 @@ def authenticate(self, token):
 
 
 class RBACAuthorizationBackend(LoggingConfigurable):
+    role_mappings_version = Integer(
+        1,
+        help="Role mappings version to use: 1 (default, legacy), 2 (new, recommended)",
+        config=True,
+    )
```
nkaretnikov, line 77: I make 1 the default so that users who already have existing role mappings could use conda-store without updating the config. Otherwise, their authorization would be broken until they either (1) set the value to 1 in the config or (2) inserted new roles into v2 tables using the new APIs.

docs/administration.md
```diff
@@ -413,6 +413,38 @@ should be fully configurable from those options.
 `GithubOAuthAuthentication.jupyterhub_url` is the url for connecting
 to JupyterHub. The URL should not include the `/hub/`.
 
+### `conda_store_server.server.auth.RBACAuthorizationBackend`
+
+`c.RBACAuthorizationBackend.role_mappings_version` specifies the role mappings
+version to use: 1 (default, legacy), 2 (new, recommended)
```
nkaretnikov, line 419: Dot missing at the end of the sentence.

nkaretnikov:
Since this uses a DB migration, this will need to be rebased on updated main before merging.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -254,6 +254,14 @@ async def http_exception_handler(request, exc):
                 status_code=exc.status_code,
             )
 
+        # Prints exceptions to the terminal
+        # https://fastapi.tiangolo.com/tutorial/handling-errors/#re-use-fastapis-exception-handlers
+        # https://github.com/tiangolo/fastapi/issues/1241
+        @app.exception_handler(Exception)
+        async def exception_handler(request, exc):
+            print(exc)
+            return await http_exception_handler(request, exc)
```
nkaretnikov, line 263: Added this to print an exception to the terminal when an invalid `role_mappings_version` value is used. Previously, nothing was printed except "Invalid server error".

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: This is documented in `docs/administration.md`.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -58,18 +70,88 @@ def authenticate(self, token):
...
+                ),
+                {"primary_namespace": entity.primary_namespace},
+            )
+            raw_role_mappings = result.mappings().all()
+
+            db_role_mappings = defaultdict(set)
+            for row in raw_role_mappings:
+                db_role_mappings[row["entity"]].add(row["role"])
+
+        return db_role_mappings
```
nkaretnikov, line 98: This code used to be in `database_role_bindings`. I moved it to the start of the class to use it in `_role_mappings_versions`.

conda-store-server/tests/test_auth.py
```diff
@@ -140,36 +141,191 @@ def test_authorization(conda_store, entity_bindings, arn, permissions, authorize
     assert authorized == authorization.authorize(entity, arn, permissions)
 
 
-def test_end_to_end_auth_flow(conda_store):
+@pytest.mark.parametrize("role_mappings_version", [1, 2])
```
nkaretnikov, line 144: Updated this test to match the newly added `test_end_to_end_auth_flow_v2`. The v1 test calls HTTP APIs that use the v1 role mappings table. The v2 test does the same for the v2 table.

conda-store-server/tests/test_auth.py
```diff
@@ -140,36 +141,191 @@ def test_authorization(conda_store, entity_bindings, arn, permissions, authorize
...
+            Permissions.NAMESPACE_READ,
+            Permissions.NAMESPACE_UPDATE,
+            Permissions.NAMESPACE_ROLE_MAPPING_CREATE,
+            Permissions.NAMESPACE_ROLE_MAPPING_READ,
+            Permissions.NAMESPACE_ROLE_MAPPING_UPDATE,
+            Permissions.NAMESPACE_ROLE_MAPPING_DELETE,
+            Permissions.SETTING_READ,
+            Permissions.SETTING_UPDATE,
+        },
+    )
```
nkaretnikov, line 194: TODO: the above block should be indented.

conda-store-server/conda_store_server/api.py
```diff
@@ -82,6 +83,176 @@ def update_namespace(
...
+def get_namespace_roles(
+    db,
+    name: str,
+):
+    """Which namespaces can access namespace 'name'?"""
+    namespace = get_namespace(db, name)
+    if namespace is None:
+        raise ValueError(f"Namespace='{name}' not found")
+
+    nrm = aliased(orm.NamespaceRoleMappingV2)
```
dcmcand, line 112: Can we please have more descriptive variable names here? Readability is part of maintainability.

conda-store-server/conda_store_server/api.py
```diff
@@ -82,6 +83,176 @@ def update_namespace(
...
+def get_other_namespace_roles(
+    db,
+    name: str,
+):
+    """To which namespaces does namespace 'name' have access?"""
+    namespace = get_namespace(db, name)
+    if namespace is None:
+        raise ValueError(f"Namespace='{name}' not found")
+
+    nrm = aliased(orm.NamespaceRoleMappingV2)
```
dcmcand, line 135: Same comment on variable names.

conda-store-server/conda_store_server/api.py
```diff
@@ -82,6 +83,176 @@ def update_namespace(
...
+    db.query(nrm).filter(nrm.namespace_id == namespace.id).delete()
+
+
+# v2 API
+def get_namespace_role(
+    db,
+    name: str,
+    other: str,
+):
+    namespace = get_namespace(db, name)
```
dcmcand, line 167: This block appears a lot of times. It may be worth extracting it into something like `get_existing_namespace_by_name` or something like that which wraps `get_namespace` and raises a valueerror if the name doesn't exist.

This is a stylistic suggestion and not a blocker if you choose not to implement it.

conda-store-server/conda_store_server/orm.py
```diff
@@ -89,6 +89,37 @@ def validate_role(self, key, role):
...
+
+    __tablename__ = "namespace_role_mapping_v2"
+
+    id = Column(Integer, primary_key=True)
+    # Provides access to this namespace
+    namespace_id = Column(Integer, ForeignKey("namespace.id"), nullable=False)
+    namespace = relationship(Namespace, foreign_keys=[namespace_id])
+
+    # ... for other namespace
+    other_namespace_id = Column(Integer, ForeignKey("namespace.id"), nullable=False)
```
dcmcand, line 103: As mentioned above, these names don't make it clear which namespace has permissions to which.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -233,31 +317,21 @@ def is_subset_entity_permissions(self, entity, new_entity):
...
-            raw_role_mappings = result.mappings().all()
-
-            db_role_mappings = defaultdict(set)
-            for row in raw_role_mappings:
-                db_role_mappings[row["entity"]].add(row["role"])
-
-        return db_role_mappings
+    def database_role_bindings(self, entity: schema.AuthenticationToken):
+        # This method can be reached from the router_ui via filter_environments.
+        # Since the UI routes are not versioned, we don't know which API version
```
dcmcand, line 329: Why are ui routes not versioned? should they be?

dcmcand (commented):
The comments in this review are mostly stylistic and are not blocking for approval.

conda-store-server/conda_store_server/alembic/versions/46bdf428642d_change_namespacerolemapping_and_.py
```diff
@@ -0,0 +1,56 @@
+"""Change NamespaceRoleMapping and Namespace
```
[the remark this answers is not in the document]
nkaretnikov: I think you commented on an old change. When I click on the file, I cannot find it and it has the "Outdated" label.

conda-store-server/conda_store_server/alembic/versions/771180018e1b_add_v2_role_mappings.py
```diff
@@ -0,0 +1,39 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "namespace_role_mapping_v2",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("namespace_id", sa.Integer(), nullable=False),
+        sa.Column("other_namespace_id", sa.Integer(), nullable=False),
```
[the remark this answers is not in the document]
nkaretnikov: Chris and I discussed it over a call before. There's no obviously good name to use here that would immediately be descriptive. I use the C++-style naming here that's used in C++ for classes. "This" means "me" and "other" means "something else that's not me". Comments are here to clarify.

conda-store-server/conda_store_server/api.py
nkaretnikov: In general, I agree. But not here. I deliberately use short names so that queries don't get split up over multiple lines by the linter, which would be hard to read. All these DB functions are very small, so there's no concern of this clashing with something else causing confusion. I couldn't think of a short name for NamespaceRoleMapping other than using the first letters.

conda-store-server/conda_store_server/api.py
nkaretnikov: Addressed in a comment above.

conda-store-server/conda_store_server/api.py
nkaretnikov: I agree in general, but not going to change this. I had similar code in the original PR that I abstracted out into a common function. I was later told in a meeting that copy-pasting is preferred. Because it allows reviewers to focus on part of the code without referencing anything else and helps with refactoring since each part is independent from one another.

conda-store-server/conda_store_server/orm.py
nkaretnikov: Addressed in another comment above.

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: This is out of scope of this PR. Maybe. It's just how `router_ui` is defined.

But even if these were versioned, it wouldn't solve the problem here because nothing prevents someone from calling HTTP routes that access v1 and v2 tables. If they do, which table should be used here? Well, no way to tell without the command-line switch, which is why I introduced it, because it shows the intent and allows to ignore the other table.

conda-store-server/conda_store_server/api.py
dcmcand: I disagree with that reasoning, but ultimately it is stylistic and not a big enough deal to worry too much about.

conda-store-server/conda_store_server/server/auth.py
dcmcand: I agree it is out of scope, but I think it is worth talking about. Maybe not in this PR, but we should probably add a new issue to discuss it.

dcmcand:
One thought that occurred to me is that introducing the version to our routes actually is breaking if anyone has previously had naked routes. I think adding a router for naked routes that redirects to v1 of the route and documenting this would handle that issue. This could be a seperate issue/pr if that is simpler, but without it this will be a breaking change.

nkaretnikov:
> One thought that occurred to me is that introducing the version to our routes actually is breaking if anyone has previously had naked routes.

Did you mean to reply to the router_ui thread? This PR doesn't make changes to existing routes.

nkaretnikov:
Yeah, no action is needed here. I'll mention this for whoever is reading these later. The API I'm using is already versioned:

```py
router_api = APIRouter(
    tags=["api"],
    prefix="/api/v1",
)
```

conda-store-server/conda_store_server/server/auth.py
costrouc: What does this mean? Does this mean that new installs will use (1)?

conda-store-server/tests/test_db_api.py
```diff
@@ -71,6 +71,108 @@ def test_namespace_role_mapping(db):
     NamespaceRoleMapping(namespace=namespace, namespace_id=namespace.id, entity="*/*")
 
 
+def test_namespace_role_mapping_v2(db):
```
costrouc, line 74: Like the detailed tests on this feature

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: It's the same syntax as everywhere else: 1 is the default. What it means in practice: old installs will work without any changes. They'll use the existing (old) table.

conda-store-server/tests/test_auth.py
```diff
@@ -140,36 +141,191 @@ def test_authorization(conda_store, entity_bindings, arn, permissions, authorize
...
-            },
+            primary_namespace=namespace,
+            # No default roles
+            role_bindings={},
         )
     )
 
     token_model = authentication.authenticate(token)
 
     authorization = RBACAuthorizationBackend(
```
dcmcand, line 165: Pulling all this auth stuff that is setup for the auth tests into a helper function would help readability of these tests imo. This won't block approval, but would help maintainability I think.

conda-store-server/tests/test_db_api.py
```diff
@@ -71,6 +71,108 @@ def test_namespace_role_mapping(db):
...
+    namespace_name = "pytest-namespace"
+    other_namespace_name1 = "pytest-other-namespace1"
+    other_namespace_name2 = "pytest-other-namespace2"
+    other_namespace_name3 = "pytest-other-namespace3"
+
+    # Starts with no namespaces
+    assert len(api.list_namespaces(db).all()) == 0
```
dcmcand, line 81: It might be simpler to just clear out the namespaces like you do in the auth tests than to assert here.

conda-store-server/tests/test_db_api.py
dcmcand: In general, relying on having a pure db and checking lengths has led to flaky tests for me in the past, especially if you ever try to run tests in parallel. I always prefer to check for concrete records than for lengths.

costrouc (approved):

PR #610, opened 2023-10-08
DOC - Add alembic example for postgres in docker

nkaretnikov:


nkaretnikov:
Had to do a migration, this seems to work with postgres and docker-compose.

@costrouc PTAL.

costrouc (approved):
LGTM thanks @nkaretnikov. I've always had a tricky time with the alembic migrations as well and what the correct thing to run is.

Issue #599, opened 2023-09-27
[BUG] - Review and improve docker images


pavithraes:
### Describe the bug

Currently, we have 2 Dockerfiles used to build images:
- [conda-store](https://github.com/conda-incubator/conda-store/blob/main/conda-store/Dockerfile), which does an editable install:
https://github.com/conda-incubator/conda-store/blob/64552a28e649d7305ec971332cfb739286769ec9/conda-store/Dockerfile#L22-L23
❗️this is also the file used during releases which should not be the case
- [conda-store-server](https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/Dockerfile), which installs a local directory:
https://github.com/conda-incubator/conda-store/blob/64552a28e649d7305ec971332cfb739286769ec9/conda-store-server/Dockerfile#L49-L51

These are used by the `docker-compose.yaml` in:
- the [root dir](https://github.com/conda-incubator/conda-store/blob/main/docker-compose.yaml), used by developers for local development
- the examples dir intended for end-users --❗️we should use released versions from PyPI or conda-forge here, or refer to the docker images built on releases
  - [`docker/docker-compose.yaml`](https://github.com/conda-incubator/conda-store/blob/main/examples/docker/docker-compose.yaml) - ❗️this file also has defined `redis` twice, which causes errors while building
  - [`docker-without-nfs/docker-compose.yaml`](https://github.com/conda-incubator/conda-store/blob/main/examples/docker-without-nfs/docker-compose.yaml)

☝🏽Additionally, the port number 5000, where the UI is deployed for development does not work reliably on MacOS. We need to pick default ports that work on all platforms.

### Expected behavior

We need to have separate images for building dev and stable versions of conda-store. The releases should use to stable versions.

### Anything else?

xref: #589, #566

pavithraes:
I think we can use the [`docker-compose.yaml` in conda-store-ui](https://github.com/conda-incubator/conda-store-ui/blob/main/docker-compose.yml), as is and as a ref, for the examples directory.

kcpevey:
@dcmcand it looks like the review is complete and all of the issues have been opened as separate issues. Can we close this in favor or the individual issues?

dcmcand:
Yes @kcpevey, we can close this one.  The only outstanding issue is #618. That should be a pretty easy one.

PR #621, opened 2023-10-12
update conda store dockerfile to have prod target

### LINKED ISSUE #614: [BUG] - conda-store image needs a release target and dev target in the Dockerfile
### Describe the bug

Split from #599 

[conda-store](https://github.com/conda-incubator/conda-store/blob/main/conda-store/Dockerfile), which does an editable install:

[conda-store/conda-store/Dockerfile](https://github.com/conda-incubator/conda-store/blob/64552a28e649d7305ec971332cfb739286769ec9/conda-store/Dockerfile#L22-L23)

Lines 22 to 23 in [64552a2](https://github.com/conda-incubator/conda-store/commit/64552a28e649d7305ec971332cfb739286769ec9)
 RUN cd /opt/conda-store && \ 
     pip install -e . 

❗️this is also the file used during releases which should not be the case

### Expected behavior

To fix this, refactor `conda-store/Dockerfile` to have 2 different build targets. 

1. The `dev` target should install dependencies in editable mode as the the current file does
2. The `release` target should not install dependencies in editable mode
3. The `.github/workflows/build_docker_image.yaml` workflow should be changed to use the `release` target

N/A

### Output

_No response_

_No response_

### Anything else?

_No response_

dcmcand:
Fixes #614
Part of #599 

## Description

updates Dockerfile for conda-store to have a prod target that does not install pip dependencies in editable mode, and a dev target that does.

This pull request:

- Update Dockerfile as described above
- Set github actions to use prod targets

## Pull request checklist

### DIFF
--- a/conda-store/Dockerfile
+++ b/conda-store/Dockerfile
@@ -28,7 +28,8 @@ USER conda-store
 WORKDIR /opt/jupyterhub
 
 FROM base as prod
+ARG RELEASE_VERSION
 RUN cd /opt/conda-store && \
-    pip install .
+    pip install conda-store==${RELEASE_VERSION} 
 USER conda-store
 WORKDIR /opt/jupyterhub
--- a/conda-store/Dockerfile
+++ b/conda-store/Dockerfile
@@ -21,15 +21,15 @@ ENV PATH=/opt/conda/condabin:/opt/conda/envs/conda-store/bin:/opt/conda/bin:/usr
 
 COPY ./ /opt/conda-store/
 
-FROM base as dev
-RUN cd /opt/conda-store && \
-    pip install -e .
-USER conda-store
-WORKDIR /opt/jupyterhub
-
 FROM base as prod
 ARG RELEASE_VERSION
 RUN cd /opt/conda-store && \
     pip install conda-store==${RELEASE_VERSION}
+USER conda-store
+WORKDIR /opt/jupyterhub
+
+FROM base as dev
+RUN cd /opt/conda-store && \
+    pip install -e .
 USER conda-store
 WORKDIR /opt/jupyterhub
--- a/conda-store/Dockerfile
+++ b/conda-store/Dockerfile
@@ -30,6 +30,6 @@ WORKDIR /opt/jupyterhub
 FROM base as prod
 ARG RELEASE_VERSION
 RUN cd /opt/conda-store && \
-    pip install conda-store==${RELEASE_VERSION} 
+    pip install conda-store==${RELEASE_VERSION}
 USER conda-store
 WORKDIR /opt/jupyterhub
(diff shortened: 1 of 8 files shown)


conda-store/Dockerfile
```diff
@@ -19,12 +21,14 @@ ENV PATH=/opt/conda/condabin:/opt/conda/envs/conda-store/bin:/opt/conda/bin:/usr
...
+WORKDIR /opt/jupyterhub
 
-RUN mkdir -p /opt/jupyterhub && \
-    chown -R 1000:1000 /opt/jupyterhub
-
-USER 1000:1000
-
+FROM base as prod
+RUN cd /opt/conda-store && \
+    pip install .
```
trallard, line 32: I'd prefer installing directly from PyPA than from the repository.

dcmcand: @trallard Do you mean to install the conda-store package from pypi when packing the docker release?

right now, docker images are built and pushed on every merge to main, and on every release. Using PyPi would mean that we should split the workflow, so pushes to main used this repo (since it isn't pushed to PyPi), and releases installed from PyPi, which would add a dependency in the order that the jobs run.

I am fine with that if we want to do it that way, I just want to make sure we are on the same page here.

trallard:
While reviewing the PR I realised that the GH workflow logic actually needs some tuning too. Right now the triggers are:

```yaml
on:
  push:
    branches:
      - "main"
  release:
    types: [created]
```

While this PR creates separate dev and prod Docker targets, the images are built more frequently than releases are made, and since conda-store is installed with `pip install .` this still does not ensure we are using an officially released version (which I'd prefer doing) so we might need to adapt to some logic where

* `dev` is built on pushes to main and optionally when a new version has been released on pip or conda-forge
* `prod` is built when a new version has been released on pip or conda-forge only

dcmcand:
> While reviewing the PR I realised that the GH workflow logic actually needs some tuning too. Right now the triggers are:
> 
> ```yaml
> on:
>   push:
>     branches:
>       - "main"
>   release:
>     types: [created]
> ```
> 
> While this PR creates separate dev and prod Docker targets, the images are built more frequently than releases are made, and since conda-store is installed with `pip install .` this still does not ensure we are using an officially released version (which I'd prefer doing) so we might need to adapt to some logic where
> 
>     * `dev` is built on pushes to main and optionally when a new version has been released on pip or conda-forge
> 
>     * `prod` is built when a new version has been released on pip or conda-forge only
> 
> 
> @dcmcand

Ok, no problem. I will make these changes and let you know when it is ready for :eyes:

dcmcand:
@trallard the same applies to conda-store-server correct?

trallard:
yes this should apply for both

dcmcand:
@trallard This is ready for another look. We can't test the the release workflow without cutting a release though. Can we cut a test release and then delete it from pypi after?

dcmcand:
@trallard is this one good to go?

trallard:
🤔  I got confused since the changes here are also in #625, so it seems some of my changes will not be introduced here (e.g. timezone) but in the other PR which is fine by me so will mark this as merge ready. 

As for the release - I will cut a release candidate today/tomorrow so that should work as a test

trallard (approved):

PR #622, opened 2023-10-13
Bug fix: use context manager for Sessions

### LINKED ISSUE #598: [BUG] - Submitting multiple builds can fill queue causing frontend to be stop working
### Describe the bug

I'll describe this issue as I've seen it through the Nebari jlab conda-store UI, but I believe the issue is likely the backend. 

Open an existing env in the UI, go to edit, make a minor change, click Create. Normally upon clicking Create, you'd be redirected back to the "non-edit" UI screen, but now clicking the button doesn't appear to do anything. 

Now attempt to go to the `/conda-store/admin` page - it will not load anything, just spin. 

Now shutdown your server, and log back in. Going to the conda-store UI and attempting to log in will have no effect, you just get the spinning logo in the browser tab. The rest of nebari remains operational, but there is no way to get conda-store out of this state immediately. 

image

This happened to me last night and when I came back this morning it had worked itself into an operational state again. I went through the above process this morning on one nebari deployment and the same thing happened. Then I went through the above process on a different nebari deployment and the same thing happened. 

I will also note that I tried to Delete environments and saw similar behavior but I'm not sure if that was because it was already in a broken state or if that also is causing something similar to happen. 

### Expected behavior

I expect to be able to Edit and Save environments. 

Reproducer explained above. 

### Output

_No response_

_No response_

### Anything else?

_No response_

nkaretnikov:
Fixes #598.

### DIFF
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import sys
+from contextlib import contextmanager
 from typing import Any, Dict
 
 import pydantic
@@ -360,6 +361,18 @@ class CondaStore(LoggingConfigurable):
             poolclass=QueuePool,
         )
         return self._session_factory
+
+    # Do not define this as a FastAPI dependency! That would cause Sessions
+    # not to be closed, which would lead to DB pool exhaustion and requests
+    # getting blocked.
+    # https://github.com/conda-incubator/conda-store/issues/598
+    @contextmanager
+    def get_db(self):
+        db = self.session_factory()
+        try:
+            yield db
+        finally:
+            db.close()
 
     @property
     def redis(self):
--- a/conda-store-server/conda_store_server/server/views/api.py
+++ b/conda-store-server/conda_store_server/server/views/api.py
@@ -8,7 +8,6 @@ from conda_store_server.schema import Permissions
 from conda_store_server.server import dependencies
 from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request
 from fastapi.responses import PlainTextResponse, RedirectResponse
-from sqlalchemy.orm import Session
 
 router_api = APIRouter(
     tags=["api"],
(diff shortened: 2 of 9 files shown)


conda-store-server/conda_store_server/app.py
```diff
@@ -352,14 +352,14 @@ def _docker_base_image(build: orm.Build):
...
     @property
     def session_factory(self):
-        if hasattr(self, "_session_factory"):
-            return self._session_factory
-
-        self._session_factory = orm.new_session_factory(
+        # Note: the Session object is not thread-safe, do not save or share it.
+        # Instead, create a new Session for each request.
+        # https://github.com/conda-incubator/conda-store/issues/598
+        # https://docs.sqlalchemy.org/en/20/orm/session_basics.html
```
nkaretnikov, line 358: Link to a page section here.

conda-store-server/conda_store_server/app.py
```diff
@@ -352,14 +352,14 @@ def _docker_base_image(build: orm.Build):
 
     @property
     def session_factory(self):
-        if hasattr(self, "_session_factory"):
-            return self._session_factory
-
-        self._session_factory = orm.new_session_factory(
```
nkaretnikov, line 358: Because this calls our orm wrapper, this also create a new engine each time. The engine needs to be created once.

nkaretnikov:
Note: I deliberately open this new context manager at the start of each request to avoid any unintended side-effects related to DB object lifetimes. This also make it easier to review since there are no logic changes. Later (and with more tests), this can be scoped better -- only when `db` is used.

I also make this context manager a method of `CondaStore`, to avoid refactoring of `session_factory`, like introducing globals. 

Also: this diff will be more readable if you hide whitespace changes in the GitHub UI:
Screen Shot 2023-10-15 at 13 44 03

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -538,50 +554,51 @@ async def api_get_specification(
...
-    )
+        default_namespace = (
+            entity.primary_namespace if entity else conda_store.default_namespace
+        )
 
-    try:
-        build_id = conda_store.register_environment(
-            db, specification, namespace_name, force=True
+        namespace_name = namespace or default_namespace
+
```
nkaretnikov, line 570: Remove.

nkaretnikov:
@costrouc PTAL.

costrouc (approved):
:heart: this change since it is even more explicit on the lifetime of a session. Sounds like in conclusion we cannot trust a fastapi dependency's lifetime to span the request. I like that we can keep the queuepool. LGTM and I was able to run this locally.

Also the reason we likely could not encounter this issue in standalone more with sqlite is that sqlite only allows one concurrent writer and a queuepool/connection is probably faked in sqlalchemy.

iameskild:
Amazing work @nkaretnikov! 🎉

PR #625, opened 2023-10-17
Dcmcand optimize docker files

dcmcand:
Follow on to #621 that should be merged before this is reviewed

## Description

This PR optimizes the Docker files for the conda-store and conda-store-server images. 

Currently the conda-store image is 1.54GB. This PR reduces that to  890MB by using multistage builds. This is a reduction of 43%.

The current conda-store-server image is 1.96GB. This PR reduces that to 1.24GB, a reduction of 37%.

ChatGPT

Smaller Docker images offer faster container deployment, reduced storage and bandwidth costs, and enhanced security by minimizing potential attack surfaces.

This pull request:

- Utilizes multistage builds to reduce the size of images
- Uses caches to speed the dev workflow and make rebuilding containers faster
- Removes unneeded files from images.

## Pull request checklist


conda-store-server/Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM --platform=linux/amd64 condaforge/mambaforge:23.3.1-1
+FROM --platform=linux/amd64 condaforge/mambaforge:23.3.1-1 as base
```
trallard, line 1: Do we need to specify the `PLATFORM` here? this is now a deprecated flag and in any case we can use the build action to make platforms explicit https://github.com/docker/setup-buildx-action/tree/v1/

conda-store-server/Dockerfile
```diff
@@ -1,7 +1,10 @@
...
 
 LABEL org.opencontainers.image.authors="conda-store development team"
 
+ENV PATH=/opt/conda/condabin:/opt/conda/envs/conda-store-server/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:${PATH}
+ENV TZ=America/New_York
```
trallard, line 6: Is this the same timezone we use for conda-store general config? I thought we were using UTC

conda-store-server/Dockerfile
```diff
@@ -38,19 +41,25 @@ USER 1000:1000
 
 COPY environment.yaml /opt/conda-store-server/environment.yaml
 
-RUN mamba env create -f /opt/conda-store-server/environment.yaml
+RUN mamba env create -f /opt/conda-store-server/environment.yaml && conda clean  --force-pkgs-dirs
```
trallard, line 44:
```suggestion
RUN mamba env create -f /opt/conda-store-server/environment.yaml && \
	conda clean  --force-pkgs-dirs
```

readability nit

conda-store/Dockerfile
```diff
@@ -1,30 +1,35 @@
...
+    rm -rf /tmp/*   &&\
+    groupadd -g 1000 conda-store &&\
+    useradd -M -r -s /usr/sbin/nologin -u 1000 -g 1000  conda-store && \
+    mkdir -p /opt/jupyterhub && \
+    chown -R conda-store:conda-store /opt/jupyterhub
 
 COPY environment.yaml /opt/conda-store/environment.yaml
 
-RUN mamba env create -f /opt/conda-store/environment.yaml
+RUN mamba env create -f /opt/conda-store/environment.yaml && mamba clean --all -y && conda clean --force-pkgs-dirs
```
trallard, line 18:
```suggestion
RUN mamba env create -f /opt/conda-store/environment.yaml &&\
	mamba clean --all -y &&\
	conda clean --force-pkgs-dirs
```

readability nit

trallard (commented):
Neat I love some image optimisation 🎉 It seems now that we are duplicating some bits in the `build_docker_image.yaml` and the `release.yaml` to keep these aligned and reduce duplication. Shall we move these tidbits into a reusable action and call it from the workflows instead?

conda-store-server/Dockerfile
dcmcand: No, we don't. I was a considering removing it, but don't like bundling too many changes into a PR, so wasn't sure if that change belonged here. I will remove it though.

conda-store-server/Dockerfile
dcmcand: This is what it had. I think UTC would make more sense personally, but I didn't want to change it without having context around why it was set to that. Would you like me to change it?

dcmcand:
> Neat I love some image optimisation 🎉 It seems now that we are duplicating some bits in the `build_docker_image.yaml` and the `release.yaml` to keep these aligned and reduce duplication. Shall we move these tidbits into a reusable action and call it from the workflows instead?

I think that makes sense. I will make another issue and do that in a separate PR.

conda-store/Dockerfile
dcmcand: :heavy_check_mark:

nkaretnikov:
FYI: integration test failure is unrelated, I'm investigating: https://github.com/conda-incubator/conda-store/issues/632

trallard (approved):
This looks good so will go ahead and merge and we can see how these changes look with the release candidate tomorrow 🤞🏽 
Thanks @dcmcand

PR #636, opened 2023-10-20
DEV -  Add build check and test release workflow

trallard:
## Description

This pull request:

- Ensures all merges to main are testes for build and release

## Pull request checklist


dcmcand (approved):
I like doing the test release for every push to main. That really helps to verify that we can deploy anytime we need to.

PR #639, opened 2023-10-21
Store the state and database files in ~/.conda-store by default

nkaretnikov:
This work is part of #513. Original PR: #554.

### DIFF
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -17,6 +17,7 @@ from conda_store_server import (
     storage,
     utils,
 )
+from conda_store_server import CONDA_STORE_DIR
 from sqlalchemy.orm import Session
 from sqlalchemy.pool import QueuePool
 from traitlets import (
@@ -71,9 +72,6 @@ def conda_store_validate_action(
         raise utils.CondaStoreError(
             f"`CondaStore.storage_threshold` reached. Action {action.value} prevented due to insufficient storage space"
         )
-
-
-CONDA_STORE_DIR = Path.home() / ".conda-store"
 
 
 class CondaStore(LoggingConfigurable):
@@ -205,7 +203,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     database_url = Unicode(
-        "sqlite:///" + str(PurePosixPath(CONDA_STORE_DIR) / "conda-store.sqlite"),
+        "sqlite:///" + str(CONDA_STORE_DIR / "conda-store.sqlite"),
         help="url for the database. e.g. 'sqlite:///conda-store.sqlite' tables will be automatically created if they do not exist",
         config=True,
     )
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -73,7 +73,7 @@ def conda_store_validate_action(
         )
 
 
-CONDA_STORE_DIR = (Path.home() / ".conda-store",)
+CONDA_STORE_DIR = Path.home() / ".conda-store"
 
 
 class CondaStore(LoggingConfigurable):
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -2,6 +2,7 @@ import datetime
 import os
 import sys
 from contextlib import contextmanager
+from pathlib import Path, PosixPath
 from typing import Any, Dict
 
 import pydantic
@@ -88,7 +89,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     store_directory = Unicode(
-        "conda-store-state",
+        str(Path.home() / ".conda-store" / "conda-store-state"),
         help="directory for conda-store to build environments and store state",
         config=True,
     )
@@ -201,7 +202,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     database_url = Unicode(
-        "sqlite:///conda-store.sqlite",
+        "sqlite:///" + str(PosixPath.home()) + "/.conda-store/conda-store.sqlite",
         help="url for the database. e.g. 'sqlite:///conda-store.sqlite' tables will be automatically created if they do not exist",
         config=True,
     )
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -207,7 +207,7 @@ class CondaStore(LoggingConfigurable):
     database_url = Unicode(
         "sqlite:///"
         + str(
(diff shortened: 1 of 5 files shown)


nkaretnikov:
This should land only once Windows lands, to make sure this is compatible.

conda-store-server/conda_store_server/app.py
```diff
@@ -72,6 +73,9 @@ def conda_store_validate_action(
         )
 
 
+CONDA_STORE_DIR = Path.home() / ".conda-store"
```
nkaretnikov, line 76: This needs to be moved to storage (to avoid import errors) and used there in `storage_path` as well.

conda-store-server/conda_store_server/app.py
```diff
@@ -201,7 +205,7 @@ class CondaStore(LoggingConfigurable):
     )
 
     database_url = Unicode(
-        "sqlite:///conda-store.sqlite",
+        "sqlite:///" + str(PurePosixPath(CONDA_STORE_DIR) / "conda-store.sqlite"),
```
nkaretnikov, line 208: This needs to be verified and used everywhere.

```py
>>> str(PurePosixPath(CONDA_STORE_DIR) / "foo")
'C:\\/Users/nikita/.conda-store/foo'

>>> str(CONDA_STORE_DIR / "foo")
'C:\\Users\\nikita\\.conda-store\\foo'

>>> str(PosixPath(CONDA_STORE_DIR) / "foo")
NotImplementedError: cannot instantiate 'PosixPath' on your system
```

Here people say that it works fine with Windows-style paths:
https://stackoverflow.com/questions/19260067/sqlalchemy-engine-absolute-path-url-in-windows

conda-store-server/conda_store_server/server/app.py
```diff
@@ -179,6 +179,14 @@ def initialize(self, *args, **kwargs):
 
         self.conda_store = CondaStore(parent=self, log=self.log)
 
+        self.conda_store.ensure_directories()
+        self.log.info(
+            f"Running conda-store with database: {self.conda_store.database_url}"
+        )
+        self.log.info(
+            f"Running conda-store with storage: {self.conda_store.store_directory}"
```
nkaretnikov, line 187: Replace `storage` with `store`. There's a different param called storage.

conda-store-server/tests/conftest.py
```diff
@@ -14,19 +14,26 @@
...
+    filename = tmp_path / ".conda-store" / "database.sqlite"
+
+    store_directory = tmp_path / ".conda-store" / "state"
+    store_directory.mkdir(parents=True)
 
     with utils.chdir(tmp_path):
         yield Config(
             CondaStore=dict(
+                store_directory=str(store_directory),
                 database_url=f"sqlite:///{filename}?check_same_thread=False"
```
nkaretnikov, line 37: This needs to be rebased on the Windows PR, to make sure these work on the Windows builder.

nkaretnikov:
Rebased on the Windows PR to be able to test on Windows, will remove redundant commits later.

conda-store-server/tests/conftest.py
```diff
@@ -14,19 +14,26 @@
 @pytest.fixture
 def celery_config(conda_store):
     config = conda_store.celery_config
-    config["traitlets"] = {"CondaStore": {"database_url": conda_store.database_url}}
+    config["traitlets"] = {"CondaStore": {
+        "database_url": conda_store.database_url,
+        "store_directory": conda_store.store_directory,
+    }}
```
nkaretnikov, line 20: Set `beat_schedule_filename` and `storage` here as well.

conda-store-server/tests/conftest.py
```diff
@@ -7,26 +7,37 @@
...
 def conda_store_config(tmp_path, request):
     from traitlets.config import Config
 
-    filename = pathlib.Path(tmp_path) / "database.sqlite"
+    filename = tmp_path / ".conda-store" / "database.sqlite"
+
+    store_directory = tmp_path / ".conda-store" / "state"
+    store_directory.mkdir(parents=True)
+
+    storage.LocalStorage.storage_path = str(tmp_path / "storage")
```
nkaretnikov, line 34: Needs `".conda-store"` here as well.

nkaretnikov:
Contents of `CONDA_STORE_DIR`:

```
% ls ~/.conda-store
celerybeat-schedule.bak  celerybeat-schedule.dat  celerybeat-schedule.dir  conda-store.sqlite  state  storage

% ls ~/.conda-store/state
default

% ls ~/.conda-store/state/default
a54148c1d1d435985989ea232bc982f0eb877991f6c70c4de1390dd76e7ef564-20231025-184722-649393-2-test-env  envs

ls ~/.conda-store/storage
archive  docker  lockfile  logs  yaml
```

nkaretnikov:
Tested on Windows and Linux with this env. No files are created in current dir anymore, as expected.

```yaml
channels:
- conda-forge
dependencies:
- python
- pip:
  - nothing
- ipykernel
- pytest
- requests
description: ''
name: test-env
prefix: null
variables: null
```

conda-store-server/conda_store_server/app.py
```diff
@@ -434,6 +435,7 @@ def celery_config(self):
                     "kwargs": {},
                 },
             },
+            "beat_schedule_filename": f"{CONDA_STORE_DIR}/celerybeat-schedule",
```
nkaretnikov, line 438: This needs to use `/` and then convert to str.

nkaretnikov:
This change shouldn't affect Nebari since it uses a config file, which sets `database_url`, `storage_class`, and `store_directory`.

conda-store-server/conda_store_server/app.py
asmeurer: FYI, if this is related to `--beat`, that's disabled on Windows because Celery doesn't support it on Windows (I'm not clear what functionality we're losing because of that).

costrouc (approved):
LGTM. I don't think there is anything controvertial in this PR and I was able to run it locally and have it use `~/.conda-store/...`

nkaretnikov:
@jaimergp I'm not quite sure that is better. Also, on Windows, I see this:
Screen Shot 2023-11-01 at 08 50 58

Conda and vscode (which is developed by MS) use `/Users/<User>`.

jaimergp:
`.conda` needs to change too but that ship sailed too long ago and people's environments may live there, so it's tricky to change now. `.vscode`'s issue is inherited from Electron ([see lengthy thread](https://github.com/microsoft/vscode/issues/3884)). We don't have a good excuse I'm afraid :D

nkaretnikov:
Yeah, I've changed my mind after reading relevant threads from other projects. We should use well-accepted standards.

PR #640, opened 2023-10-22
Windows support

nkaretnikov:
See #507. Original PR: #559.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -19,13 +19,14 @@ jobs:
     name: "unit-test conda-store-server"
     strategy:
       matrix:
-        # cannot run on windows due to needing fake-chroot for conda-docker
-        os: ["ubuntu", "macos"]
+        os: ["ubuntu", "macos", "windows"]
         include:
           - os: ubuntu
             environment-file: conda-store-server/environment-dev.yaml
           - os: macos
             environment-file: conda-store-server/environment-macos-dev.yaml
+          - os: windows
+            environment-file: conda-store-server/environment-windows-dev.yaml
     runs-on: ${{ matrix.os }}-latest
     defaults:
       run:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -52,7 +52,7 @@ jobs:
 
       - name: "Unit tests ✅"
         run: |
-          pytest tests
+          pytest -vvv tests
 
   integration-test-conda-store-server:
     name: "integration-test conda-store-server"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -62,7 +62,7 @@ jobs:
 
       - name: "Unit tests ✅"
         run: |
-          pytest -vvv tests
+          pytest tests
 
   integration-test-conda-store-server:
     name: "integration-test conda-store-server"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,15 +44,15 @@ jobs:
 
       - name: "Linting Checks 🧹"
         run: |
-          python -m hatch env run -e dev lint
+          hatch env run -e dev lint
 
       - name: "Build package 📦"
         run: |
-          python -m hatch build
+          hatch build
 
       - name: "Unit tests ✅"
         run: |
-          python -m pytest tests
+          pytest tests
 
   integration-test-conda-store-server:
     name: "integration-test conda-store-server"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -42,6 +42,16 @@ jobs:
             environment-file: ${{ matrix.environment-file }}
             miniforge-version: latest
 
+      # This fixes a "DLL not found" issue importing ctypes from the hatch env
+      - name: Reinstall Python 3.10 on Windows runner
+        uses: nick-fields/retry@v2.8.3
+        with:
+          timeout_minutes: 9999
+          max_attempts: 6
+          command:
+            conda install --channel=conda-forge --quiet --yes python=${{ matrix.python }}
+        if: matrix.os == 'windows'
+
       - name: "Linting Checks 🧹"
         run: |
           hatch env run -e dev lint
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,15 +44,15 @@ jobs:
 
       - name: "Linting Checks 🧹"
         run: |
-          hatch env run -e dev lint
+          python -m hatch env run -e dev lint
 
       - name: "Build package 📦"
         run: |
-          hatch build
+          python -m hatch build
 
       - name: "Unit tests ✅"
         run: |
-          pytest tests
+          python -m pytest tests
 
   integration-test-conda-store-server:
     name: "integration-test conda-store-server"
--- a/conda-store-server/conda_store_server/action/base.py
+++ b/conda-store-server/conda_store_server/action/base.py
@@ -18,7 +18,7 @@ def action(f: typing.Callable):
             # redirect stdout -> action_context.stdout
             stack.enter_context(contextlib.redirect_stdout(action_context.stdout))
 
-            # redirect stderr -> action_context.stdout
+            # redirect stderr -> action_context.stderr
             stack.enter_context(contextlib.redirect_stderr(action_context.stdout))
 
             # create a temporary directory
@@ -38,6 +38,7 @@ class ActionContext:
     def __init__(self):
         self.id = str(uuid.uuid4())
         self.stdout = io.StringIO()
+        self.stderr = io.StringIO()
         self.log = logging.getLogger(f"conda_store_server.action.{self.id}")
         self.log.propagate = False
         self.log.addHandler(logging.StreamHandler(stream=self.stdout))
@@ -45,13 +46,15 @@ class ActionContext:
         self.result = None
         self.artifacts = {}
 
-    def run(self, *args, **kwargs):
+    def run(self, *args, redirect_stderr=True, **kwargs):
         result = subprocess.run(
             *args,
             **kwargs,
             stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
+            stderr=subprocess.STDOUT if redirect_stderr else subprocess.PIPE,
             encoding="utf-8",
         )
         self.stdout.write(result.stdout)
+        if not redirect_stderr:
+            self.stderr.write(result.stderr)
         return result
--- a/conda-store-server/conda_store_server/action/base.py
+++ b/conda-store-server/conda_store_server/action/base.py
@@ -18,7 +18,7 @@ def action(f: typing.Callable):
             # redirect stdout -> action_context.stdout
             stack.enter_context(contextlib.redirect_stdout(action_context.stdout))
 
-            # redirect stderr -> action_context.stderr
+            # redirect stderr -> action_context.stdout
             stack.enter_context(contextlib.redirect_stderr(action_context.stdout))
 
             # create a temporary directory
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -32,8 +32,8 @@ from traitlets import (
 )
 from traitlets.config import LoggingConfigurable
 
-
 ON_WIN = sys.platform.startswith("win")
+
 
 def conda_store_validate_specification(
     db: Session,
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -32,8 +32,6 @@ from traitlets import (
 )
 from traitlets.config import LoggingConfigurable
 
-ON_WIN = sys.platform.startswith("win")
-
 
 def conda_store_validate_specification(
     db: Session,
@@ -304,21 +302,21 @@ class CondaStore(LoggingConfigurable):
     )
 
     default_uid = Integer(
-        None if ON_WIN else os.getuid(),
+        None if sys.platform == "win32" else os.getuid(),
         help="default uid to assign to built environments",
         config=True,
         allow_none=True,
     )
 
     default_gid = Integer(
-        None if ON_WIN else os.getgid(),
+        None if sys.platform == "win32" else os.getgid(),
         help="default gid to assign to built environments",
         config=True,
         allow_none=True,
     )
 
     default_permissions = Unicode(
-        None if ON_WIN else "775",
+        None if sys.platform == "win32" else "775",
         help="default file permissions to assign to built environments",
         config=True,
         allow_none=True,
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -33,6 +33,8 @@ from traitlets import (
 from traitlets.config import LoggingConfigurable
 
 
+ON_WIN = sys.platform.startswith("win")
+
 def conda_store_validate_specification(
     db: Session,
     conda_store: "CondaStore",
@@ -302,21 +304,24 @@ class CondaStore(LoggingConfigurable):
     )
 
     default_uid = Integer(
-        os.getuid(),
+        None if ON_WIN else os.getuid(),
         help="default uid to assign to built environments",
         config=True,
+        allow_none=True,
     )
 
     default_gid = Integer(
-        os.getgid(),
+        None if ON_WIN else os.getgid(),
         help="default gid to assign to built environments",
         config=True,
+        allow_none=True,
     )
 
     default_permissions = Unicode(
-        "775",
+        None if ON_WIN else "775",
         help="default file permissions to assign to built environments",
         config=True,
+        allow_none=True,
     )
 
     default_docker_base_image = Union(
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -517,7 +517,7 @@ class CondaChannel(Base):
             batch_size = 990
             all_package_keys = list(package_builds.keys())
             for i in range(0, len(all_package_keys), batch_size):
-                logger.info(f"handling subset at index {i} (batch size {batch_size}")
+                logger.info(f"handling subset at index {i} (batch size {batch_size})")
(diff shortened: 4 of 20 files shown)


nkaretnikov:
Removed some original CI changes while rebasing, since main uses miniforge.

.github/workflows/tests.yaml
```diff
@@ -51,7 +52,7 @@ jobs:
 
       - name: "Unit tests ✅"
         run: |
-          pytest tests
+          pytest -vvv tests
```
nkaretnikov, line 55: Remove.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -335,6 +336,10 @@ def start(self):
 
         # start worker if in standalone mode
         if self.standalone:
+            address = "localhost" if self.address == "0.0.0.0" else self.address
+            print(
+                f"Starting standalone conda-store server at http://{address}:{self.port}"
+            )
```
nkaretnikov, line 342: Remove.

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,52 @@ def chdir(directory: pathlib.Path):
...
+            if st.st_ino in have:
+                continue
+            have.add(st.st_ino)
+            apparent_total_bytes += st.st_size
+        for d in dirnames:
+            dp = os.path.join(dirpath, d)
+            if os.path.islink(dp):
+                apparent_total_bytes += os.lstat(dp).st_size
+
+    return apparent_total_bytes
```
nkaretnikov, line 83: Needs clean up.

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,52 @@ def chdir(directory: pathlib.Path):
...
         cmd = ["du", "-sb", str(path)]
+    else:
+        return str(du(path))
 
-    return subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    output = subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    if sys.platform == "darwin":
+        # mac du does not have the -b option to return bytes
+        output = str(int(output) * 512)
+    return output
```
nkaretnikov, line 98: Move macOS code to a separate PR.

conda-store-server/environment-windows-dev.yaml
```diff
@@ -0,0 +1,56 @@
+name: conda-store-server-dev
+channels:
+  - conda-forge
+  - Microsoft
+dependencies:
```
nkaretnikov, line 5: Needs to match other dev envs on main.

conda-store-server/environment-windows-dev.yaml
```diff
@@ -0,0 +1,56 @@
+name: conda-store-server-dev
```
nkaretnikov, line 1: Needs to match other dev envs on main.

.github/workflows/tests.yaml
```diff
@@ -19,13 +19,14 @@ jobs:
...
-        # cannot run on windows due to needing fake-chroot for conda-docker
-        os: ["ubuntu", "macos"]
+        os: ["ubuntu", "macos", "windows"]
         include:
           - os: ubuntu
             environment-file: conda-store-server/environment-dev.yaml
           - os: macos
             environment-file: conda-store-server/environment-macos-dev.yaml
+          - os: windows
+            environment-file: conda-store-server/environment-windows-dev.yaml
```
nkaretnikov, line 29: Check that this works. Might need additional changes.

conda-store-server/tests/test_usage.py
```diff
@@ -0,0 +1,32 @@
...
+    # on Mac and Linux but uses the Python du() function on Windows.
+    val = disk_usage(test_file)
+    assert isinstance(val, str)
+    assert 1000 <= int(val) <= 1500
+    assert 1000 <= du(test_file) <= 1500
+
+    val = disk_usage(test_dir)
+    assert isinstance(val, str)
+    assert 2000 + dir_size <= int(val) <= 2700 + dir_size
+    assert 2000 + dir_size <= du(test_dir) <= 2700 + dir_size
```
nkaretnikov, line 32: This could use more tests. Bounds are 50% of file size, too.

conda-store-server/tests/test_actions.py
```diff
@@ -155,6 +164,7 @@ def test_remove_conda_prefix(tmp_path, simple_conda_lock):
     assert not conda_prefix.exists()
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="permissions are not supported on Windows")
```
nkaretnikov, line 167: Use the same ON_WIN check here.

nkaretnikov:
Note: Windows test takes 25 min. macOS: 14 min. Ubuntu: 10 min.

conda-store-server/conda_store_server/utils.py
nkaretnikov: This is needed for the du test to pass on macOS CI runner.

conda-store-server/tests/test_actions.py
nkaretnikov: I'll do the other way around and use the win32 check instead of the constant everywhere.

conda-store-server/conda_store_server/app.py
```diff
@@ -302,21 +304,24 @@ def _default_celery_results_backend(self):
     )
 
     default_uid = Integer(
-        os.getuid(),
+        None if ON_WIN else os.getuid(),
```
nkaretnikov, line 307: Use sys.platform here instead of the constant.

conda-store-server/conda_store_server/schema.py
```diff
@@ -194,20 +196,20 @@ class Settings(BaseModel):
         metadata={"global": True},
     )
 
-    default_uid: int = Field(
-        os.getuid(),
+    default_uid: int | None = Field(
+        None if ON_WIN else os.getuid(),
```
nkaretnikov, line 200: Use sys.platform here instead of the constant.

conda-store-server/conda_store_server/schema.py
```diff
@@ -194,20 +196,20 @@ class Settings(BaseModel):
         metadata={"global": True},
     )
 
-    default_uid: int = Field(
-        os.getuid(),
+    default_uid: int | None = Field(
```
nkaretnikov, line 199: Use Optional instead of `| None`.

conda-store-server/tests/test_usage.py
nkaretnikov: Bounds:

```
ubuntu

2023-10-22T18:43:40.3618723Z tests/test_usage.py XXXXXXXXXXXXXX: 1000 1000
2023-10-22T18:43:40.3619644Z XXXXXXXXXXXXXX: 4096 6096 6096

macos

2023-10-22T18:49:11.6388490Z tests/test_usage.py XXXXXXXXXXXXXX: 1024 1000
2023-10-22T18:49:11.6389010Z XXXXXXXXXXXXXX: 64 2560 2160

windows

2023-10-22T18:56:13.8420208Z tests\test_usage.py XXXXXXXXXXXXXX: 1000 1000
2023-10-22T18:56:13.8420808Z XXXXXXXXXXXXXX: 0 2000 2000


```

nkaretnikov: I'll keep this as is right now, but it might be worth it to revise later. Especially, on macOS.

conda-store-server/conda_store_server/action/base.py
```diff
@@ -18,7 +18,7 @@ def wrapper(*args, **kwargs):
             # redirect stdout -> action_context.stdout
             stack.enter_context(contextlib.redirect_stdout(action_context.stdout))
 
-            # redirect stderr -> action_context.stdout
+            # redirect stderr -> action_context.stderr
```
nkaretnikov, line 21: Comment doesn't match code. Revert this.

conda-store-server/tests/test_actions.py
```diff
@@ -15,9 +15,17 @@ def test_action_decorator():
...
+            # echo is not a separate program on Windows
+            context.run(["cmd", "/c", "echo subprocess"])
+            context.run("echo subprocess_stdout", shell=True)
+            context.run("echo subprocess_stderr>&2", shell=True)
+            context.run("echo subprocess_stderr_no_redirect>&2", shell=True, redirect_stderr=False)
+        else:
+            context.run(["echo", "subprocess"])
+            context.run("echo subprocess_stdout", shell=True)
+            context.run("echo subprocess_stderr 1>&2", shell=True)
+            context.run("echo subprocess_stderr_no_redirect 1>&2", shell=True, redirect_stderr=False)
```
nkaretnikov, line 28: Only the first command needs to be inside if-else.

conda-store-server/tests/test_usage.py
```diff
@@ -0,0 +1,32 @@
...
+    dir_size = du(test_dir)
+    assert abs(dir_size - int(disk_usage(test_dir))) <= 1000
+
+    test_file = test_dir / "test_file"
+    test_file.write_text("a"*1000)
+    test_file2 = test_dir / "test_file2"
+    test_file2.write_text("b"*1000)
+    # Test hard links
+    test_file_hardlink = test_dir / "test_file_hardlink"
+    test_file_hardlink.hardlink_to(test_file)
```
nkaretnikov, line 19: Note for reviewers: this is tested where test_dir size is tested as a whole.

conda-store-server/conda_store_server/action/base.py
```diff
@@ -38,20 +38,23 @@ class ActionContext:
...
             **kwargs,
             stdout=subprocess.PIPE,
-            stderr=subprocess.STDOUT,
+            stderr=subprocess.STDOUT if redirect_stderr else subprocess.PIPE,
             encoding="utf-8",
         )
         self.stdout.write(result.stdout)
+        if not redirect_stderr:
+            self.stderr.write(result.stderr)
         return result
```
nkaretnikov, line 60: See https://github.com/conda-incubator/conda-store/pull/559#issuecomment-1756298903 for context on redirects.

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -436,7 +436,9 @@ async def post_login_method(
             samesite="strict",
             domain=self.cookie_domain,
             # set cookie to expire at same time as jwt
-            max_age=(authentication_token.exp - datetime.datetime.utcnow()).seconds,
+            max_age=int(
+                (authentication_token.exp - datetime.datetime.utcnow()).total_seconds()
+            ),
```
nkaretnikov, line 441: See https://github.com/conda-incubator/conda-store/pull/559/files#r1364084772 for context on this.

conda-store-server/tests/test_actions.py
nkaretnikov: Nope, this is parsed slightly differently on Windows. Whitespace would be included into the output with redirect_stderr. So this is fine.

docs/user_guide.md
```diff
@@ -73,6 +73,10 @@ conda-unpack
 
 ### Docker Registry
 
+```{note}
```
nkaretnikov, line 76: Not sure if {note} syntax is supported.

nkaretnikov:
Manually tested this env via the admin UI:

```
channels:
- conda-forge
dependencies:
- python
- pip:
  - nothing
- ipykernel
- pytest
- requests
description: ''
name: test-env
prefix: null
variables: null
```

Building, rebuilding, and deletion work.

nkaretnikov:
@costrouc This is Aaron's Windows PR with some minor changes from me. PTAL and merge if OK.

jaimergp:
> the post set up env on Windows takes 3 mins

You can do `run-post: false` (or similar name) on setup-miniconda. It's an ephemeral VM after all.

conda-store-server/conda_store_server/orm.py
```diff
@@ -513,10 +513,11 @@ def update_packages(self, db, subdirs=None):
                 package_builds[package_key].append(new_package_build_dict)
             logger.info("CondaPackageBuild objects created")
 
-            batch_size = 1000
+            # sqlite3 has a max expression depth of 1000
+            batch_size = 990
```
costrouc, line 517: I bet there is an interesting story here? Is this only something that showed up on Windows?

conda-store-server/conda_store_server/app.py
```diff
@@ -302,21 +302,24 @@ def _default_celery_results_backend(self):
     )
 
     default_uid = Integer(
-        os.getuid(),
+        None if sys.platform == "win32" else os.getuid(),
```
costrouc, line 305: I see `sys.platform == "win32"` in many places. Should we have a constant that we create for this? Since much of our code only has a special path for Windows?

conda-store-server/conda_store_server/server/app.py
```diff
@@ -198,9 +199,9 @@ def trim_slash(url):
         app = FastAPI(
             title="conda-store",
             version=__version__,
-            openapi_url=os.path.join(self.url_prefix, "openapi.json"),
-            docs_url=os.path.join(self.url_prefix, "docs"),
-            redoc_url=os.path.join(self.url_prefix, "redoc"),
+            openapi_url=posixpath.join(self.url_prefix, "openapi.json"),
```
costrouc, line 202: I'll admit using `os.path...` for urls is a hack and I should have used something like yarl for example. Thoughts? Since url paths are not the same as file paths? Maybe posixpath is the right thing to use here. It certainly seems better than os.path.

conda-store-server/conda_store_server/server/auth.py
costrouc: A fun bug :smile: glad we got to the bottom of this one.

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,44 @@ def chdir(directory: pathlib.Path):
...
+
+
 def disk_usage(path: pathlib.Path):
     if sys.platform == "darwin":
         cmd = ["du", "-sAB1", str(path)]
-    else:
+    elif sys.platform == "linux":
         cmd = ["du", "-sb", str(path)]
+    else:
+        return str(du(path))
```
costrouc, line 84: Good to see the fallback is this approach :+1: If they are using cygwin or https://learn.microsoft.com/en-us/sysinternals/downloads/du it may still be possible they have du.

conda-store-server/conda_store_server/utils.py
```diff
@@ -50,13 +50,44 @@ def chdir(directory: pathlib.Path):
...
 
-    return subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    output = subprocess.check_output(cmd, encoding="utf-8").split()[0]
+    if sys.platform == "darwin":
+        # mac du does not have the -b option to return bytes
+        output = str(int(output) * 512)
```
costrouc, line 89: Why 512? This is a weird size since it is a half a kb.

conda-store-server/conda_store_server/worker/app.py
```diff
@@ -71,9 +71,18 @@ def start(self):
         argv = [
             "worker",
             "--loglevel=INFO",
-            "--beat",
         ]
 
+        # The default Celery pool requires this on Windows. See
+        # https://stackoverflow.com/questions/37255548/how-to-run-celery-on-windows
```
costrouc, line 77: Great glad this fix was minimal.

conda-store-server/tests/test_usage.py
costrouc: > The exact values depend on the block size and other details

Oh yeah this is a fun detail :smile: guess you can't be smaller than the blocksize and also can only take up an integer number of block sizes.

costrouc (approved):
Overall these changes do not look controversial and I'm surprised that the windows fixes were actually smaller than I expected. I was really expecting issues with pathlib etc but it looks like most of those bugs didn't exist.

conda-store-server/conda_store_server/app.py
costrouc: More I think about it I don't think this is necissary

conda-store-server/conda_store_server/orm.py
nkaretnikov: I haven't seen this personally, but it's mentioned in one of the Aaron's commits:

```
Lower the batch size for updating packages from 1000 to 990
1000 hits the sqlite3 max expression depth and results in an error.
```

Seems harmless, so I decided to keep it.

conda-store-server/conda_store_server/app.py
nkaretnikov: There was a constant in Aaron's original code. Well, two constants, in different files. Since you need to share this between multiple files, you'd have to put it in `__init__` to avoid import issues. But we don't do this for any other platforms, so I decided to simply use `sys.platform`, as everywhere else, to avoid introducing new entities.

conda-store-server/conda_store_server/server/app.py
nkaretnikov: A separate entity for URLs would be better, yes. But I'm not going to block this PR just because of this issue, since everything seems to work fine.

conda-store-server/conda_store_server/utils.py
nkaretnikov: We've already looked at du from sysinternals. Besides being a separate tool, it also produces slightly different results. And the output format is also different. As for cygwin, the context for this work is to be able to use native Windows tooling.

conda-store-server/conda_store_server/utils.py
nkaretnikov: A comment from a discussion on the original PR: "Note that the numbers will be off a little bit (by less than 512) on Mac because du rounds up to the nearest multiple of 512 when converting bytes to blocks."

I'm still not certain that macOS code is correct here, but it does pass this new test (where we check all three platforms), so I'm going to keep this and will take a look later.

conda-store-server/conda_store_server/worker/app.py
nkaretnikov: It's still not clear whether it means that Celery works as intended, since it's not officially supported on Windows. But it does run, so we'll just use this for now.

PR #641, opened 2023-10-22
Fix du on macOS

nkaretnikov:


nkaretnikov:
Split out from https://github.com/conda-incubator/conda-store/pull/559.

nkaretnikov:
Nevermind, I see why it was necessary there. There's a new test that runs du across Linux, macOS, and Windows. Without this, macOS runner fails.

PR #642, opened 2023-10-27
replace port 5000 with port 8080

### LINKED ISSUE #618: [BUG] - port 5000 is a problem on MacOS
### Describe the bug

Split from #599 
References from #566

MacOS uses port 5000 for airplay with new versions of MacOS. This means that running conda-store-server on port 5000 does not work for many users.

### Expected behavior

Conda store server works on the default port for all major OS's

Try to run conda-store-server on a recent MacOS machine. 

### Output

It will fail with a message saying that port 5000 is already in use.

_No response_

### Anything else?

_No response_

dcmcand:
Fixes #618  .

## Description
Replaces port 5000 with port 8080 as the default port to run on. Port 8080 is not a privileged port, but is an alternate port for HTTP. Port 5000 is in use on some Mac computers which causes conflicts.

This pull request:

- Replaces port 5000 with port 8080 as default port.

## Pull request checklist

- [X] Did you update the documentation (if required)?
- [X] Did you add/update relevant tests for this change (if required)?

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -93,7 +93,7 @@ jobs:
 
           wait-for-it localhost:5432 # postgresql
           wait-for-it localhost:9000 # minio
-          wait-for-it localhost:5000 # conda-store-server
+          wait-for-it localhost:8080 # conda-store-server
 
       - name: "Run Playwright tests 🎭"
         run: |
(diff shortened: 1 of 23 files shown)


costrouc (approved):
This is LONG overdue :smile:. The 5000 port is used for airpods.

costrouc:
This may cause a breaking change in the client's etc. assuming port 5000 by default but I'm unsure.

kcpevey:
> This may cause a breaking change in the client's etc. assuming port 5000 by default but I'm unsure.

We'll need to make a note in the release notes.

nkaretnikov:
Have you tested this with conda-store-ui? For me, this breaks the login process (when you click on the Login button) when using conda-store-ui. Because the UI assumes port 5000 for conda-store. I'll open an issue on the UI side.

PR #647, opened 2023-10-30
Change `url_prefix` to `/conda-store`

### LINKED ISSUE #646: [BUG] - Change defaults so that standalone runs with the UI without the config
### Describe the bug

While investigating https://github.com/conda-incubator/conda-store-ui/issues/322, I discovered that you cannot just use `--standalone` when you're also running the UI, you also need to pass the config.

This is separate from the port issue, since I did change the port locally.

I had to pass the default standalone config file, while commenting out some options. See here https://github.com/conda-incubator/conda-store-ui/issues/322#issuecomment-1785488710.

This is likely happening because the config sets this `c.CondaStoreServer.url_prefix = "/conda-store"`, but I still need to investigate.

### Expected behavior

Being able to use the UI while just running with --standalone.

server  0760fa2b386bc207936df89f52b4df83f61e8af8 
UI  d52ae0734c194bcef309d4307aae16c4a9006d66 

The link to the comment above has repro instructions.

See above.

### Output

_No response_

_No response_

### Anything else?

_No response_

nkaretnikov:
This is what conda-store-ui expects, which allows running conda-store-server in standalone mode without passing the config.

Fixes #646.

### DIFF
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -88,7 +88,7 @@ class CondaStoreServer(Application):
     )
 
     url_prefix = Unicode(
-        "/",
+        "/conda-store",
         help="the prefix URL (subdirectory) for the entire application; "
         "it MUST start with a forward slash - tip: "
         "use this to run conda-store within an existing website.",
--- a/conda-store-server/tests/conftest.py
+++ b/conda-store-server/tests/conftest.py
@@ -53,6 +53,7 @@ def conda_store_config(tmp_path, request):
 
 @pytest.fixture
 def conda_store_server(conda_store_config):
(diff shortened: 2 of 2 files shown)


nkaretnikov:
With this PR, it's not possible to just do `python -m conda_store_server.server --standalone` and use the Login button on the homepage, which is served by conda-store-ui. Tested on Windows and Linux. See the issue for details.

costrouc:
@nkaretnikov I'm surprised this is needed. For the typical install I would expect url_prefix to be `/`. The only reason `/conda-store` is used is since this is required for deployments with multiple services under the same domain. We can meet to talk about this if you'd like?

nkaretnikov:
Discussed this offline. A better solution would be to do this on the UI side. The /conda-store prefix exists mostly for running within Nebari.

PR #652, opened 2023-11-04
Use shorter `build_key`

### LINKED ISSUE #611: Shorten hash used in environment path
### Context

Environment paths currently use a hash of the environment plus a timestamp plus the environment name. For example `99108419ad0fd922fdeff9bbc434b58d41f68e3f923a83f6a7ab19568463bc82-20230915-201619-848782-1-test` for an environment named "test". 

This path is very long (90 characters plus the environment name) and causes issues on Windows where the default filesystem limits paths to 260 characters, and some packages can have long paths. See https://github.com/conda-incubator/conda-store/issues/588. For example, `notebook` has a path with 122 characters (https://github.com/conda-incubator/conda-store/issues/588#issuecomment-1747541889), so 122 + 90 is 212 characters, leaving only 48 characters for the root of wherever the conda-store state directory is stored plus the environment name. 

Long paths can also cause issues on other platforms for some packages that are built to only be able to be installed in paths up to 255 characters. See https://github.com/conda-incubator/conda-store/issues/588#issuecomment-1751556643

### Value and/or benefit

On Windows, this can be fixed by setting a registry key, but this requires administrator privileges, so some deployments might not be able to do it. 

### Anything else?

See https://github.com/conda-incubator/conda-store/issues/588 for more technical details on the Windows path length problem. 

My suggestion would be to use a shorter hash. We can incorporate the timestamp into the hash, since it is already stored in the database. However, we would need to make sure this is done in a way that is backwards compatible.

nkaretnikov:
Fixes #611.


nkaretnikov:
Envs work by constructing URLs based on build_key. These are then mapped to the state/storage dirs without checking whether the files are there. Active env switching is done via symlinking, which is atomic, also without checking. All this is fine. However, to make sure that access to old files is not lost via the web UI, it'll be necessary to mark old builds as such and store it in the DB for all builds before this commit (via a migration). All builds after this commit should use a new build version, which would allow distinguishing which build_key version to use for lookup.

conda-store-server/conda_store_server/orm.py
```diff
@@ -225,15 +225,20 @@ def build_key(self):
         The build key should be a key that allows for the environment
         build to be easily identified and found in the database.
         """
-        datetime_format = "%Y%m%d-%H%M%S-%f"
-        return f"{self.specification.sha256}-{self.scheduled_on.strftime(datetime_format)}-{self.id}-{self.specification.name}"
+        hash = self.specification.sha256[:4]
+        timestamp = int(self.scheduled_on.timestamp())
+        id = self.id
+        name = self.specification.name[:16]
+        return f"{hash}-{timestamp}-{id}-{name}"
```
nkaretnikov, line 232: This could be prefixed with `v2-` (or similar) to allow for multiple versions with 3 `-` delimiters right away. But I'm in favor of adding this in v3 or later if there are clashes. The formats are distinct now and we save 3 chars by not prefixing here, see `parse_build_key`.

nkaretnikov: Nope. Cannot rely on dashes since name might contain them. Will use the hash size to differentiate between versions.

nkaretnikov:
Seems fine. Still need to do the final review and test locally.

conda-store-server/conda_store_server/orm.py
```diff
@@ -225,15 +233,29 @@ def build_key(self):
...
-        return f"{self.specification.sha256}-{self.scheduled_on.strftime(datetime_format)}-{self.id}-{self.specification.name}"
+        if self.build_key_version == 1:
+            datetime_format = "%Y%m%d-%H%M%S-%f"
+            return f"{self.specification.sha256}-{self.scheduled_on.strftime(datetime_format)}-{self.id}-{self.specification.name}"
+        elif self.build_key_version == 2:
+            hash = self.specification.sha256[:4]
+            timestamp = int(self.scheduled_on.timestamp())
+            id = self.id
+            name = self.specification.name[:16]
+            return f"{hash}-{timestamp}-{id}-{name}"
```
jaimergp, line 244: These lengths (4 and 16) are hardcoded and there's no explanation on why that was chosen. Instead I'd rather have a fixed-length hash à la nix.

nkaretnikov: I like Nix hashes as well (specifically, I'm talking about nix-store hashes, since those also use the filesystem, see [here](https://nixos.org/guides/nix-pills/nix-store-paths)), but it can happen later and in a separate PR. It requires quite a bit of work and it was pointed out to me during the last team meeting that we might want to allocate our time elsewhere for now.

Here are some questions to answer for Nix hashes:
- To make this really fixed-size, we would need to design the v2 path scheme such that it also doesn't include the namespace name. It should be just: `store_directory / hash / <env contents>` where the first two components is the prefix that needs to be <= 255 to satisfy the conda prefix contraint.
- To hash like this and avoid weird issues due to mismatched version, we cannot hash the specification as we do now, we need to hash the contents of the lock file. Specs don't have the versions pinned, lock files are deterministic.
- Since the hashes are shared between users (no namespace anymore), they need to be resistant to collisions. The size of the hash and its collision resistance are related. This can be estimated to find the best size for this use case.
- Hashes are one-way (by design). But we also use `build_key`s to get the `build_id`. See `parse_build_key` and how it's used in `get_docker_image_manifest`. For that, we'd need to store an extra `hash -> id` mapping in the DB.
- IIUC, all other parts of the `build_key` are just to help with debugging or search. You can verify which env on disk corresponds to which specification, see when it was built, and get a human-readable name of the env.
- `build_id`s are unique because they are primary keys in the `Build` table. The only concern here is that they might potentially get reused by some DB backends.

How 4 and 16 were selected:
- Because only the `build_id`s matter, everything else doesn't help with the uniqueness part, just helps with debugging
- I selected 4 to still be able to verify that a spec matches the env by hashing it and have some confidence
- I selected 16 because it contained enough information to be meaningful with some env names I've tried, but also because it's short enough.

Summary:
- Since #653 adds error checking for this and documentation, it makes it easier for users to understand what's going on
- I'm going to bump 4 to 8 since it's a more standard size for truncation (even though it's not necessary here). This makes it somewhat less arbitrary. The original proposal was a unix timestamp and a truncated hash.
- I'm going to remove the truncation from names (since the linked PR has instructions for users to use shorter names if there are problems). It removes this arbitrary decision, which might lead to confusion later.
- Let's discuss the Nix-hash-like idea separately in a new issue.

jaimergp: > It requires quite a bit of work and it was pointed out to me during the last team meeting that we might want to allocate our time elsewhere for now.

My main concern here is that we are going to break the path interface already, so maybe there's a benefit in only doing it once and for good. At the same time, if we perfect the art of migrating paths, then it's ok :) I guess it can be discussed in the weekly tomorrow.

nkaretnikov:
Manual tests I did (as of commit 738f650):
- [x] conda-store state created on main works with changes in the PR (links to the lock file and other files work in the UI)
- [x] having build_key_version=1 in the config allows to access v2 files in the UI
- [x] having build_key_version=2 in the config allows to access v1 files in the UI
- [x] having build_key_version=3 in the config raises an error.

nkaretnikov:
This is currently blocked due to an upstream test hanging (or taking too long?) (unrelated to this PR): https://github.com/conda-incubator/conda-store/issues/666

I'll try to fix the upstream issue first before landing this.

nkaretnikov:
@jaimergp Updated, PTAL. Added a command-line/config parameter, now users can switch between v1 and v2 if needed. By default the short hash version is used. With the new default, old environments are still accessible in the UI.

conda-store-server/conda_store_server/orm.py
nkaretnikov: There's no breakage. Everything is migrated transparently and you can switch back and forth. Nix-like hashes would be more different compared to what we have right now. I still see value in having just a truncated hash and timestamp (as in the v2 here). Nix-style hashes can be added later.

conda-store-server/conda_store_server/app.py
```diff
@@ -364,6 +380,11 @@ def session_factory(self):
             url=self.database_url,
             poolclass=QueuePool,
         )
+
+        # Sets the default build_key_version value in the DB based on the config
+        import conda_store_server
+
+        conda_store_server._BUILD_KEY_VERSION = self.build_key_version
```
jaimergp, line 387: This feels a bit hacky. Why do we need a global constant for this?

conda-store-server/conda_store_server/orm.py
```diff
@@ -234,15 +257,29 @@ def build_key(self):
...
-        return f"{self.specification.sha256}-{self.scheduled_on.strftime(datetime_format)}-{self.id}-{self.specification.name}"
+        if self.build_key_version == 1:
+            datetime_format = "%Y%m%d-%H%M%S-%f"
+            return f"{self.specification.sha256}-{self.scheduled_on.strftime(datetime_format)}-{self.id}-{self.specification.name}"
+        elif self.build_key_version == 2:
+            hash = self.specification.sha256[:_BUILD_KEY_V2_HASH_SIZE]
+            timestamp = int(self.scheduled_on.timestamp())
+            id = self.id
+            name = self.specification.name
+            return f"{hash}-{timestamp}-{id}-{name}"
```
jaimergp, line 268: It feels that these should be better encoded in separate callables for clarity. Something like a dictionary of versions -> callables, that implement each format cleanly. This would also double as the catalog of available versions for validation and error reporting.

conda-store-server/conda_store_server/orm.py
```diff
@@ -234,15 +257,29 @@ def build_key(self):
...
-            return None
-        return int(parts[4])  # build_id
+        # Note: cannot rely on the number of dashes to differentiate between
+        # versions because name can contain dashes. Instead, this relies on the
+        # hash size to infer the format. The name is the last field, so indexing
+        # to find the id is okay.
+        if key[_BUILD_KEY_V2_HASH_SIZE] == "-":  # v2
+            return int(parts[2])  # build_id
+        else:  # v1
+            return int(parts[4])  # build_id
```
jaimergp, line 282: This is also a bit hacky. We shouldn't rely on lossy info for this, but a well-known id->info relationship in the database. I know you know, and that that'd be considered ideal and we don't have the time but I wonder what's the rush to ship this now instead of in two weeks (still within the STF window).

If this is not going to be addressed now, and if we don't have an issue for this limitation, we should create one.

conda-store-server/conda_store_server/orm.py
```diff
@@ -182,6 +197,14 @@ class Build(Base):
     started_on = Column(DateTime, default=None)
     ended_on = Column(DateTime, default=None)
     deleted_on = Column(DateTime, default=None)
+    build_key_version = Column(Integer, default=_get_build_key_version, nullable=False)
+
+    @validates("build_key_version")
+    def validate_build_key_version(self, key, build_key_version):
+        if build_key_version not in [1, 2]:
```
jaimergp, line 204: This can be a tuple.
```suggestion
        if build_key_version not in (1, 2):
```

But hardcoding these options here (separate from the definitions) is also strange. At this stage we should know which build_key_versions are available via an external name.

conda-store-server/conda_store_server/orm.py
```diff
@@ -182,6 +197,14 @@ class Build(Base):
...
+            raise ValueError(f"invalid build_key_version={build_key_version}")
```
jaimergp, line 205:
```suggestion
            raise ValueError(f"invalid build_key_version={build_key_version}. Accepted values are {1, 2}")
```

Proof again that this is missing something. We should be able to easily say which ones are accepted in the error. But we would be forced to hardcode those again.

conda-store-server/conda_store_server/app.py
nkaretnikov: There is a big comment about it in a different file. No other way to do it due to import cycles.

```
# Avoids a cyclic dependency between the orm module and the module defining
# CondaStore.build_key_version. Because the orm module is loaded early on
# startup, we want to delay initialization of the build_key_version field until
# it's been read from the config.
```

conda-store-server/conda_store_server/orm.py
nkaretnikov: I want to ship this version of the build key regardless because:
- it somewhat mitigates the length issues people might be having (see the test, the new version is 2x shorter)
- it's close enough to what we have already so risk of potential issues is minimized
- I consider Nix-like hashes a riskier change and would like to have this as an intermediate step, so people could fall back to this if needed.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -6,5 +7,81 @@
...
+        name = build.specification.name
+        return f"{hash}-{timestamp}-{id}-{name}"
+
+    def _version2_fmt(build: "Build"):  # noqa: F821
+        tzinfo = datetime.timezone.utc
+        hash = build.specification.sha256[: BuildKey._version2_hash_size]
+        timestamp = int(build.scheduled_on.replace(tzinfo=tzinfo).timestamp())
+        id = build.id
+        name = build.specification.name
+        return f"{hash}-{timestamp}-{id}-{name}"
```
nkaretnikov, line 35: Note: without `tzinfo`, this would vary between machines because `timestamp` uses the system clock by default, which might be not in UTC.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -6,5 +7,81 @@
...
+
+    # version -> fmt function
+    _fmt = {
+        1: _version1_fmt,
+        2: _version2_fmt,
+    }
+
+    @classmethod
+    def _check_version(cls, build_key_version):
```
nkaretnikov, line 44: Forgot type hint here, will fix.

conda-store-server/conda_store_server/app.py
```diff
@@ -107,20 +108,17 @@ class CondaStore(LoggingConfigurable):
...
-        if proposal.value not in expected:
-            raise TraitError(
-                f"c.CondaStore.build_key_version: invalid build key version: "
-                f"{proposal.value}, expected: {expected}"
-            )
-        return proposal.value
+        try:
+            return BuildKey.set_current_version(proposal.value)
+        except Exception as e:
+            raise TraitError(f"c.CondaStore.build_key_version: {e}")
```
nkaretnikov, line 121: The current version is now set right after `build_key_version` is processed, which is better than before, when it was done only when creating a DB session.

conda-store-server/conda_store_server/orm.py
```diff
@@ -197,14 +182,21 @@ class Build(Base):
     started_on = Column(DateTime, default=None)
     ended_on = Column(DateTime, default=None)
     deleted_on = Column(DateTime, default=None)
+
+    @staticmethod
+    def _get_build_key_version():
+        # Uses local import to make sure current version is initialized
+        from conda_store_server import BuildKey
+
+        return BuildKey.current_version()
```
nkaretnikov, line 191: All `conda_store_server` imports are local in this file to avoid a cyclic import problem mentioned in a `BuildKey` class comment.

conda-store-server/tests/test_actions.py
```diff
@@ -238,6 +248,18 @@ def test_add_lockfile_packages(
...
+    if build_key_version == 0:  # invalid
+        with pytest.raises(TraitError, match=(
+            r"c.CondaStore.build_key_version: invalid build key version: 0, "
+            r"expected: \(1, 2\)"
+        )):
+            conda_store.build_key_version = build_key_version
+        return  # invalid, nothing more to test
+    conda_store.build_key_version = build_key_version
+    assert BuildKey.current_version() == build_key_version
+    assert BuildKey.versions() == (1, 2)
```
nkaretnikov, line 261: This now initializes `build_key_version` via `conda_store` as one would in real life (via the config)

docs/administration.md
```diff
@@ -571,6 +571,55 @@ affect both the `build_path` length and the filename length.
...
+```
+c.CondaStore.build_key_version = 1
+```
+
+The version 2 format can also be explicitly set if needed (this is the same as
+the default):
+
+```
+c.CondaStore.build_key_version = 2
+```
```
nkaretnikov, line 621: This also needs to be mentioned where all CondaStore options are listed in this file

nkaretnikov:
@jaimergp Addressed your feedback, PTAL

conda-store-server/conda_store_server/app.py
jaimergp: Is it better to use `raise X from e` here?

docs/administration.md
```diff
@@ -571,6 +574,55 @@ affect both the `build_path` length and the filename length.
...
+To help mitigate build path length issues, a shorter build key format was
+introduced, known as version 2:
+
+```
+c7afdeff-1699156450-12345678-this-is-a-long-environment-name
+^ (1)    ^ (2)      ^ (3)    ^ (4)
+```
+
+It consists of:
+1. a truncated SHA-256 hash of the environment specification
```
jaimergp, line 603: By "environment specification" you mean the input package requests, or the solved environment?

conda-store-server/conda_store_server/__init__.py
```diff
@@ -1,6 +1,88 @@
...
+import typing
 from pathlib import Path
 
 __version__ = "2023.10.1"
 
 
 CONDA_STORE_DIR = Path.home() / ".conda-store"
+
+
+class BuildKey:
```
jaimergp, line 11: Nice, this looks way better. However I still sniff some code smell here 🤔 I don't get why we are stuck with a multiversion class in the `__init__` module. I would have assumed it was simple enough to have a setup like this:

```python
# optionally, we could define a base class with the common stuff
# omitted for clarity
class BuildKeyV1:
  ...  # v1 specificities
class BuildKeyV2:
  ...  # v2 specificities, overrides as necessary

BUILD_KEY_VERSIONS = {
  1: BuildKeyV1,
  2: BuildKeyV2,
  None: BuildKeyV2,
}
```

This can be all put in a separate module if necessary. Then the consumer code will check in the versions dict which one to use according to the current config. In other words, I think we need to take a look at the problem that causes the circular import and see if there's a better way to define these ideas.

Thanks for the patience! I feel we are getting closer.

nkaretnikov: > I think we need to take a look at the problem that causes the circular import

What we want:
- We want a user-configurable build-key
- The argument/config parsing is done in `conda-store-server/conda_store_server/app.py:CondaStore`
- After reading a config value, we want to update the `orm` module.

Problems:
- The `orm` module is loaded earlier than `app` on startup
- The `app` module also depends on the `orm` module (imports it).

Solution:
- A common module (`__init__.py` in this case) and delayed initialization (`default=_get_build_key_version` in `orm.py`) is a way to solve this
- A common module avoids a cyclic import
- Delayed initialization avoids caching the value on early imports. It's always fetched from a common module after it's been initialized once we actually want to use the `orm` module (as opposed to just importing it).

----

> BUILD_KEY_VERSIONS

I don't understand this new proposal at all.

- What's exactly wrong with the `BuildKey` class?
- What issue does the new proposal solve compared to the existing `BuildKey` class?
- Why do we want to have 2 classes for this? We could, but I just don't see the benefit.

Moreover, I see some problems with this new proposal:
- Nothing prevents me from doing `BUILD_KEY_VERSIONS[non-existent-version]` and getting a generic error. In the current `BuildKey` implementation, error-checking is shared and hidden.
- This will still need another global variable to track the current version (to index into this dict). In the current `BuildKey` implementation, this is hidden inside the class. All `BuildKey`-related things are part of the class.
- Regarding `None: BuildKeyV2`: This is wrong and will hide potential issues if build key version is not initialized (while the user might be requesting BuildKeyV1 in the config). This can happen when a module is imported earlier than the build key version is set and the value is cached. This currently doesn't happen, but someone might run into this issue when refactoring. In my code, I have an assertion to guard against that.

**Summary**:
- I'm going to follow up offline to clarify this proposal
- Its value is not clear to me and I see some problems with it
- Implementing and testing this again would take time away from other things I could be working on.

conda-store-server/conda_store_server/app.py
nkaretnikov: In general, yes. But it doesn't work with traitlets. E.g.,:

```py
            raise TraitError("c.CondaStore.build_key_version") from e
```

would print

```
[CondaStoreServer] CRITICAL | Bad config encountered during initialization: c.CondaStore.build_key_version
```

No traceback is included. No additional information is printed.

nkaretnikov:
@jaimergp PTAL. Replied to all of your comments. Only made a change to the docs.

jaimergp (approved):

PR #653, opened 2023-11-05
Check the size of `build_path`

### LINKED ISSUE #649: [ENH] - Ensure that the length of all environment paths <= 255 characters
### Feature description

Comes from https://github.com/conda-incubator/conda-store/issues/588

---

`conda` packages are guaranteed to be relocatable as long as the environment prefix length is <=255 characters.

Since we can have control over that, so I am suggesting we add checks to ensure that:

- There are checks to control the length of the base installation path. Reject those that might incur in too long paths, or inform to the user the might need to choose a different base path for the environments.
- If the hash used in the prefix name is configurable (see https://github.com/conda-incubator/conda-store/issues/611), consider that value in the above checks.

The 255 char value can also be adjusted down depending on the system, because we might need more wiggle room for Windows installations. So I'd say a `MAX_PREFIX_LENGTH=255` should be default, but it can be adjusted to a value smaller than that.

### Value and/or benefit

If this is not controlled, users might run into hard to diagnose bugs like 

```
```

, or buffer overflows, linking errors, etc.

It will also help palliating the problems with long paths in Windows, in general, where we might want to set an even lower threshold.

### Anything else?

Extra explanation about relocatability:

conda packages are relocatable thanks to a neat trick: all packages are built under a 255-char long prefix. This long prefix acts as a "placeholder" that gets replaced at install-time so it matches the destination prefix.

For example: I have a binary that hardcodes (for whatever reason) a path to another executable it needs to do a certain operation. At build time, the long prefix will have something like `/opt/conda-build-workspace/project-1234abcd/envs/_h_env_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeholder_placeho`. At install time, after requesting `conda create -p /opt/my_envs/hallo`, it will be replaced with that path, and the remaining characters will be NULL, and everyone's happy.

This trick is guaranteed to work 100% AS LONG AS the target prefix is under 255 characters. In some cases, longer paths might work because the environment doesn't contain binaries (text replacement is not as constrained) that need to be patched, but it's not guaranteed.

nkaretnikov:
Fixes #649.

### DIFF
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -107,27 +107,27 @@ def build_conda_environment(db: Session, conda_store, build):
     symlink the build to a named environment
 
     """
-    set_build_started(db, build)
-    append_to_logs(
-        db,
-        conda_store,
-        build,
-        f"starting build of conda environment {datetime.datetime.utcnow()} UTC\n",
-    )
-
-    settings = conda_store.get_settings(
-        db=db,
-        namespace=build.environment.namespace.name,
-        environment_name=build.environment.name,
-    )
-
-    conda_prefix = build.build_path(conda_store)
-    conda_prefix.parent.mkdir(parents=True, exist_ok=True)
-
-    environment_prefix = build.environment_path(conda_store)
-    environment_prefix.parent.mkdir(parents=True, exist_ok=True)
-
     try:
+        set_build_started(db, build)
+        append_to_logs(
+            db,
+            conda_store,
+            build,
+            f"starting build of conda environment {datetime.datetime.utcnow()} UTC\n",
+        )
+
+        settings = conda_store.get_settings(
+            db=db,
+            namespace=build.environment.namespace.name,
+            environment_name=build.environment.name,
+        )
+
+        conda_prefix = build.build_path(conda_store)
+        conda_prefix.parent.mkdir(parents=True, exist_ok=True)
+
+        environment_prefix = build.environment_path(conda_store)
+        environment_prefix.parent.mkdir(parents=True, exist_ok=True)
+
         with utils.timer(conda_store.log, f"building conda_prefix={conda_prefix}"):
             context = action.action_solve_lockfile(
                 settings.conda_command,
@@ -202,14 +202,14 @@ def build_conda_environment(db: Session, conda_store, build):
 
         set_build_completed(db, conda_store, build)
     except subprocess.CalledProcessError as e:
+        set_build_failed(db, build)
         conda_store.log.exception(e)
         append_to_logs(db, conda_store, build, e.output)
-        set_build_failed(db, build)
         raise e
     except Exception as e:
+        set_build_failed(db, build)
         conda_store.log.exception(e)
         append_to_logs(db, conda_store, build, traceback.format_exc())
-        set_build_failed(db, build)
         raise e
 
 
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -10,6 +10,7 @@ import typing
 
 import yaml
 from conda_store_server import action, api, conda_utils, orm, schema, utils
+from conda_store_server.utils import BuildPathError
 from sqlalchemy.orm import Session
 
 
@@ -38,8 +39,11 @@ def set_build_started(db: Session, build: orm.Build):
     db.commit()
 
 
-def set_build_failed(db: Session, build: orm.Build):
+def set_build_failed(
+    db: Session, build: orm.Build, status_info: typing.Optional[str] = None
+):
     build.status = schema.BuildStatus.FAILED
+    build.status_info = status_info
     build.ended_on = datetime.datetime.utcnow()
     db.commit()
 
@@ -109,6 +113,9 @@ def build_conda_environment(db: Session, conda_store, build):
     """
     try:
         set_build_started(db, build)
+        # Note: even append_to_logs can fail due to filename size limit, so
+        # check build_path length first
+        conda_prefix = build.build_path(conda_store)
         append_to_logs(
             db,
             conda_store,
@@ -122,7 +129,6 @@ def build_conda_environment(db: Session, conda_store, build):
             environment_name=build.environment.name,
         )
 
-        conda_prefix = build.build_path(conda_store)
         conda_prefix.parent.mkdir(parents=True, exist_ok=True)
 
         environment_prefix = build.environment_path(conda_store)
@@ -201,10 +207,19 @@ def build_conda_environment(db: Session, conda_store, build):
         build.size = context.result["disk_usage"]
 
         set_build_completed(db, conda_store, build)
+    # Always mark build as failed first since other functions may throw an
+    # exception
     except subprocess.CalledProcessError as e:
         set_build_failed(db, build)
         conda_store.log.exception(e)
         append_to_logs(db, conda_store, build, e.output)
+        raise e
+    except BuildPathError as e:
+        # Provide status_info, which will be exposed to the user, ONLY in this
+        # case because the error message doesn't expose sensitive information
+        set_build_failed(db, build, status_info=e.message)
+        conda_store.log.exception(e)
+        append_to_logs(db, conda_store, build, traceback.format_exc())
         raise e
     except Exception as e:
         set_build_failed(db, build)
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -7,6 +7,7 @@ import shutil
 
 from conda_store_server import conda_utils, schema, utils
 from conda_store_server.environment import validate_environment
+from conda_store_server.utils import BuildPathError
 from sqlalchemy import (
     JSON,
     BigInteger,
@@ -173,6 +174,9 @@ class Build(Base):
     package_builds = relationship("CondaPackageBuild", secondary=build_conda_package)
 
     status = Column(Enum(schema.BuildStatus), default=schema.BuildStatus.QUEUED)
+    # Additional status info that will be provided to the user. DO NOT put
+    # sensitive data here
+    status_info = Column(UnicodeText, default=None)
     size = Column(BigInteger, default=0)
     scheduled_on = Column(DateTime, default=datetime.datetime.utcnow)
     started_on = Column(DateTime, default=None)
@@ -202,9 +206,7 @@ class Build(Base):
         # conda prefix must be less or equal to 255 chars
         # https://github.com/conda-incubator/conda-store/issues/649
         if len(str(res)) > 255:
-            raise ValueError(
-                f"build_path too long: {res} must be <= 255 chars, got {len(str(res))}"
-            )
+            raise BuildPathError("build_path too long: must be <= 255 chars")
         return res
 
     def environment_path(self, conda_store):
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -206,7 +206,7 @@ class Build(Base):
         # conda prefix must be less or equal to 255 chars
         # https://github.com/conda-incubator/conda-store/issues/649
         if len(str(res)) > 255:
-            raise BuildPathError("build_path too long: must be <= 255 chars")
+            raise BuildPathError("build_path too long: must be <= 255 characters")
         return res
 
     def environment_path(self, conda_store):
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -190,15 +190,22 @@ class Build(Base):
         """
         store_directory = os.path.abspath(conda_store.store_directory)
         namespace = self.environment.namespace.name
-        name = self.specification.name
-        return (
+        res = (
             pathlib.Path(
                 conda_store.build_directory.format(
-                    store_directory=store_directory, namespace=namespace, name=name
+                    store_directory=store_directory,
+                    namespace=namespace,
                 )
             )
             / self.build_key
         )
+        # conda prefix must be less or equal to 255 chars
+        # https://github.com/conda-incubator/conda-store/issues/649
+        if len(str(res)) > 255:
+            raise ValueError(
+                f"build_path too long: {res} must be <= 255 chars, got {len(str(res))}"
+            )
+        return res
 
     def environment_path(self, conda_store):
         """Environment path is the path for the symlink to the build
--- a/conda-store-server/conda_store_server/utils.py
+++ b/conda-store-server/conda_store_server/utils.py
@@ -15,6 +15,10 @@ class CondaStoreError(Exception):
         return self.args[0]
 
 
+class BuildPathError(CondaStoreError):
+    pass
+
+
 def symlink(source, target):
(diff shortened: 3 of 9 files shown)


conda-store-server/conda_store_server/orm.py
```diff
@@ -190,15 +190,22 @@ def build_path(self, conda_store):
         """
         store_directory = os.path.abspath(conda_store.store_directory)
         namespace = self.environment.namespace.name
-        name = self.specification.name
```
nkaretnikov, line 193: Not used in `build_directory`.

nkaretnikov:
TODO:
- [ ] I think this needs a unit test, to make sure the error is raised.
- [ ] A simple integration test would be trying to create an environment with a very long environment name. This affects the build_key and hence the build_path/prefix, so should error out.

conda-store-server/tests/test_server.py
```diff
@@ -335,6 +336,58 @@ def test_create_specification_unauth(testclient):
...
+    r = schema.APIPostSpecification.parse_obj(response.json())
+    assert r.status == schema.APIStatus.OK
+
+    # Try checking that the status is 'FAILED'
+    is_updated = False
+    for _ in range(5):
+        time.sleep(5)
+
+        # check for the given build
+        response = testclient.get(f"api/v1/build/{r.data.build_id}")
```
nkaretnikov, line 374: This is a bug: it should copy `build_id` from the `r` above since `r` is reused in a loop. See the other test that does this right.

nkaretnikov:
This PR should be merged first since it has new tests, which are useful. I think it will conflict with #652 because it uses a migration (migration hashes won't match). #531 might require rebasing as well.

nkaretnikov:
Companion PR on the UI side (see screenshot there): https://github.com/conda-incubator/conda-store-ui/pull/332

conda-store-server/tests/test_server.py
```diff
@@ -335,6 +336,60 @@ def test_create_specification_unauth(testclient):
...
+# cause errors as well, but would be flaky since the test conda-store state
+# directory might have different lengths on different systems, for instance,
+# due to different username lengths.
+@pytest.mark.parametrize(
+    "size",
+    [
+        # OSError: [Errno 36] File name too long
+        255,
+        # OSError: [Errno 36] File name too long
+        256,
```
nkaretnikov, line 349: These error messages the code used to throw before I moved the `build_path` to the start of the function in `build_conda_environment`

tests/test_api.py
```diff
@@ -458,6 +459,67 @@ def test_create_specification_auth(testclient):
...
+    response = testclient.post(
+        "api/v1/specification",
+        json={
+            "namespace": namespace,
+            "specification": json.dumps({"name": environment_name}),
+        },
+    )
+    if size > 255:
+        assert response.status_code == 500
+        return  # error, nothing to do
```
nkaretnikov, line 492: Note the change here compared to `test_server.py`. Again, this is due to different backends being used for these.

conda-store-server/conda_store_server/utils.py
```diff
@@ -15,6 +15,10 @@ def message(self):
         return self.args[0]
 
 
+class BuildPathError(CondaStoreError):
```
jaimergp, line 18: I'm more used to having exceptions in an `exceptions.py` module, but I guess there's a precedence to have them in `utils.py`? Probably worth its own issue, not blocking here.

conda-store-server/tests/test_server.py
```diff
@@ -335,6 +336,60 @@ def test_create_specification_unauth(testclient):
...
+# Only testing size values that will always cause errors. Smaller values could
+# cause errors as well, but would be flaky since the test conda-store state
+# directory might have different lengths on different systems, for instance,
+# due to different username lengths.
+@pytest.mark.parametrize(
+    "size",
+    [
+        # OSError: [Errno 36] File name too long
+        255,
+        # OSError: [Errno 36] File name too long
```
jaimergp, line 348: This kind of message is Unix specific, right? How would they look on Windows?

tests/test_api.py
```diff
@@ -458,6 +459,67 @@ def test_create_specification_auth(testclient):
...
+# cause errors as well, but would be flaky since the test conda-store state
+# directory might have different lengths on different systems, for instance,
+# due to different username lengths.
+@pytest.mark.parametrize(
+    "size",
+    [
+        # minio.error.MinioException: S3 operation failed; code:
+        # XMinioInvalidObjectName, message: Object name contains unsupported
+        # characters.
+        # The error message is misleading: it's a size issue.
```
jaimergp, line 472: Yea this is concerning. Any idea why we get this error only with 255?

conda-store-server/conda_store_server/build.py
```diff
@@ -201,15 +207,24 @@ def build_conda_environment(db: Session, conda_store, build):
         build.size = context.result["disk_usage"]
 
         set_build_completed(db, conda_store, build)
+    # Always mark build as failed first since other functions may throw an
+    # exception
     except subprocess.CalledProcessError as e:
+        set_build_failed(db, build)
```
jaimergp, line 213: Don't we want to add more info here too via `status_info`?

conda-store-server/conda_store_server/build.py
```diff
@@ -38,8 +39,11 @@ def set_build_started(db: Session, build: orm.Build):
     db.commit()
 
 
-def set_build_failed(db: Session, build: orm.Build):
+def set_build_failed(
+    db: Session, build: orm.Build, status_info: typing.Optional[str] = None
+):
     build.status = schema.BuildStatus.FAILED
+    build.status_info = status_info
```
jaimergp, line 46: What's the maximum size allowed? Do we need to check?

conda-store-server/conda_store_server/build.py
```diff
@@ -201,15 +207,24 @@ def build_conda_environment(db: Session, conda_store, build):
...
-    except Exception as e:
+    except BuildPathError as e:
+        # Provide status_info, which will be exposed to the user, ONLY in this
+        # case because the error message doesn't expose sensitive information
+        set_build_failed(db, build, status_info=e.message)
         conda_store.log.exception(e)
         append_to_logs(db, conda_store, build, traceback.format_exc())
+        raise e
+    except Exception as e:
         set_build_failed(db, build)
```
jaimergp, line 225: Same, any `status_info` needed here?

conda-store-server/conda_store_server/orm.py
```diff
@@ -173,6 +174,9 @@ class Build(Base):
     package_builds = relationship("CondaPackageBuild", secondary=build_conda_package)
 
     status = Column(Enum(schema.BuildStatus), default=schema.BuildStatus.QUEUED)
+    # Additional status info that will be provided to the user. DO NOT put
+    # sensitive data here
+    status_info = Column(UnicodeText, default=None)
```
jaimergp, line 179: Any maximum length recommended?

conda-store-server/conda_store_server/orm.py
```diff
@@ -190,15 +194,20 @@ def build_path(self, conda_store):
...
+                    store_directory=store_directory,
+                    namespace=namespace,
                 )
             )
             / self.build_key
         )
+        # conda prefix must be less or equal to 255 chars
+        # https://github.com/conda-incubator/conda-store/issues/649
+        if len(str(res)) > 255:
+            raise BuildPathError("build_path too long: must be <= 255 characters")
```
jaimergp, line 209: Maybe we can afford to be more informative here and report which components are responsible for `build_path`. What about:

```python
f"'build_path' is too long ({len(str(build_path))}). Maximum allowed length is 255. See docs at xxx for more information.
```

This is also an opportunity to start codifying error numbers and have a reference online so folks can troubleshoot, à la https://flake8.pycqa.org/en/latest/user/error-codes.html. (worth its own issue).

conda-store-server/conda_store_server/orm.py
jaimergp: How come? Not anymore? I thought it had the hash, timestamp and name? Or is that part of the 'build_key'  thingy?

jaimergp (commented):
Looks good!

The main part of this PR is:

```python
if len(str(res)) > 255:
    raise BuildPathError("build_path too long: must be <= 255 characters")
```

I think we can afford to be more detailed here, but I don't know if that's the correct place. Maybe just codifying the error type with an identifier+number we can add some documentation on how to work around this issue?

nkaretnikov:
> I think we can afford to be more detailed here, but I don't know if that's the correct place. Maybe just codifying the error type with an identifier+number we can add some documentation on how to work around this issue?

I initially printed the whole path and its size, too. But that'd be flaky on CI because different directories are used everywhere. It also seems like an implementation detail, which is not helpful unless you have admin rights. I'll just add documentation for now and we can improve later.

conda-store-server/conda_store_server/orm.py
nkaretnikov: That what I had originally, but I removed it. See here https://github.com/conda-incubator/conda-store/pull/653#issuecomment-1807184045. As for codifying the errors: not adding that here either. There's no policy/format on this right now. Similarly, saying "see the docs" is not helpful. This also needs to be linked and done in a systematic way, otherwise links will point to nowhere over time. Not adding this also because there's no infra for this right now. Let's keep the PR self-contained, we can always change this later.

conda-store-server/conda_store_server/orm.py
nkaretnikov: SQLAlchemy doesn't makes any promises about this type: https://docs.sqlalchemy.org/en/20/core/type_basics.html#sqlalchemy.types.UnicodeText. I think it depends on your database backend, so I'm not making any promises here. I also feel like it's not something worth pointing out next to every field. This is not the only place where we use UnicodeText, so it's not something special either (see `Environment.description`).

tests/test_api.py
nkaretnikov: It's a common limit in filesystems, for instance: https://serverfault.com/questions/9546/filename-length-limits-on-linux

conda-store-server/conda_store_server/build.py
nkaretnikov: We don't need to check here as it'd clutter the code. SQLAlchemy will throw an error on inserts if this is a problem for your backend. And the intent here is that error messages that appear in `status_info` are human-written and are user-facing. No one should be passing random exceptions here because that might leak sensitive data depending on the context. You know how popular websites show you "Something went wrong" without explaining why. Or sometimes they might do "Error 12345: something failed". No real info is shared or it's encoded via a number for the dev team to look up later (but we don't have a policy for that at the moment, so not adding the number encoding thing). Same idea here.

conda-store-server/conda_store_server/build.py
nkaretnikov: No, see the comment right below this:

>     except BuildPathError as e:
>        # Provide status_info, which will be exposed to the user, ONLY in this
>        # case because the error message doesn't expose sensitive information

It's only done in specific cases where we control the error message (right now, it's only `BuildPathError` added in this PR). We don't pass random data here. See https://github.com/conda-incubator/conda-store/pull/653/files/34ac2dd4380dda1da3264e0482bafcd0e79976bf#r1390449347

conda-store-server/conda_store_server/build.py
nkaretnikov: No, see https://github.com/conda-incubator/conda-store/pull/653/files/34ac2dd4380dda1da3264e0482bafcd0e79976bf#r1390449615

conda-store-server/conda_store_server/orm.py
nkaretnikov: It was not used before this PR, just a mistake that I fixed since I was making changes here anyway.

This is the one used below, no `name` here:

```py
    build_directory = Unicode(
        "{store_directory}/{namespace}",
        help="Template used to form the directory for storing conda environment builds. Available keys: store_directory, namespace, name. The default will put all built environments in the same namespace within the same directory.",
        config=True,
    )

```

This is the one you're talking about (I don't touch this one):
```py
    environment_directory = Unicode(
        "{store_directory}/{namespace}/envs/{name}",
        help="Template used to form the directory for symlinking conda environment builds. Available keys: store_directory, namespace, name. The default will put all environments in the same namespace within the same directory.",
        config=True,
    )
```

conda-store-server/conda_store_server/utils.py
nkaretnikov: Yeah, agree. But not changing it now. Added it next to the existing exception.

conda-store-server/tests/test_server.py
nkaretnikov: This is with LongPathsEnabled = 1, otherwise Windows has the MAX_PATH limit anyway.

```py
>>> open(f"c:\{'a' * 252}.txt", 'w').write('test\n')
...
OSError: [Errno 22] Invalid argument: 'c:\\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.txt'
```

I wouldn't rely on these messages being the same between Python versions either. This is just to have some intuition why things fail. Since this is a test, I don't think the message matters much.

nkaretnikov:
Hey @jaimergp, thanks for your comments! I'm going to document the `build_path` in `docs/administration.md`, similar to how we document the `MAX_PATH` issue on Windows. Otherwise, I don't think there's a need to change anything else.

conda-store-server/conda_store_server/orm.py
jaimergp: > Let's keep the PR self-contained, we can always change this later.

Works for me, but let's make sure there's an issue we can iterate on.

conda-store-server/conda_store_server/utils.py
jaimergp: Let's add an issue to cover this later.

jaimergp (approved):
The PR covers enough to have some protection against too long paths. The information conveyed could be more detailed, and less hardcoded. But that can be worked on in future PRs.

For example, I could imagine the UI receiving a JSON payload with the details of the error instead of a simple string. And then we  decide what to do with the payload depending on the user role

At the same time, we've identified an opportunity to standardize error messages with error codes for easier (and more accurate) documentation.

These two improvements could help in future efforts like e.g. localization or theming. 

Issues need to be created to cover these suggestions, but fwiw, this PR is good to go on my side.

trallard:
@jaimergp  is this ready for merge? if so then perhaps you and/or @nkaretnikov should open a follow-up issue with potential improvements or areas to explore and merge this straight away

Issue #666, opened 2023-11-18
[BUG] - unit-test conda-store-server (ubuntu) hangs on CI


nkaretnikov:
### Describe the bug

Open a no-op PR against current main a5680f855fbf8317b8a9dbac43017ccc4d07927b, like in https://github.com/conda-incubator/conda-store/pull/665.
unit-test conda-store-server (ubuntu) hangs.

### Expected behavior

Test passes, no hang.

### How to Reproduce the problem?

.

nkaretnikov:
Currently looking into this.

- Test that hangs is `test_generate_conda_docker`
- It uses `action_generate_conda_docker`, which relies on conda-docker
- In conda-docker, the problematic function is `chroot_install`
- I think something is different in the environment now, which changed its behavior
- I modified the code like this to see the output:
```py
    subprocess.check_output(
        [
            "fakechroot",
            "chroot",
            new_root,
            "/_conda.exe",
            "install",
            "--offline",
            "--file",
            "/opt/conda/pkgs/env.txt",
            "-y",
            "--prefix",
            "/opt/conda",
        ],
        env=env,
        cwd=host_conda_opt,
        # stdout=subprocess.DEVNULL,
        # stderr=subprocess.DEVNULL,
    )
```
- On my machine where the test passes, it prints:
```
conda_package_streaming/package_streaming.py:25: UserWarning: zstandard could not be imported. Running without .conda support.
conda_package_handling/api.py:29: UserWarning: Install zstandard Python bindings for .conda support
Error while loading conda entry point: conda-libmamba-solver (libmamba.so.2: cannot open shared object file: No such file or directory)
```

- On CI, it prints a bunch of warnings instead:
```
[6472] WARNING: file already exists but should not: /_brotli.cpython-39-x86_64-linux-gnu.so
[6472] WARNING: file already exists but should not: /_ruamel_yaml.cpython-39-x86_64-linux-gnu.so
[6472] WARNING: file already exists but should not: /lib-dynload/_asyncio.cpython-39-x86_64-linux-gnu.so
[6472] WARNING: file already exists but should not: /lib-dynload/_bisect.cpython-39-x86_64-linux-gnu.so
[6472] WARNING: file already exists but should not: /lib-dynload/_blake2.cpython-39-x86_64-linux-gnu.so
[6472] WARNING: file already exists but should not: /lib-dynload/_bz2.cpython-39-x86_64-linux-gnu.so
[...]

PR #667, opened 2023-11-21
Disable code that depends on conda-docker

nkaretnikov:
See #666.

### DIFF
--- a/conda-store-server/conda_store_server/action/generate_conda_docker.py
+++ b/conda-store-server/conda_store_server/action/generate_conda_docker.py
@@ -1,5 +1,4 @@
 import pathlib
-import sys
 
 from conda_store_server import action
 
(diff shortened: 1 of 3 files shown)


costrouc:
The conda-store docker bits are not well maintained at this moment. I would be fine with merging this but would like us to have a plan for if in the future we would like to support this. This feature is powerful since it allows conda-store to build docker images of environments without docker/root.

nkaretnikov:
@costrouc I'm planning to fix the upstream issue in conda-docker ASAP. There are some suggestions from Jaime to try, but I cannot say anything concrete since I haven't tried anything yet.

@jaimergp Updated. I've used the same message everywhere that we have elsewhere instead of the "Temporary" one. I also don't like the word "temporary" since it's somewhat meaningless (everything is temporary to some extent). I think the other message conveys the intention to support this feature in the future as well. Using the same message in the warning and error will also make it easier to discover both once people search for one of them. I also use a full URL instead of an issue number to make it easier to understand which repo this refers to.

jaimergp (approved):

PR #693, opened 2023-12-11
Set channel priority to strict, print config info

### LINKED ISSUE #359: Channel priority should be set to strict and conda info should be printed with conda-store

### LINKED ISSUE #659: [ENH] - Run conda info and mamba info before solve
### Feature description

We need more info within the build process to help debug issues. We need to excessively print information when building environments.

### Value and/or benefit

Helps debugging.

### Anything else?

_No response_

nkaretnikov:
This PR sets the channel priority to strict to avoid issues during solves when  multiple channels that were not specified are being used. This is done by passing the command line option `--strict-channel-priority` via the environment variable `CONDA_FLAGS`, which is used by conda-lock, which is a library conda-store uses to solve dependencies.

`CONDA_FLAGS` is used because it's a method of passing options that's local to conda-lock as opposed to editing global conda configuration files. Editing user configuration files may lead to unexpected issues in contexts where conda-store is used in standalone mode. Those configuration files affect how conda works outside of conda-store, which might be important to end users.

From the documentation at https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html:
> With strict channel priority, packages in lower priority channels are not considered if a package with the same name
appears in a higher priority channel.

Additionally, this PR adds code that prints useful system information related to conda configuration before the package solving process. This is useful for debugging. The commands used are:

- `conda/mamba info`, depending on the value of `conda_store.conda_command`
- `conda config --show`
- `conda config --show-sources`.

Note that the latter two commands are not supported by mamba, which is why conda is used.

Fixes #359, fixes #659.

### DIFF
--- a/conda-store-server/conda_store_server/action/generate_lockfile.py
+++ b/conda-store-server/conda_store_server/action/generate_lockfile.py
@@ -26,8 +26,10 @@ def action_solve_lockfile(
         json.dump(specification.dict(), f)
 
     def print_cmd(cmd):
-        print(f"Running command: {' '.join(cmd)}")
-        print(subprocess.check_output(cmd, stderr=subprocess.STDOUT, encoding="utf-8"))
+        context.log.info(f"Running command: {' '.join(cmd)}")
+        context.log.info(
+            subprocess.check_output(cmd, stderr=subprocess.STDOUT, encoding="utf-8")
+        )
 
     # The info command can be used with either mamba or conda
     print_cmd([conda_command, "info"])
--- a/conda-store-server/conda_store_server/action/generate_lockfile.py
+++ b/conda-store-server/conda_store_server/action/generate_lockfile.py
@@ -1,5 +1,7 @@
 import json
+import os
 import pathlib
+import subprocess
 import typing
 
 import yaml
@@ -13,6 +15,8 @@ def action_solve_lockfile(
     conda_command: str,
     specification: schema.CondaSpecification,
     platforms: typing.List[str] = [conda_utils.conda_platform()],
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html
+    conda_flags: str = "--strict-channel-priority",
 ):
     environment_filename = pathlib.Path.cwd() / "environment.yaml"
     lockfile_filename = pathlib.Path.cwd() / "conda-lock.yaml"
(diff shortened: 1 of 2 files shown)


nkaretnikov:
"Place a bound on docker-py" - this commit is a workaround for a CI issue, might need to be removed before merge.

conda-store-server/tests/test_actions.py
```diff
@@ -69,6 +69,29 @@ def test_solve_lockfile(conda_store, specification, request):
...
+def test_solve_lockfile_invalid_conda_flags(conda_store, simple_specification):
+    with pytest.raises(Exception, match=(
+        r"Command.*--this-is-invalid.*returned non-zero exit status"
+    )):
+        action.action_solve_lockfile(
+            conda_command=conda_store.conda_command,
+            specification=simple_specification,
+            platforms=[conda_utils.conda_platform()],
+            conda_flags="--this-is-invalid",
+        )
```
nkaretnikov, line 92: This test makes sure that the conda-lock library actually uses this parameter. If the internals change, this will fail and we'll update accordingly.

conda-store-server/tests/test_actions.py
```diff
@@ -69,6 +69,29 @@ def test_solve_lockfile(conda_store, specification, request):
...
 
 
+def test_solve_lockfile_valid_conda_flags(conda_store, simple_specification):
+    context = action.action_solve_lockfile(
+        conda_command=conda_store.conda_command,
+        specification=simple_specification,
+        platforms=[conda_utils.conda_platform()],
+        conda_flags="--strict-channel-priority",
+    )
+    assert len(context.result["package"]) != 0
```
nkaretnikov, line 79: This is the same as the default parameter.

conda-store-server/conda_store_server/action/generate_lockfile.py
```diff
@@ -13,19 +15,39 @@ def action_solve_lockfile(
     conda_command: str,
     specification: schema.CondaSpecification,
     platforms: typing.List[str] = [conda_utils.conda_platform()],
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html
+    conda_flags: str = "--strict-channel-priority",
```
nkaretnikov, line 19: From the docs linked above:

> We recommend setting channel priority to "strict" when possible.

> channel_priority (ChannelPriority)
Accepts values of 'strict', 'flexible', and 'disabled'. The default value is 'flexible'.

nkaretnikov:
This is ready for review. Tested manually as well. See https://github.com/conda-incubator/conda-store/pull/693#issuecomment-1849389728 before merge.

trallard (changes requested):
Thanks @nkaretnikov - I left a few comments and questions.
I generally do not like to use print for what should be logged or streamed, so unless there is a strong reason to use `print` this should be changed

conda-store-server/conda_store_server/action/generate_lockfile.py
```diff
@@ -13,19 +15,39 @@ def action_solve_lockfile(
     conda_command: str,
     specification: schema.CondaSpecification,
     platforms: typing.List[str] = [conda_utils.conda_platform()],
```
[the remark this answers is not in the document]
nkaretnikov: I propose we do this later in one go, to avoid unrelated issues with linters etc. I've opened an issue: https://github.com/conda-incubator/conda-store/issues/711

trallard (approved):
Thank you, this LGTM 🚀  merging now!

PR #705, opened 2023-12-13
Check if worker is initialized

### LINKED ISSUE #605: [BUG] - without --standalone, server runs but fails to build env
### Describe the bug

When starting the server without the `--standalone` parameter, this error appears when trying to build an env.
No user-friendly error is raised on server start up, which is confusing.

```
'NoneType' object has no attribute 'disk_free'
```

Log for Windows below, but it repros the same on Linux.

### Expected behavior

A user-friendly error is raised on server start-up

Screen Shot 2023-10-01 at 19 10 29

Env:
```
channels:
  - conda-forge
dependencies:
  - python
  - pip:
    - nothing
  - ipykernel
  # tests
  - pytest

  - requests  # for app

# other fields
description: ''
name: test-env    # required
prefix: null
variables: null
```

### Output

_No response_

On Linux, this was tested on  a8d352443795a6660e0344c93fd43de24c171ebe, which is on top of 432eaa11ef349fdb848215113caea22b09a23a32

### Anything else?

_No response_

nkaretnikov:
Fixes #605.

## Description

conda-store can run via docker and in standalone mode:
- in standalone mode, the `--standalone` CLI parameter must be used to start the worker sub-process
- when running via docker, this CLI parameter is not used because a dedicated worker process is started by docker-compose.

However, there might be cases where users forget to pass the `--standalone` parameter when running in standalone mode. This is a problem because the server cannot execute any tasks without the worker, so it must be running.

This PR adds a check that verifies whether the worker is running and prints an error message if not.

The worker is configured via a number of parameters, like `celery_broker_url`, which can either point to a database connection or use Redis:

```py
    @default("celery_broker_url")
    def _default_celery_broker_url(self):
        if self.redis_url is not None:
            return self.redis_url
        return f"sqla+{self.database_url}"
```

Similarly, checks like `conda_store.celery_app.control.inspect().ping()` won't work outside of Redis and RabbitMQ.

So we use a database entry to check whether the worker is initialized. The database is used because it will work regardless of parameters used to configure the worker.

Example output:
```
[CondaStoreServer] WARNING | Waiting for worker... Use --standalone if running outside of docker
[CondaStoreServer] WARNING | Waiting for worker... Use --standalone if running outside of docker
[CondaStoreServer] WARNING | Waiting for worker... Use --standalone if running outside of docker
[CondaStoreServer] Worker initialized
```

The warning message is printed in red, the info message is printed in green.

The simplest way to test this:
- run the server in one terminal, wait for warnings to be printed
- run the worker in another terminal, wait for the info message to be printed.

The commands to run from `conda-store/conda-store-server`:
```
terminal1: python -m conda_store_server.server
terminal2: python -m conda_store_server.worker
```

To test in standalone mode (also starts the worker):
```
python -m conda_store_server.server --standalone
```

To test via docker:
```
docker-compose build --build-arg RELEASE_VERSION=2023.10.1 && docker-compose up
```

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -11,6 +11,7 @@ from conda_store_server.utils import BuildPathError
 from sqlalchemy import (
     JSON,
     BigInteger,
+    Boolean,
     Column,
     DateTime,
     Enum,
@@ -39,6 +40,22 @@ logger = logging.getLogger("orm")
 Base = declarative_base()
 
 ARN_ALLOWED_REGEX = re.compile(schema.ARN_ALLOWED)
+
+
+class Worker(Base):
+    """Used to communicate with the worker process"""
+
+    __tablename__ = "worker"
+
+    id = Column(Integer, primary_key=True)
+
+    # Used to check whether the worker is initialized
+    initialized = Column(Boolean, default=False)
+
+    __table_args__ = (
+        # Ensures no duplicates can be added with this combination of fields.
+        UniqueConstraint("initialized", name="_uc"),
+    )
 
 
 class Namespace(Base):
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -43,13 +43,13 @@ ARN_ALLOWED_REGEX = re.compile(schema.ARN_ALLOWED)
 
 
 class Worker(Base):
-    """Used to communicate with the worker process"""
+    """For communicating with the worker process"""
 
     __tablename__ = "worker"
 
     id = Column(Integer, primary_key=True)
 
-    # Used to check whether the worker is initialized
+    # For checking whether the worker is initialized
     initialized = Column(Boolean, default=False)
 
     __table_args__ = (
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -52,11 +52,6 @@ class Worker(Base):
     # For checking whether the worker is initialized
     initialized = Column(Boolean, default=False)
 
-    __table_args__ = (
-        # Ensures no duplicates can be added with this combination of fields.
-        UniqueConstraint("initialized", name="_uc_worker"),
-    )
-
 
 class Namespace(Base):
     """Namespace for resources"""
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -54,7 +54,7 @@ class Worker(Base):
 
     __table_args__ = (
         # Ensures no duplicates can be added with this combination of fields.
-        UniqueConstraint("initialized", name="_uc"),
+        UniqueConstraint("initialized", name="_uc_worker"),
     )
 
 
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -3,6 +3,7 @@ import os
 import posixpath
 import sys
 import time
+from enum import Enum
 from threading import Thread
 
 import conda_store_server
@@ -30,6 +31,12 @@ from traitlets import (
     validate,
 )
 from traitlets.config import Application, catch_config_error
+
+
+class _Color(str, Enum):
+    GREEN = "\x1b[32m"
+    RED = "\x1b[31m"
+    RESET = "\x1b[0m"
 
 
 class CondaStoreServer(Application):
@@ -346,11 +353,6 @@ class CondaStoreServer(Application):
         return app
 
     def _check_worker(self, delay=5):
-        # Uses colors to make the output more visible
-        green = "\x1b[32m"
-        red = "\x1b[31m"
-        reset = "\x1b[0m"
-
         # Creates a new DB connection since this will be run in a separate
         # thread and connections cannot be shared between threads
         session_factory = orm.new_session_factory(
@@ -364,15 +366,17 @@ class CondaStoreServer(Application):
             with session_factory() as db:
                 q = db.query(orm.Worker).first()
                 if q is not None and q.initialized:
-                    self.log.info(f"{green}" "Worker initialized" f"{reset}")
+                    self.log.info(
+                        f"{_Color.GREEN}" "Worker initialized" f"{_Color.RESET}"
+                    )
                     break
 
             time.sleep(delay)
             self.log.warning(
-                f"{red}"
+                f"{_Color.RED}"
                 "Waiting for worker... "
                 "Use --standalone if running outside of docker"
-                f"{reset}"
+                f"{_Color.RESET}"
             )
 
     def start(self):
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -28,9 +28,6 @@ from traitlets import (
 )
 from traitlets.config import Application, catch_config_error
 
-logger = logging.getLogger("app")
-logger.setLevel(logging.INFO)
-
 
 class CondaStoreServer(Application):
     aliases = {
@@ -394,11 +391,11 @@ class CondaStoreServer(Application):
                 with session_factory() as db:
                     q = db.query(orm.Worker).first()
                     if q is not None and q.initialized:
-                        logger.info(f"{green}" "Worker initialized" f"{reset}")
+                        self.log.info(f"{green}" "Worker initialized" f"{reset}")
                         break
 
                 time.sleep(5)
-                logger.critical(
+                self.log.warning(
                     f"{red}"
                     "Waiting for worker... "
                     "Use --standalone if running outside of docker"
--- a/conda-store-server/conda_store_server/server/app.py
+++ b/conda-store-server/conda_store_server/server/app.py
@@ -28,6 +28,9 @@ from traitlets import (
 )
 from traitlets.config import Application, catch_config_error
 
+logger = logging.getLogger("app")
+logger.setLevel(logging.INFO)
+
 
 class CondaStoreServer(Application):
     aliases = {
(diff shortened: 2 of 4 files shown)


conda-store-server/conda_store_server/server/app.py
```diff
@@ -200,6 +203,12 @@ def initialize(self, *args, **kwargs):
         # ensure checks on redis_url
         self.conda_store.redis_url
 
+        # ensure the worker was started and celery is available
+        # self.conda_store.celery_app
+        # import time
+        # time.sleep(5)
+        # assert self.conda_store.celery_app.control.inspect().ping() is not None
```
nkaretnikov, line 210: Debug, remove

conda-store-server/conda_store_server/server/app.py
```diff
@@ -340,16 +349,53 @@ async def favicon():
...
+        from conda_store_server import orm
+
         with self.conda_store.session_factory() as db:
             self.conda_store.ensure_settings(db)
             self.conda_store.ensure_namespace(db)
             self.conda_store.ensure_conda_channels(db)
 
+            # Needs to delete this entry to check if the worker is available
+            db.query(orm.Worker).delete()
+            db.commit()
```
nkaretnikov, line 397: This needs to happen on the server side before the worker is started to avoid a race condition. Moreover, it needs to happen on the server side because it will work in case the server is terminated and there are existing entries in the DB from previous runs.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -340,16 +349,53 @@ async def favicon():
...
+                        logger.info(f"{green}" "Worker initialized" f"{reset}")
+                        break
+
+                time.sleep(5)
+                logger.critical(
+                    f"{red}"
+                    "Waiting for worker... "
+                    "Use --standalone if running outside of docker"
+                    f"{reset}"
+                )
```
nkaretnikov, line 381: Added a signal handler because uvicorn ignores all signals in the start event. You wouldn't be able to terminate this loop otherwise.

conda-store-server/conda_store_server/server/app.py
nkaretnikov: I need to add this as a comment in the code

conda-store-server/conda_store_server/server/app.py
nkaretnikov: Also add a comment for this

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -46,6 +47,19 @@ def worker(self):
         return self._worker
 
 
+# Signals to the server that the worker is running, see startup_event in
+# CondaStoreServer.init_fastapi_app
```
nkaretnikov, line 51: This comment needs to be updated. I've renamed the functions

conda-store-server/conda_store_server/server/app.py
```diff
@@ -343,13 +346,73 @@ async def favicon():
...
```
nkaretnikov, line 406: Try using `self.log` here if it works

conda-store-server/conda_store_server/server/app.py
```diff
@@ -28,6 +28,9 @@
 )
 from traitlets.config import Application, catch_config_error
 
+logger = logging.getLogger("app")
+logger.setLevel(logging.INFO)
```
nkaretnikov, line 32: This needs to be updated to use `basicConfig` because this uses the implicit handler.

nkaretnikov:
Tested this on Linux and Windows. Example output from Windows (server and worker are running in separate terminals):

Screen Shot 2023-12-18 at 23 25 02

conda-store-server/conda_store_server/server/app.py
```diff
@@ -343,13 +343,73 @@ async def favicon():
...
+            #
+            # The database is used for worker-server communication because it
+            # will work regardless of celery_broker_url used, which can be Redis
+            # or just point to a database connection.
+            db.query(orm.Worker).delete()
+            db.commit()
+
+        def check_worker():
+            # Uses colors to make the output more visible
+            green = "\x1b[32m"
```
dcmcand, line 377: I feel like these should be constants. There are likely going to be other places that we want to use those colors and not redefining them would be nice.

conda-store-server/conda_store_server/server/app.py
```diff
@@ -343,13 +343,73 @@ async def favicon():
         return app
 
     def start(self):
+        import time
```
dcmcand, line 346: Why are you importing these here instead of with all the other imports?

conda-store-server/conda_store_server/server/app.py
```diff
@@ -343,13 +343,73 @@ async def favicon():
...
+            # Waits in a loop for the worker to become ready, which is
+            # communicated by the worker via task_initialize_worker
+            while True:
+                with session_factory() as db:
+                    q = db.query(orm.Worker).first()
+                    if q is not None and q.initialized:
+                        self.log.info(f"{green}" "Worker initialized" f"{reset}")
+                        break
+
+                time.sleep(5)
```
dcmcand, line 397: It would be nice to have this be passed in as a parameter with a default value so that for testing we could make the time shorter.

conda-store-server/conda_store_server/orm.py
```diff
@@ -41,6 +42,22 @@
...
+    __tablename__ = "worker"
+
+    id = Column(Integer, primary_key=True)
+
+    # For checking whether the worker is initialized
+    initialized = Column(Boolean, default=False)
+
+    __table_args__ = (
+        # Ensures no duplicates can be added with this combination of fields.
+        UniqueConstraint("initialized", name="_uc_worker"),
```
dcmcand, line 57: This unique constraint means that if you ever initialize more than one worker, then you get sql errors. You can try this by starting two workers and you will see:
'''
[2024-01-05 09:16:14,152: ERROR/ForkPoolWorker-3] Task task_initialize_worker[62967c9a-aca7-4f00-b056-15ec3a606507] raised unexpected: IntegrityError('(sqlite3.IntegrityError) UNIQUE constraint failed: worker.initialized')
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: worker.initialized

The above exception was the direct cause of the following exception:

  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
  File "/home/chuck/devel/conda-store/conda-store-server/conda_store_server/worker/tasks.py", line 61, in task_initialize_worker
    db.commit()
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 1451, in commit
    self._transaction.commit(_to_root=self.future)
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 829, in commit
    self._prepare_impl()
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 808, in _prepare_impl
    self.session.flush()
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 3444, in flush
    self._flush(objects)
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 3583, in _flush
    with util.safe_reraise():
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/util/langhelpers.py", line 70, in __exit__
    compat.raise_(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/util/compat.py", line 211, in raise_
    raise exception
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 3544, in _flush
    flush_context.execute()
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/unitofwork.py", line 456, in execute
    rec.execute(self)
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/unitofwork.py", line 630, in execute
    util.preloaded.orm_persistence.save_obj(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/persistence.py", line 245, in save_obj
    _emit_insert_statements(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/orm/persistence.py", line 1238, in _emit_insert_statements
    result = connection._execute_20(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1705, in _execute_20
    return meth(self, args_10style, kwargs_10style, execution_options)
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 334, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1572, in _execute_clauseelement
    ret = self._execute_context(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1943, in _execute_context
    self._handle_dbapi_exception(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2124, in _handle_dbapi_exception
    util.raise_(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/util/compat.py", line 211, in raise_
    raise exception
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1900, in _execute_context
    self.dialect.do_execute(
  File "/home/chuck/.micromamba/envs/conda-store-server/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: worker.initialized
[SQL: INSERT INTO worker (initialized) VALUES (?)]
[parameters: (1,)]
(Background on this error at: https://sqlalche.me/e/14/gkpj)
'''

I suggest a better approach would be to remove the unique constraint and simply query the worker table for initialized = true. If there is at least one row that is returned then would can proceed.

dcmcand (changes requested):

conda-store-server/conda_store_server/server/app.py
```diff
@@ -350,6 +387,30 @@ def start(self):
             self.conda_store.ensure_namespace(db)
             self.conda_store.ensure_conda_channels(db)
 
+            # This ensures the database has no Worker table entires when the
```
nkaretnikov, line 390: Typo: entires, will fix

nkaretnikov:
@dcmcand Please take a look!

Summary of changes:
- The constants and method are accessible globally, but they are marked as private to avoid polluting the public namespace.
- The time delay value is now a parameter.
- Removed the DB constraint, too.
- Fixed a typo.

dcmcand (approved):
:rocket: Thanks @nkaretnikov

PR #709, opened 2023-12-17
Terminate worker tasks on SIGINT

### LINKED ISSUE #707: Terminate worker subprocess in standalone mode
### Context

When running with `--standalone`, conda-store spins up a celery worker subprocess. That subprocess is not terminated when you Ctrl-C the main process. This needs to be done in a way that would work on Linux, Windows, macOS.

https://github.com/conda-incubator/conda-store/issues/513#issuecomment-1766776871

> If we do want it, it's possible to forcibly control-c out of the standalone server in a way that leaves the worker processes intact. See Revisit the implementation of --standalone as well as the assumptions made #513 (comment). However, I've also noticed that this is somewhat of an issue with celery itself, even when the worker is run separately, that cancellation doesn't work very smoothly. The difference is that when workers are run separately, the process doesn't end until the workers are actually shutdown. Not clear how important this issue is. It comes up during development, but I don't know if it comes up in deployment.

### Value and/or benefit

Improved UX.

### Anything else?

_No response_

nkaretnikov:
Fixes #707.

## Description

This pull request: installs a signal handler that terminates worker tasks on Ctrl-C (SIGINT).

If the requested signal isn't supported on the current platform, the operation will be ignored.

Notes:
- we cannot use celery control because it's broker-dependent, which wouldn't work with standalone
- printing a worker pid and asking users to terminate that wouldn't work because celery beat would respawn
- using `install_worker_term_hard_handler` wouldn't work because not all tasks would be terminated
- using a signal handler is the only solution I've found given the above constraints, it's what `install_worker_term_hard_handler` uses internally as well.

Tested on Windows and Linux. All processes are terminated as expected.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/worker/tasks.py
+++ b/conda-store-server/conda_store_server/worker/tasks.py
@@ -1,10 +1,11 @@
 import datetime
 import os
 import shutil
+import sys
 import typing
 
 import yaml
-from celery import Task, shared_task
+from celery import Task, platforms, shared_task
 from celery.execute import send_task
 from celery.signals import worker_ready
 from conda_store_server import api, environment, schema, utils
@@ -43,6 +44,17 @@ class WorkerTask(Task):
                 self._worker.update_config(config)
 
             self._worker.initialize()
+
+        # Installs a signal handler that terminates the process on Ctrl-C
+        # (SIGINT)
+        # Note: the following would not be enough to terminate beat and other
+        # tasks, which is why the code below explicitly calls sys.exit
+        # > from celery.apps.worker import install_worker_term_hard_handler
+        # > install_worker_term_hard_handler(self._worker, sig='SIGINT')
+        def _shutdown(*args, **kwargs):
+            return sys.exit(1)
+
+        platforms.signals["INT"] = _shutdown
 
         return self._worker
 


conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -43,6 +44,17 @@ def worker(self):
...
             self._worker.initialize()
 
+        # Installs a signal handler that terminates the process on Ctrl-C
+        # (SIGINT)
+        # Note: the following would not be enough to terminate beat and other
+        # tasks, which is why the code below explicitly calls sys.exit
+        # > from celery.apps.worker import install_worker_term_hard_handler
+        # > install_worker_term_hard_handler(self._worker, sig='SIGINT')
+        def _shutdown(*args, **kwargs):
+            return sys.exit(1)
```
dcmcand, line 54: should this be exit value of 1? A requested exit that exits cleanly should be a zero exit status (the default in this case)

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -43,6 +44,17 @@ def worker(self):
...
+
+        platforms.signals["INT"] = _shutdown
```
dcmcand, line 56: is sigint the only signal we need to worry about here? Do we face similar issues with sigterm or sigkill? Do those need handled too?

conda-store-server/conda_store_server/worker/tasks.py
nkaretnikov: It *should* be non-zero, yes. Despite being user-requested, it's still abnormal termination since tasks are immediately terminated mid-execution, which can have side-effects (like partially downloaded files, etc.).

The reason I terminate tasks immediately is because some of them might take a very long time to terminate cleanly and that use case is already [handled by SIGTERM in Celery](https://docs.celeryq.dev/en/stable/userguide/workers.html#stopping-the-worker) anyway. Waiting for tasks to terminate is not something I want on Ctrl-C.

Because of how we start Celery via multiprocessing and Celery additionally spawning processes on its own, the current mechanism is the only way I've found that doesn't leave stale processes.

Going back to the exit code, this is up to the application to define. I use `exit 1` because it's a good default as we don't have a map of exit codes (I don't think it's very important for conda-store because it's not a script that will be used in shell pipelines).

For example, `ls` has these exit statuses (via `man ls`):

```
   Exit status:
       0      if OK,
       1      if minor problems (e.g., cannot access subdirectory),
       2      if serious trouble (e.g., cannot access command-line argument).
```

Another example, `conda` returns 1 on Ctrl-C:
```
$ conda install constructor
Retrieving notices: ...working... done
^C
CondaError: KeyboardInterrupt

$ echo $?
1
```

conda-store-server/conda_store_server/worker/tasks.py
nkaretnikov: I don't think we need to touch those. `SIGINT` is special because it can be sent via Ctrl-C. The two signals you mention need to be sent via something like `kill` or `pkill`, at which point the user needs to provide a proper process name or process id. Also, Celery has [pre-defined handlers for SIGTERM and SIGKILL](https://docs.celeryq.dev/en/stable/userguide/workers.html#stopping-the-worker), which correspond to soft and cold shutdowns, so I don't want to override that behavior.

dcmcand (approved):

trallard:
@nkaretnikov there is a merge conflict - can you fix this and we can merge right away

PR #713, opened 2023-12-19
Add extended-length prefix support

### LINKED ISSUE #588: Figure out how to handle long paths on Windows
### Context

This was originally discussed at https://github.com/conda-incubator/conda-store/pull/559#issuecomment-1721915623

Windows support is still in progress (#559), but an issue that comes up is that Windows by default has a filesystem path length limit of 260 characters. You end up with errors like

```
```

See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

It is possible to remove this limitation in Windows 10 and later if you set a registry value https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later. However, this (presumably) requires administrator privileges so I don't know if it's a viable thing to require. 

The exact length of the path depends on the specific conda package. However, it's pretty obvious that the issue is the long UUID generated by conda-store: `99108419ad0fd922fdeff9bbc434b58d41f68e3f923a83f6a7ab19568463bc82-20230915-201619-848782-1-test`. That's 90 characters (plus the 4 character environment name). Shortening or removing this UUID from the path would solve the issue for this specific package. One could also store the environment closer to the root (see also #554). However, I'm not really sure what the longest "real world" path in a conda package is. The conda issue mentions that this has been fixed upstream in many conda-forge packages with longer paths, but I don't know just how long of a path is considered "OK" for a Windows conda package, but it's still something that could pop up even if we fix both of those things. At that point, that would probably be considered an upstream issue with conda and/or the specific package, but it's worth being aware of this problem. 

This is a known issue in conda. See https://github.com/conda/conda/issues/7203.

Based on this comment, https://github.com/conda/conda/issues/7203#issuecomment-1722123864, if we want to try to make the path no longer than a typical conda install default, the part of the path between `C:\Users\Username` and the individual package files (like `\Lib\site-packages\...`) should be no longer than `\AppData\Local\conda\conda\pkgs`, which is 31 characters. But `\.conda-store\conda-store-state`, as proposed in #554, is already 31 characters. 

It might be possible to try a scheme where we store packages in paths with very short hashes. But otherwise, we might just have to require people to set that registry key if they want to use things on Windows. 

### Value and/or benefit

Some packages inherently have long paths in them, and this would make them unusable on Windows. I haven't done a thorough investigation of how long the longest paths are on various packages on Windows, but this could be something that could be done. 

### Anything else?

Regardless of what we do, we need to make sure error messages are properly shown to the user (but this is sort of moot at this point because Windows support itself is still not truly functional).

nkaretnikov:
Fixes #588.

## Description

This pull request:

- adds support for the Windows [extended-length prefix](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) (`"\\?\"`) via the `win_extended_length_prefix` option
- adds a test
- updates the docs.

## Pull request checklist

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -62,7 +62,20 @@ jobs:
 
       - name: "Unit tests ✅"
         run: |
-          pytest tests
+          pytest -m "not extended_prefix" tests
+
+      # https://github.com/actions/runner-images/issues/1052
+      - name: "Windows extended prefix unit tests ✅"
+        shell: pwsh
+        run: |
+          Set-ItemProperty "HKLM:\System\CurrentControlSet\Control\FileSystem" `
+            -Name "LongPathsEnabled" `
+            -Value 0 `
+            -Type DWord
+          (Get-ItemProperty "HKLM:System\CurrentControlSet\Control\FileSystem").LongPathsEnabled
+          pytest -m "extended_prefix" tests
+        if: matrix.os == 'windows'
+
 
   integration-test-conda-store-server:
     name: "integration-test conda-store-server"
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -120,6 +120,12 @@ class CondaStore(LoggingConfigurable):
         except Exception as e:
             raise TraitError(f"c.CondaStore.build_key_version: {e}")
 
+    win_extended_length_prefix = Bool(
+        False,
+        help="Use the extended-length prefix '\\\\?\\' (Windows-only), default: False",
+        config=True,
+    )
+
     conda_command = Unicode(
         "mamba",
         help="conda executable to use for solves",
--- a/conda-store-server/conda_store_server/orm.py
+++ b/conda-store-server/conda_store_server/orm.py
@@ -4,6 +4,7 @@ import os
 import pathlib
 import re
 import shutil
+import sys
 
 from conda_store_server import conda_utils, schema, utils
 from conda_store_server.environment import validate_environment
@@ -254,7 +255,12 @@ class Build(Base):
         # https://github.com/conda-incubator/conda-store/issues/649
         if len(str(res)) > 255:
             raise BuildPathError("build_path too long: must be <= 255 characters")
-        return res
+        # Note: cannot use the '/' operator to prepend the extended-length
+        # prefix
+        if sys.platform == "win32" and conda_store.win_extended_length_prefix:
+            return pathlib.Path(f"\\\\?\\{res}")
+        else:
+            return res
 
     def environment_path(self, conda_store):
         """Environment path is the path for the symlink to the build
@@ -264,11 +270,17 @@ class Build(Base):
         store_directory = os.path.abspath(conda_store.store_directory)
         namespace = self.environment.namespace.name
         name = self.specification.name
-        return pathlib.Path(
+        res = pathlib.Path(
             conda_store.environment_directory.format(
                 store_directory=store_directory, namespace=namespace, name=name
             )
         )
+        # Note: cannot use the '/' operator to prepend the extended-length
+        # prefix
+        if sys.platform == "win32" and conda_store.win_extended_length_prefix:
+            return pathlib.Path(f"\\\\?\\{res}")
+        else:
+            return res
 
     @property
     def build_key(self):
--- a/conda-store-server/tests/test_server.py
+++ b/conda-store-server/tests/test_server.py
@@ -449,7 +449,7 @@ def test_create_specification_auth_extended_prefix(win_extended_length_prefix, t
         r = schema.APIGetBuild.parse_obj(response.json())
         assert r.status == schema.APIStatus.OK
         assert r.data.specification.name == environment_name
-        if r.data.status in ["QUEUED", "BUILDING"]:
+        if r.data.status in ("QUEUED", "BUILDING"):
             continue  # checked too fast, try again
 
         if win_extended_length_prefix:
@@ -465,7 +465,7 @@ def test_create_specification_auth_extended_prefix(win_extended_length_prefix, t
 
     # If we're here, the task didn't update the status on failure
     if not is_updated:
-        assert False, f"failed to get status"
+        assert False, "failed to update status"
 
 
 def test_create_specification_auth(testclient, celery_worker, authenticate):
--- a/conda-store-server/tests/test_server.py
+++ b/conda-store-server/tests/test_server.py
@@ -3,6 +3,7 @@ import sys
 import time
 
 import pytest
+import traitlets
 import yaml
 from conda_store_server import __version__, schema
 
@@ -400,9 +401,12 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker, 
 @pytest.fixture
 def win_extended_length_prefix(request):
     # Overrides the attribute before other fixtures are called
-    import conda_store_server
-    conda_store_server.app.CondaStore.win_extended_length_prefix = request.param
+    from conda_store_server.app import CondaStore
+    assert type(CondaStore.win_extended_length_prefix) is traitlets.Bool
+    old_prefix = CondaStore.win_extended_length_prefix.default_value
+    CondaStore.win_extended_length_prefix.default_value = request.param
     yield request.param
+    CondaStore.win_extended_length_prefix.default_value = old_prefix
 
 
 @pytest.mark.skipif(sys.platform != "win32", reason="tests a Windows issue")
--- a/conda-store-server/tests/test_server.py
+++ b/conda-store-server/tests/test_server.py
@@ -1,4 +1,5 @@
 import json
+import sys
 import time
 
 import pytest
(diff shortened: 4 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -62,7 +62,20 @@ jobs:
...
+      - name: "Windows extended prefix unit tests ✅"
+        shell: pwsh
+        run: |
+          Set-ItemProperty "HKLM:\System\CurrentControlSet\Control\FileSystem" `
+            -Name "LongPathsEnabled" `
+            -Value 0 `
+            -Type DWord
+          (Get-ItemProperty "HKLM:System\CurrentControlSet\Control\FileSystem").LongPathsEnabled
+          pytest -m "extended_prefix" tests
+        if: matrix.os == 'windows'
```
nkaretnikov, line 77: This has to be run separately because `LongPathsEnabled=0` causes other tests to fail on CI due to long prefixes.

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
...
+    response.raise_for_status()
+
+    r = schema.APIPostSpecification.parse_obj(response.json())
+    assert r.status == schema.APIStatus.OK
+    build_id = r.data.build_id
+
+    # Try checking that the status is 'FAILED'
+    is_updated = False
+    for _ in range(30):
+        time.sleep(5)
```
nkaretnikov, line 443: Time and range were chosen to give enough time for the build to complete.

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
...
+
+        # check for the given build
+        response = testclient.get(f"api/v1/build/{build_id}", timeout=30)
+        response.raise_for_status()
+
+        r = schema.APIGetBuild.parse_obj(response.json())
+        assert r.status == schema.APIStatus.OK
+        assert r.data.specification.name == environment_name
+        if r.data.status in ["QUEUED", "BUILDING"]:
```
nkaretnikov, line 452: Nit: use a tuple

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
...
+            response = testclient.get(f"api/v1/build/{build_id}/logs", timeout=30)
+            response.raise_for_status()
+            assert "[WinError 206] The filename or extension is too long" in response.text
+
+        is_updated = True
+        break
+
+    # If we're here, the task didn't update the status on failure
+    if not is_updated:
+        assert False, f"failed to get status"
```
nkaretnikov, line 468: Nit: don't use a format string

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
...
+
+
```
nkaretnikov, line 470: Note for reviewers: the test above is based on `test_create_specification_auth_env_name_too_long`. I deliberately didn't move common bits into a separate function because I think the current version is more readable

conda-store-server/conda_store_server/orm.py
```diff
@@ -254,7 +255,12 @@ def build_path(self, conda_store):
         # https://github.com/conda-incubator/conda-store/issues/649
         if len(str(res)) > 255:
             raise BuildPathError("build_path too long: must be <= 255 characters")
-        return res
+        # Note: cannot use the '/' operator to prepend the extended-length
```
jaimergp, line 258: Hm, this is surprising. Why? Do you have a reference? I would have assumed that this is ok:

```python
pathlib.Path(r"\\?\", res)
```

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
         assert False, f"failed to update status"
 
 
+@pytest.fixture
+def win_extended_length_prefix(request):
+    # Overrides the attribute before other fixtures are called
+    import conda_store_server
+    conda_store_server.app.CondaStore.win_extended_length_prefix = request.param
+    yield request.param
```
jaimergp, line 405: Should we revert to the previous value after the `yield`?

conda-store-server/tests/test_server.py
```diff
@@ -396,6 +397,77 @@ def test_create_specification_auth_env_name_too_long(testclient, celery_worker,
...
+
+@pytest.mark.skipif(sys.platform != "win32", reason="tests a Windows issue")
+@pytest.mark.parametrize('win_extended_length_prefix', [True, False], indirect=True)
+@pytest.mark.extended_prefix
+def test_create_specification_auth_extended_prefix(win_extended_length_prefix, testclient, celery_worker, authenticate):
+    # Adds padding to cause an error if the extended prefix is not enabled
+    namespace = "default" + 'A' * 10
+    environment_name = "pytest"
+
+    # The debugpy 1.8.0 package was deliberately chosen because it has long
```
jaimergp, line 416: Awesome find :)

docusaurus-docs/conda-store/references/faq.md
```diff
@@ -115,29 +115,60 @@ c.CondaStore.build_key_version = 2
...
 ERROR:root:[WinError 206] The filename or extension is too long: 'C:\\...'
 ```
 
 This error is due to the fact that Windows has a limitation that file paths
-cannot be more than 260 characters. The fix is to set the registry key
+cannot be more than 260 characters.
+
+See [conda-store issue #588][max-path-issue] for more details.
+
+### Solution 1: Extended-length path prefix (`\\?\`)
```
jaimergp, line 129: I thought solution (2) is required to enable solution (1) too? Maybe I am misremembering, though.

conda-store-server/conda_store_server/orm.py
nkaretnikov: Hmm, I cannot reproduce this anymore for some reason! Not sure what I've tried. E.g., this works fine:

```py
>>> str("\\\\?\\" / pathlib.Path("foo/bar"))
'\\\\?\\foo\\bar'
```

docusaurus-docs/conda-store/references/faq.md
nkaretnikov: Nope! That's the whole point of using the extended-length path prefix. Note how I set `LongPathsEnabled=0` in the CI job, too.

conda-store-server/conda_store_server/orm.py
nkaretnikov: Ah, I see what the problem is. In my comment above, the `Path` part is a relative path. But in the above code, `res` is absolute and it just ignores the prefix (I've printed the value to check):

```py
>>> str("\\\\?\\" / pathlib.Path("foo"))
'\\\\?\\foo'  # ok, but Path is relative
>>> str("\\\\?\\" / pathlib.Path("C:\\foo")) 
'C:\\foo'  # ignored
>>> str("\\\\?\\" / pathlib.WindowsPath("C:\\foo")) 
'C:\\foo'  # ignored (it's WindowsPath on Windows)
```

Same with your variant:

```py
>>> str(pathlib.Path("\\\\?\\", pathlib.Path("foo")))
'\\\\?\\foo'  # ok, but Path is relative
>>> str(pathlib.Path("\\\\?\\", pathlib.Path("C:\\foo")))
'C:\\foo'  # ignored
>>> str(pathlib.Path("\\\\?\\", pathlib.WindowsPath("C:\\foo")))
'C:\\foo'  # ignored
```

So I'll keep my current version with string interpolation.

nkaretnikov:
@jaimergp PTAL. I'll rebase after you approve since a rebase will make it harder to review the recent changes, which is just a fixture update.

conda-store-server/conda_store_server/orm.py
jaimergp: Ah, I see. Actually, it [looks like](https://stackoverflow.com/a/13426839) `\\?\` can _only_ be used with absolute paths. So maybe we want to add a .resolve() call below.

jaimergp: nvm it's abspath'd above. Marking this as resolved. Thanks!

conda-store-server/conda_store_server/app.py
```diff
@@ -120,6 +120,12 @@ def _check_build_key_version(self, proposal):
         except Exception as e:
             raise TraitError(f"c.CondaStore.build_key_version: {e}")
 
+    win_extended_length_prefix = Bool(
+        False,
+        help="Use the extended-length prefix '\\\\?\\' (Windows-only), default: False",
```
jaimergp, line 125: In general I prefer using `r''` strings whenever I have backslashes in the string, but this kind of thing is better enforced by linters and style formatters rather than code reviews. Only leaving it as a statement of preference for a potential discussion down the line, not asking for a change now.

```suggestion
        help=r"Use the extended-length prefix '\\?\' (Windows-only), default: False",
```

Issue #71, opened 2021-06-24
Handle multiple OS/architectures builders for workers


jaimergp:
There are some places where os/arch is assumed to be `linux-64`. For example, [lockfiles](https://github.com/Quansight/conda-store/blob/main/conda-store-server/conda_store_server/api.py#L82).

AFAIK there's no way to tell `conda-store` which os/arch should be targeted either. It would just assume it is the one running in the underlying Docker image for `conda-store-server`.

costrouc:
I think that it is important to support arm64 and x86. However I don't see much value at the moment of supporting osx since the main motivation for this work is allowing dockerized environments and linux hpc environments access to conda environments. Going to put in the icebox.

jaimergp:
Agree that `osx` is not priority at all. But `aarch64` and `ppc64le` might be needed at some point. With #77 once can get those environments as long as `conda-store-server` is running on those archs, but #67 forces the Dockerfiles to be x64.

trallard:
Related to #507

asmeurer:
What exactly is the issue to be done here? With #539 conda-store in standalone mode works on macOS, and generates lockfiles with osx-64. I don't have an ARM mac to test, but it would presumably work there too. The Docker generation is disabled on Mac, because conda-docker does not work on Mac. Is the point here to be able to generate environments for architectures different from the one conda-store is running in?

costrouc:
@asmeurer changed the name of the issue since I think this issue is independant from the standalone work. But I do think that working on the standalone work makes this work much easier.

I'm not sure how to test or do this but I would like to have a separate celery worker/queue per OS/architecture. This would allow us to truely build environments for all OS's. I think this is a hard issue.

asmeurer:
I see. Presumably, for most actions, you could just build an environment for a different platform than the one the worker is running on by setting the appropriate environment variables/flags to conda to make it use a different arch. Although there are definitely some actions that just won't work with, e.g., what comes to mind would be

- Building a docker container (because of the conda-docker issues noted elsewhere)
- pip installing anything that involves compilation. Presumably building a cross-platform env with a pure Python .tar.gz or a wheel is fine, although I don't actually know how to do it. 

And possibly some other things as well. Has this sort of thing already been considered?

For actually having the workers run on multiple OSes, the obvious blocker is #507. macOS support should be working now (sans Docker container building). Windows support is next but will probably be harder. 

[...]

PR #714, opened 2023-12-21
Generate `constructor` artifacts

### LINKED ISSUE #657: [ENH] - Teach conda-store how to generate `constructor` artifacts
### Feature description

conda-store is able to generate several artifacts from an existing environment:

- Docker environment
- Lockfile
- conda-pack archive

`constructor` is able to generate more formats, like self-contained installer scripts for Unix, or native macOS or Windows step-by-step GUI installers.

This would need extra configuration fields that `constructor` expects.

### Value and/or benefit

More diversity of exported formats, which would come handy for application distribution, opening the doors for new personas in conda-store (e.g. product manager that needs to generate installers for the latest release of their app).

### Anything else?

This is part of the STF funding, deliverable 4.

It will also allow to use a hypothetically unified interface for all types of "environment artifacts" (e.g. see https://github.com/conda/conda-pack/issues/294), so we can simplify our integrations.

nkaretnikov:
Fixes #657.

## Description

This pull request:

- adds `action_generate_constructor_installer`, which generates `sh` and `exe` installers via `constructor`
- exposes this in the API via `/api/v1/build/{build_id}/installer/`
- adds tests
- updates the docs.

## Pull request checklist


conda-store-server/conda_store_server/action/generate_constructor_artifacts.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Creates the construct.yaml file and post_install script
+    ext = ".exe" if sys.platform == "win32" else ".sh"
+    pi_ext = ".bat" if sys.platform == "win32" else ".sh"
+    installer_dir = pathlib.Path(installer_dir)
+    installer_filename = (installer_dir / specification.name).with_suffix(ext)
+
+    os.makedirs(installer_dir)
+
+    with tempfile.TemporaryDirectory(ignore_cleanup_errors=True) as tmp_dir:
```
nkaretnikov, line 50: ignore_cleanup_errors=True

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+        # Installs constructor
+        command = [
+            conda_command,
+            "create",
+            "-y",
+            "-p",
+            str(env_dir),
+            "constructor",
+        ]
+        print_cmd(command)
```
nkaretnikov, line 94: It _is_ wasteful to download `constructor` every time, but I do it for the following reasons:
- I don't want to assume that `constructor` is already available in the user environment, so I need to install it
- I'm not sure I can rely on conda-store installation method to provide runtime deps. E.g., if users install conda-store via `pip`, how would they get `constructor`?
- This needs to be installed into a separate env to avoid clashing with user packages in standalone mode
- This ensures the latest `constructor` version is used, without requiring manual intervention
- If a failure occurs for whatever reason, the next call to this action will try again (actions are state-independent).

I think we can try the following in a _separate PR_, which I didn't implement due to time constraints:
- have a predefined env name, which is unlikely to clash with user envs, like `conda-store-runtime`, and installing `constructor` there on first call of this action
- then checking if it's already available in the env by calling `constructor --help`
- if a new version is available upstream, downloading and installing that one
- if `--help` fails, I think the entire environment needs to be removed because it might be in an invalid state (a user hit Ctrl-C while this action was running, network error, etc.)
- removing the env needs to be done via python APIs for dir removal and not via `conda env remove` because I had issues with it where it failed to remove an existing env.

nkaretnikov: Actually, we might not want to update automatically just to avoid bumping into unexpected issues. Instead, we could pin the constructor version here and update it ourselves once in a while if new features are needed.

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -63,6 +64,7 @@ async def ui_list_environments(
             "environments": orm_environments.all(),
             "registry_external_url": server.registry_external_url,
             "entity": entity,
+            "platform": "Windows" if sys.platform == "win32" else "Linux, macOS",
```
nkaretnikov, line 67: This needs to be updated because Linux and macOS installers are not compatible: cannot use a Linux installer on macOS

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -208,6 +210,7 @@ async def ui_get_build(
             "registry_external_url": server.registry_external_url,
             "entity": entity,
             "spec": yaml.dump(build.specification.spec),
+            "platform": "Windows" if sys.platform == "win32" else "Linux, macOS",
```
nkaretnikov, line 213: This needs to be updated because Linux and macOS installers are not compatible: cannot use a Linux installer on macOS

nkaretnikov:
Tested this manually on Linux (x86_64), Windows (ARM64), macOS (x86_64):
- I was able to build the installer for the following env (uses conda and pip packages)
- After executing `python` from the install dir, it was able to find both Python packages.

```yaml
channels:
- conda-forge
dependencies:
- flask
- pip:
  - alembic
description: test
name: test
prefix: null
variables: null
```

Screenshots from Windows:

Screen Shot 2023-12-25 at 04 08 16

Screen Shot 2023-12-25 at 03 09 36

Screen Shot 2023-12-25 at 06 51 22

nkaretnikov:
Updated to use `sys.platform` directly. I think the values it returns are descriptive (darwin, win32, linux) and it saves the need to write a custom helper, which might need to be updated in the future.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+    specification: schema.CondaSpecification,
+    installer_dir: pathlib.Path,
+    version: str,
+):
+    # Helpers
+    def print_cmd(cmd):
+        context.log.info(f"Running command: {' '.join(cmd)}")
+        context.log.info(
+            subprocess.check_output(cmd, stderr=subprocess.STDOUT, encoding="utf-8")
+        )
```
nkaretnikov, line 24: Same helper as in `conda-store-server/conda_store_server/action/generate_lockfile.py`

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+            "name": specification.name,
+            "channels": specification.channels,
+            "specs": dependencies,
+            "version": version,
+        }
+
+        if sys.platform == "win32":
+            post_install = """\
+call "%PREFIX%\Scripts\activate.bat"
+"""
```
nkaretnikov, line 68: Needs to be:

```py
call "%PREFIX%\Scripts\\activate.bat"
```

Since `\a` is a special character.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            post_install = """\
+#!/usr/bin/env bash
+set -euxo pipefail
+source "$PREFIX/etc/profile.d/conda.sh"
+conda activate "$PREFIX"
+"""
```
nkaretnikov, line 75: See https://github.com/conda/constructor/blob/main/CONSTRUCT.md#post_install

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+        if pip_dependencies:
+            post_install += f"""
+pip install {' '.join(pip_dependencies)}
+"""
```
nkaretnikov, line 79: See https://github.com/conda/constructor/issues/515#issuecomment-1460610464

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+        else:
+            post_install = """\
+#!/usr/bin/env bash
+set -euxo pipefail
+source "$PREFIX/etc/profile.d/conda.sh"
+conda activate "$PREFIX"
+"""
+        if pip_dependencies:
+            post_install += f"""
+pip install {' '.join(pip_dependencies)}
```
nkaretnikov, line 78: Change to `python -m pip install` to match the comment below

conda-store-server/conda_store_server/build.py
```diff
@@ -382,3 +382,43 @@ def build_conda_docker(db: Session, conda_store, build: orm.Build):
...
+    ):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            context = action.action_generate_constructor_installer(
+                conda_command=settings.conda_command,
+                specification=schema.CondaSpecification.parse_obj(
+                    build.specification.spec
+                ),
+                installer_dir=pathlib.Path(tmpdir),
+                version=build.build_key,
+            )
```
nkaretnikov, line 407: Note for reviewers: the settings/specification part is similar to what's done in `build_conda_environment`.

The rest of this function is similar to `build_conda_pack` or `build_conda_docker ` in this file.

conda-store-server/conda_store_server/orm.py
```diff
@@ -313,6 +314,11 @@ def conda_pack_key(self):
     def docker_manifest_key(self):
         return f"docker/manifest/{self.build_key}"
 
+    @property
+    def constructor_installer_key(self):
+        ext = "exe" if sys.platform == "win32" else "sh"
+        return f"installer/{self.build_key}.{ext}"
```
nkaretnikov, line 320: Note for reviewers: this is similar to other keys in this file

conda-store-server/conda_store_server/orm.py
```diff
@@ -344,6 +350,13 @@ def has_docker_manifest(self):
             for artifact in self.build_artifacts
         )
 
+    @hybrid_property
+    def has_constructor_installer(self):
+        return any(
+            artifact.artifact_type == schema.BuildArtifactType.CONSTRUCTOR_INSTALLER
+            for artifact in self.build_artifacts
+        )
```
nkaretnikov, line 358: Note for reviewers: this is similar to other properties in this file

conda-store-server/conda_store_server/server/templates/build.html
```diff
@@ -79,6 +79,9 @@ <h3 class="card-title">Conda Environment Artifacts</h3>
         {% if build.has_conda_pack %}
         <li class="list-group-item"><a href="https://github.com/conda/conda-pack">Conda-Pack</a> archive: <a href="{{ url_for('api_get_build_archive', build_id=build.id) }}">environment.tar.gz</a></li>
         {% endif %}
+        {% if build.has_constructor_installer %}
+        <li class="list-group-item"><a href="https://github.com/conda/constructor">Constructor</a> installer: <a href="{{ url_for('api_get_build_installer', build_id=build.id) }}" download>installer ({{ platform }})</a></li>
+        {% endif %}
```
nkaretnikov, line 84: Note for reviewers: similar to other artifacts in this file

conda-store-server/conda_store_server/server/templates/home.html
```diff
@@ -41,6 +41,9 @@ <h5 class="card-title">
         {% if environment.current_build.has_docker_manifest %}
         <a class="card-link" onclick="setClipboard('{{ registry_external_url }}/{{ environment.namespace.name }}/{{ environment.name }}:{{ environment.current_build.build_key }}')" role="button" data-toggle="popover" data-trigger="hover focus" data-content="Click to copy!"><ion-icon name="logo-docker"></ion-icon> Docker</a>
         {% endif %}
+        {% if environment.current_build.has_constructor_installer %}
+        <a class="card-link" href="{{ url_for('api_get_build_installer', build_id=environment.current_build_id) }}" download><ion-icon name="archive-outline"></ion-icon> Installer ({{ platform }})</a>
+        {% endif %}
```
nkaretnikov, line 46: Note for reviewers: similar to other artifacts in this file

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -1279,6 +1280,34 @@ async def api_get_build_docker_image_url(
...
+        if build.has_constructor_installer:
+            return RedirectResponse(
+                conda_store.storage.get_url(build.constructor_installer_key)
+            )
+
+        else:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Build {build_id} doesn't have an installer",
+            )
```
nkaretnikov, line 1308: Note for reviewers: similar to other artifacts in this file

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -211,6 +212,14 @@ def task_build_conda_docker(self, build_id):
         build_conda_docker(db, conda_store, build)
 
 
+@shared_task(base=WorkerTask, name="task_build_constructor_installer", bind=True)
+def task_build_constructor_installer(self, build_id):
+    conda_store = self.worker.conda_store
+    with conda_store.session_factory() as db:
+        build = api.get_build(db, build_id)
+        build_constructor_installer(db, conda_store, build)
```
nkaretnikov, line 220: Note for reviewers: similar to other tasks in the file

conda-store-server/tests/test_actions.py
```diff
@@ -372,6 +424,59 @@ def lockfile_url(build_key):
...
+
+    # redirects to installer
+    def installer_url(build_key):
+        ext = "exe" if sys.platform == "win32" else "sh"
+        return f"installer/{build_key}.{ext}"
+
+    assert type(res) is RedirectResponse
+    assert build.constructor_installer_key == res.headers['location']
+    assert installer_url(build.build_key) == build.constructor_installer_key
+    assert res.status_code == 307
```
nkaretnikov, line 477: Note for reviewers: similar test to `test_api_get_build_lockfile`

conda-store-server/conda_store_server/action/generate_constructor_installer.py
jaimergp: I really don't like this. We should add it as a runtime dependency. If `pip install` compatibility is a problem, it is too for the `conda` dependency. Worst case we can try-import it and disable the option if `constructor` is not available. It will save tons of complexity we don't really need.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,108 @@
...
+            post_install = """\
+#!/usr/bin/env bash
+set -euxo pipefail
+source "$PREFIX/etc/profile.d/conda.sh"
+conda activate "$PREFIX"
+"""
+        if pip_dependencies:
+            post_install += f"""
+python -m pip install {' '.join(pip_dependencies)}
+"""
```
jaimergp, line 79: I don't think we need a triple-backtick string here?

conda-store-server/conda_store_server/action/generate_constructor_installer.py
nkaretnikov: > Worst case we can try-import it and disable the option if constructor is not available.

I like this suggestion! This solves the compatibility/distribution problem while still making this feature available.

conda-store-server/conda_store_server/orm.py
jaimergp: We can also have PKG installers on macOS. Is this handled?

jaimergp:
> Updated to use `sys.platform` directly in the UI. I think the values it returns are descriptive (darwin, win32, linux) and it saves the need to write a custom helper, which might need to be updated in the future.
> 
> Screenshot 2023-12-25 at 07 14 36 Screen Shot 2023-12-25 at 07 16 12 Screen Shot 2023-12-25 at 07 13 31

This assumes we are always building for the current (native) platform, but constructor supports cross-building to different OS and architectures. I'm not sure we can just assume this. I'd also like if we could print the architecture? Some folks might be unaware they are building ARM installers if their server runs on aarch64, etc.

jaimergp:
Btw, great job researching the integrations here. constructor is not super intuitive :)

conda-store-server/conda_store_server/orm.py
nkaretnikov: > We can also have PKG installers on macOS. Is this handled?

It's handled by constructor, but this PR generates .sh on Linux and macOS. .sh is the default installer type on Linux and macOS in constructor.

It makes everything either if you have one installer per platform (less things to test, one endpoint for all installers on the API side). Do you think it's still worth it to have two installers on macOS?

conda-store-server/conda_store_server/action/generate_constructor_installer.py
nkaretnikov: Technically, we don't. I just find it easier to read because you don't need to insert newlines at the start and end of the string.

nkaretnikov:
> I'd also like if we could print the architecture? Some folks might be unaware they are building ARM installers if their server runs on aarch64, etc.

I agree, I'll make this change.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
jaimergp: Personal preference I guess. I like `"\n...\n"` better, but it's not important.

conda-store-server/conda_store_server/orm.py
jaimergp: We don't _need_ it but at some point I am sure someone will end up asking for them (they can be branded, they feel more "OS-native", etc). Just wanted to make sure this assumption was made knowing there's a non-sh installer on macOS.

nkaretnikov:
Note: marking as a draft as I still need to test the migration. SQLite doesn't support `ALTER COLUMN` and `enum`s (they are simply represented as `VARCHAR`) while `enum`s are immutable in Postgres, which makes it tricky to migrate both DBs.

nkaretnikov:
Status update:

Need to re-read the code again. If no issues there, I'll open for review.

Tested on 7969569:
- SQLite: upgrade/downgrade migration works, installer works
- Postgres: upgrade/downgrade migration works, installer works
- A warning is shown in console if constructor is not available
- Tested in Docker as well on Linux
- Windows: installer works in standalone mode
- macOS: installer works in standalone mode.

New screenshots:
- macOS
Screen Shot 2024-01-17 at 12 01 04
Screen Shot 2024-01-17 at 12 00 56

- Windows
Screen Shot 2024-01-17 at 11 29 31
Screen Shot 2024-01-17 at 11 38 44

- Linux
Screen Shot 2024-01-17 at 09 57 02
Screen Shot 2024-01-17 at 09 57 10

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,127 @@
...
+        command = [
+            "constructor",
+            "--help",
+        ]
+        print_cmd(command)
+    except FileNotFoundError:
+        warnings.warn(
+            "Installer generation requires constructor: https://github.com/conda/constructor"
+        )
+        return
```
nkaretnikov, line 52: I now check whether constructor is available instead of creating a temporary environment and installing it every time.

constructor is part of the conda environment provided in the repo, so it'd be available in Docker, etc.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,127 @@
...
+        command = [
+            "constructor",
+            "-v",
+            "--cache-dir",
+            str(cache_dir),
+            "--platform",
+            get_installer_platform(),
+            str(tmp_dir),
+        ]
+        print_cmd(command)
```
nkaretnikov, line 125: This now uses the platform argument.

conda-store-server/conda_store_server/server/views/ui.py
```diff
@@ -2,6 +2,9 @@
 
 import yaml
 from conda_store_server import api
+from conda_store_server.action.generate_constructor_installer import (
+    get_installer_platform,
+)
```
nkaretnikov, line 7: The UI now prints the architecture/platform that's actually built by constructor.

nkaretnikov (commented):
@jaimergp Addressed your comments, added notes to parts of the code that are new. Please take a look.

Quick summary of changes:
- I do not download constructor every time the action is run, I check whether it's available and exit if not
- the installer architecture is now printed in the UI
- added a DB migration
- add a note to the docs for existing deployments
- added new screenshots in a comment.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
jaimergp: If this helper already exists somewhere else, is it worth refactoring it to a utils module? I'm not a big fan of the `print_cmd` name because it doesn't imply execution. Maybe something `logged_subprocess` or something. I'd also forward **kwargs to the subprocess call in case we want to add options (e.g. timeouts; see below).

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,127 @@
...
+    except FileNotFoundError:
+        warnings.warn(
+            "Installer generation requires constructor: https://github.com/conda/constructor"
+        )
+        return
+
+    # pip dependencies are not directly supported by constructor, they will be
+    # installed via the post_install script:
```
jaimergp, line 55: Hopefully something we can mitigate with conda-pip soon.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
jaimergp: Alternatively, raw string:

```python
post_install = r"""\
call "%PREFIX%\Scripts\activate.bat"
"""
```

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -0,0 +1,127 @@
...
+from conda_store_server import action, schema
+
+
+def get_installer_platform():
+    # This is how the default platform name is generated internally by
+    # constructor. For example: osx-arm64, linux-64, win-64.
+    # https://github.com/conda/constructor/blob/main/CONSTRUCT.md#available-platforms
+    from conda.base.context import context
+
+    return context.subdir
```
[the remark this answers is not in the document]
jaimergp: Note the value of `context.subdir` can be overridden by the `CONDA_SUBDIR` env var. `context.subdir` will assume the OS and architecture of whatever Python `conda` is running on by default, but it can also be overridden by its `--platform` flag.

I assume this function is running on the same server so we are not offering any kind of cross-building or cross-solving yet. However, we should add a note that this is cross-friendly because at some point we will want to tackle that feature, I assume.

xref https://github.com/conda-incubator/conda-store/issues/71#issuecomment-1790332315

jaimergp (approved):
Added non-blocking comments about style. The approach seems reasonable. Only concerned, long-term, with cross-platform/arch builders.

If you are unsure about the database bits, then someone else should chime in because I'm absolutely out of my depth there 😬

PR #721, opened 2024-01-05
Pass `CONDA_OVERRIDE_CUDA` to `with_cuda` of conda-lock

### LINKED ISSUE #719: [BUG] - Conda-Store strips environment vars from env.yaml spec.
### Describe the bug

The conda env.yaml spec allows setting environment variables :

https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#setting-environment-variables

We need this to enable GPU builds when using conda-store. It looks like conda-store is stripping the env variable section from the yaml. 

### Expected behavior

The environment variables are maintained in the yaml and used in building the environment.

Use the following env:

```yaml
channels:
  - conda-forge
dependencies:
  - pytorch
  - ipykernel
variables:
   CONDA_OVERRIDE_CUDA: 11.8
 ```

build the environment and then open the yaml again and the variables section is missing.

if you look at the generated lock file you can also see that the env variables were not used:

```json
{
version: 1,
metadata: {
content_hash: {
linux-64: "2a255c84a3c70a2fec49eb99f5f45a464a73d812efc7902ae201538876fcb49a"
},
channels: [
{
url: "conda-forge",
used_env_vars: [ ]
}
],
platforms: [
"linux-64"
],
sources: [
"environment.yaml"
]
},
```

### Output

_No response_

_No response_

### Anything else?

_No response_

nkaretnikov:
Fixes #719.

## Description

This pull request makes it possible to set the CUDA version by passing the value of the `CONDA_OVERRIDE_CUDA` specification variable to the `with_cuda` parameter of conda-lock.

See https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages

How I tested this:

- Created this test env _via the server admin environment page_ because _conda-store UI strips variables even in raw YAML mode_, so it doesn't work there:

```yaml
channels:
- conda-forge
dependencies:
- pytorch
- ipykernel
- pip:
  - nothing
description: test2
name: test3
prefix: null
variables:
  CONDA_OVERRIDE_CUDA: '12.0'
```

- Checked that the pytorch package in the generated lockfile has constraints matching this version of cuda. The url of the pytorch package should also reflect that, e.g., `https://conda.anaconda.org/conda-forge/linux-64/pytorch-2.1.0-cuda120py312hfe5e8c6_301.conda`.

- Additionally, on a machine with an NVIDIA card that's configured to use CUDA:

```
% CONDA_STORE_USERNAME=test CONDA_STORE_PASSWORD=password  python -m conda_store --conda-store-url="http://localhost:8080/conda-store/" --auth=basic run test/test3:4 -- python
>>> import torch
>>> torch.__file__
'/tmp/conda-store/4/lib/python3.12/site-packages/torch/__init__.py'
>>> print(torch.version.cuda)
12.0
```

This version matches the one in the lockfile, which comes from the variable. Note: I had to set the url when calling conda-store CLI in order to match the url used by the server when running via docker.

- Repeated the same with versions `10.0` and `11.0`. Observed that these affect constraints in the pytorch package in the lockfile. Note that I've used different major versions as minor versions don't affect the constraints in the lockfile.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/action/generate_lockfile.py
+++ b/conda-store-server/conda_store_server/action/generate_lockfile.py
@@ -37,6 +37,16 @@ def action_solve_lockfile(
     print_cmd(["conda", "config", "--show"])
     print_cmd(["conda", "config", "--show-sources"])
 
+    # conda-lock ignores variables defined in the specification, so this code
+    # gets the value of CONDA_OVERRIDE_CUDA and passes it to conda-lock via
+    # the with_cuda parameter, see:
+    # https://github.com/conda-incubator/conda-store/issues/719
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages
+    if specification.variables is not None:
+        cuda_version = specification.variables.get("CONDA_OVERRIDE_CUDA")
+    else:
+        cuda_version = None
+
     # CONDA_FLAGS is used by conda-lock in conda_solver.solve_specs_for_arch
     try:
         conda_flags_name = "CONDA_FLAGS"
@@ -48,6 +58,7 @@ def action_solve_lockfile(
             platforms=platforms,
             lockfile_path=lockfile_filename,
             conda_exe=conda_command,
+            with_cuda=cuda_version,
         )
     finally:
         os.environ.pop(conda_flags_name, None)
--- a/conda-store-server/conda_store_server/action/generate_lockfile.py
+++ b/conda-store-server/conda_store_server/action/generate_lockfile.py
@@ -42,6 +42,8 @@ def action_solve_lockfile(
     # the with_cuda parameter, see:
     # https://github.com/conda-incubator/conda-store/issues/719
     # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages
+    # TODO: Support all variables once upstream fixes are made to conda-lock,
+    # see the discussion in issue 719.
     if specification.variables is not None:
         cuda_version = specification.variables.get("CONDA_OVERRIDE_CUDA")
     else:


nkaretnikov:
@amjames Do you have cycles to review this one?

conda-store-server/conda_store_server/action/generate_lockfile.py
```diff
@@ -37,6 +37,16 @@ def print_cmd(cmd):
     print_cmd(["conda", "config", "--show"])
     print_cmd(["conda", "config", "--show-sources"])
 
+    # conda-lock ignores variables defined in the specification, so this code
+    # gets the value of CONDA_OVERRIDE_CUDA and passes it to conda-lock via
+    # the with_cuda parameter, see:
+    # https://github.com/conda-incubator/conda-store/issues/719
+    # https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html#overriding-detected-packages
+    if specification.variables is not None:
+        cuda_version = specification.variables.get("CONDA_OVERRIDE_CUDA")
```
amjames, line 46: The `CONDA_OVERRIDE_CUDA` variable could come in as a string (expected by conda-lock) or a float. 

The example from the description of this PR
```yaml
...
variables:
  CONDA_OVERRIDE_CUDA: '12.0'
```

Will come through as a string and be forwarded to conda-lock, but the example in the original issue has:

```yaml
...
variables:
  CONDA_OVERRIDE_CUDA: 11.8
```

Which would be parsed as float and be forwarded to conda-lock as a float. 

I don't see a test update associated with this, so I am assuming we don't have any coverage here, could you try this out locally to see if we need to convert the value as we pull it off the specification?

amjames (commented):
One question otherwise looks good for a temp fix.

conda-store-server/conda_store_server/action/generate_lockfile.py
nkaretnikov: Added a print to `action_solve_lockfile` to verify. It's always a string, even if you don't quote it. Tried with 12, 12.0, and 12.2.

(In fact, the admin UI renders previously unquoted versions in quotes once you submit an env.)

So no action is needed here.

----

You also asked offline whether using `CONDA_OVERRIDE_CUDA` on a machine without a GPU allows for using PyTorch with CUDA on a GPU-enabled machine later. I've tried the following:

- On a machine with an Intel GPU (so no CUDA):
  - added the env from the top comment to conda-store, which has `CONDA_OVERRIDE_CUDA: '12.0'`
  - waited for it to generate the lockfile (but didn't wait for it to build)
  - downloaded the lockfile
- On a machine with an NVIDIA GPU (and CUDA configured):
  - created an env from the downloaded lockfile via `conda-lock install -n mytest1 --micromamba ~/mytest1.json`
  - activated the env
  - confirmed that CUDA works:
  
```
>>> import torch
>>> torch.__file__
'/home/nkaretnikov/.conda/envs/mytest1/lib/python3.12/site-packages/torch/__init__.py'
>>> print(torch.version.cuda)
12.0
>>> x = torch.tensor(1, device='cuda')
>>> x.device
device(type='cuda', index=0)
>>> x
tensor(1, device='cuda:0')
```

Please let me know if you have any other questions!

amjames (approved):

nkaretnikov:
@trallard Could you approve so I could merge? Andrew has reviewed this.

conda-store-server/conda_store_server/action/generate_lockfile.py
```diff
@@ -37,6 +37,16 @@ def print_cmd(cmd):
     print_cmd(["conda", "config", "--show"])
     print_cmd(["conda", "config", "--show-sources"])
 
+    # conda-lock ignores variables defined in the specification, so this code
+    # gets the value of CONDA_OVERRIDE_CUDA and passes it to conda-lock via
+    # the with_cuda parameter, see:
```
trallard, line 42: Nit, so non blocking but since we said this is an "interim" solution it would be best to add also a TODO/something indicating that we should look at a more robust approach

trallard (approved):

trallard:
Added a non-blocking comment otherwise we can merge

PR #736, opened 2024-01-19
Fix issue when pip dependencies from poetry throw an error

### LINKED ISSUE #735: [BUG] - conda-store doesn't report error in logs when conda-lock poetry has a failure to solve
### Describe the bug

Prior to this PR create an environment that can't solve e.g. and the solve with fail and conda-store will not report the build as failed. This happens when you specify a pip package that doesn't exist.

```yaml
name: test
channels:
 - conda-forge
dependencies:
 - python
 - pip
 - pip:
    - numpyyyyy
``` 

### Expected behavior

Solve should succeed

```yaml
name: test
channels:
 - conda-forge
dependencies:
 - python
 - pip
 - pip:
    - numpyyyyy
``` 

### Output

_No response_

_No response_

### Anything else?

_No response_

costrouc:
Fixes #735 

## Description

Prior to this PR create an environment that can't solve e.g.

```yaml
name: test
channels:
 - conda-forge
dependencies:
 - python
 - pip
 - pip:
    - numpyyyyy
``` 

## Pull request checklist

- [X] Did you update the documentation (if required)? (no update needed)

### DIFF
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -9,9 +9,9 @@ import traceback
 import typing
 
 import yaml
+from conda_lock._vendor.poetry.puzzle.exceptions import SolverProblemError
 from conda_store_server import action, api, conda_utils, orm, schema, utils
 from conda_store_server.utils import BuildPathError
-from conda_lock._vendor.poetry.puzzle.exceptions import SolverProblemError
 from sqlalchemy.orm import Session
 
 
--- a/conda-store-server/conda_store_server/build.py
+++ b/conda-store-server/conda_store_server/build.py
@@ -11,6 +11,7 @@ import typing
 import yaml
 from conda_store_server import action, api, conda_utils, orm, schema, utils
 from conda_store_server.utils import BuildPathError
+from conda_lock._vendor.poetry.puzzle.exceptions import SolverProblemError
 from sqlalchemy.orm import Session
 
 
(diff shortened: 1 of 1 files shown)


costrouc:
@nkaretnikov I do think that we need to think through how to better catch these exceptions. Maybe just have a single catch-all and return a traceback in the logs. It feels like this could become "wack-a-mole" in the future if things change in the libraries that we depend on.

nkaretnikov (changes requested):
@costrouc I might be missing something, but it looks like this PR doesn't add anything new. This issue is already solved on main, so no action necessary.

I've tested on `f63b4d7bb42c9bc18dfaca1e3308f294c9b9c06c`.

The traceback is printed to the console and in the web UI. The build is marked as "Failed".

In the same function below, there's already a block that handles `Exception`, which `SolverProblemError` inherits from: https://github.com/python-poetry/poetry/blob/605e6c9fc1cf1f8ff6d855cdfda86197d47bfee1/src/poetry/puzzle/exceptions.py#L13

```py
class SolverProblemError(Exception):
...
```

```py
    except Exception as e:
        set_build_failed(db, build)
        conda_store.log.exception(e)
        append_to_logs(db, conda_store, build, traceback.format_exc())
        raise e
```

The code in this block and the one added in this PR is exactly the same.

Screenshots:

Logs in the UI:

Screen Shot 2024-01-19 at 14 47 01

Build marked as FAILED:

Screen Shot 2024-01-19 at 14 46 48

PR #738, opened 2024-01-21
Add support for the `editor` role

### LINKED ISSUE #675: [ENH] - Support "editor" as an alias for "developer"
### Feature description

- Add another role to default role_mappings called "editor" with the same permissions as "developer"
- Update the DB, docs, and tests to support both of these aliases

### Value and/or benefit

This is considered a better name because the set of permissions is about editing. This has nothing to do with development.

### Anything else?

_No response_

nkaretnikov:
Fixes #675.

## Description

This pull request:

- adds the alias `editor` for `developer` to role mappings, which is supported by the HTTP API and the DB
- adds tests
- updates documentation.

The auth flow is as follows:
- an HTTP API is called to set the role (can be either `editor` or `developer`)
- the role is written to the DB (always written as `developer`)
- when some page is accessed, the role is checked on auth by matching the DB value against `role_mappings`
  - `developer` from the DB is matched against `editor` or `developer` in `role_mappings` (in this order).

All user-visible parameters (the HTTP API and the `role_mappings` setting) allow for both `editor` and `developer`.
The name `editor` is preferred, which is reflected in the documentation.
The name `developer` is supported for compatibility reasons.

How I tested this locally:
- [x] to make sure that `role_mappings` are actually used by the test, I commented out `editor` in the default `role_mappings`
  - the test errored out as expected
- [x] to make sure that `developer` is still supported, I replaced `editor` with `developer` in the default `role_mappings`
  - the test succeeded as expected
- [x] to make sure that permissions are checked, I replaced the permissions of `editor` in the default `role_mappings`
  - the test errored out as expected

After making each of these changes in the application, I ran

```sh
PYTEST_ADDOPTS="-rpfxs -k test_end_to_end_auth_flow" pytest tests
```

and looked at the test output.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/server/auth.py
+++ b/conda-store-server/conda_store_server/server/auth.py
@@ -138,42 +138,47 @@ class RBACAuthorizationBackend(LoggingConfigurable):
             )
         return proposal.value
 
+    _viewer_permissions = {
+        schema.Permissions.ENVIRONMENT_READ,
+        schema.Permissions.NAMESPACE_READ,
+        schema.Permissions.NAMESPACE_ROLE_MAPPING_READ,
+    }
+    _developer_permissions = {
+        schema.Permissions.BUILD_CANCEL,
+        schema.Permissions.ENVIRONMENT_CREATE,
+        schema.Permissions.ENVIRONMENT_READ,
+        schema.Permissions.ENVIRONMENT_UPDATE,
+        schema.Permissions.ENVIRONMENT_SOLVE,
+        schema.Permissions.NAMESPACE_READ,
+        schema.Permissions.NAMESPACE_ROLE_MAPPING_READ,
+        schema.Permissions.SETTING_READ,
+    }
+    _admin_permissions = {
+        schema.Permissions.BUILD_DELETE,
+        schema.Permissions.BUILD_CANCEL,
+        schema.Permissions.ENVIRONMENT_CREATE,
+        schema.Permissions.ENVIRONMENT_DELETE,
+        schema.Permissions.ENVIRONMENT_READ,
+        schema.Permissions.ENVIRONMENT_UPDATE,
+        schema.Permissions.ENVIRONMENT_SOLVE,
+        schema.Permissions.NAMESPACE_CREATE,
+        schema.Permissions.NAMESPACE_DELETE,
+        schema.Permissions.NAMESPACE_READ,
+        schema.Permissions.NAMESPACE_UPDATE,
+        schema.Permissions.NAMESPACE_ROLE_MAPPING_CREATE,
+        schema.Permissions.NAMESPACE_ROLE_MAPPING_READ,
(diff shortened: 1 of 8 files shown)


docusaurus-docs/conda-store/references/auth.md
```diff
@@ -101,7 +101,9 @@ c.RBACAuthorizationBackend.role_mappings = {
 }
 ```
 
-Lets go through a few examples to make this more concrete and assume
+Additionally, the role `editor` can be used, which is an alias for `developer`.
+
+Let's go through a few examples to make this more concrete and assume
```
nkaretnikov, line 106: Typo fix: "Lets" -> "Let's"

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -271,7 +275,23 @@ def get_entity_bindings(self, entity: schema.AuthenticationToken):
...
         for role in roles:
-            permissions = permissions | self.role_mappings[role]
+            # 'editor' is the new alias of 'developer'. The new name is
+            # preferred in user-visible settings (like 'role_mappings') and when
+            # calling the role-mappings HTTP APIs, but it's ALWAYS mapped to
+            # 'developer' in the database for compatibility reasons.
+            # Additionally, this code allows for legacy 'role_mappings' that
+            # used to specify the role as 'developer'. Because it's a
+            # user-visible setting, we cannot break compatibility here
+            assert role != "editor"  # must NEVER be 'editor' in the DB
```
dcmcand, line 285: An assert outside of a test strikes me as odd. Looking into it, it seems like some compilation flags remove assert statements. If this check is valuable, I think raising an exception here would be more valuable. However, I am not sure that this represents an unrecoverable error, since we would still know what to do with that data. We could just handle it and log an warning that data that shouldn't be in the db is. Or if it really does represent an unrecoverable error, than raising an exception seems better than an assert.

dcmcand (changes requested):
Just the one area that I think needs changed. Other than that, looks good.

conda-store-server/conda_store_server/server/auth.py
nkaretnikov: @dcmcand Changed to `raise`. It's not unrecoverable, but it's a mistake that's easy to make when potentially refactoring or adding new DB APIs.

For example, see the change in `conda-store-server/conda_store_server/api.py` in this PR. The old code would bypass `validate_role` because `.update` is a lowel-level API, which just skips ORM validation logic. That was caught by the tests, which is how I noticed. But we might not have good coverage for some new yet to be added APIs, so I want to make sure this is never a problem.

nkaretnikov:
@dcmcand PTAL. Note: I want to keep the diff minimal to make it easier to review, I'll rebase after this is approved.

dcmcand (approved):
:rocket:

PR #745, opened 2024-01-24
Add a test for parallel builds, fix race conditions due to the shared conda cache

### LINKED ISSUE #548: [ENH] - Test for parallel builds

nkaretnikov:
Fixes #548.

## Description

This pull request:
- Tests whether parallel builds are working by spawning N jobs and checking whether these complete faster as they would do sequentially.
- **Fixes race conditions causing dependencies being overwritten on concurrent builds**. This is due to the package cache not being atomic, which is a conda upstream issue. This issue got exposed thanks to the new test added in this PR. Here are the relevant issues from the upstream repo, but they were closed without any solution or not yet addressed:
  - https://github.com/conda/conda/issues/9447
  - https://github.com/conda/conda/issues/5997
  - https://github.com/conda/conda/issues/7741
- I had to increase timeouts of some unrelated tests because the parallel builds test creates multiple jobs, so it might now take longer for the job scheduler to go through those before jobs matching the thing being tested actually start executing.

More about the mentioned race conditions:
1. The main issue, which was causing ~50% of jobs to fail when running the new test, is because conda created `repodata_record.json` with the `channel` field set to `"<unknown>"`. This was mitigated by creating the said file in conda-store after extracting the package tarball, which causes conda to skip this step. The jobs were failing because `channel` is one of the fields used when matching packages in the cache against the provided specification. In the specification, `channel` was never `"<unknown>"`, so conda failed to find the proper package in the cache and attempted to download it without any locking mechanism (because the cache in conda is not atomic). If multiple builds had the same dependencies, they were racing each other while trying to download "missing" packages into the cache, so the cache got corrupted because the builds were overwriting the same dependencies, which caused build failures. You can verify that the cache has no packages with an invalid channel by using the following command: `find -L /opt/conda/pkgs -name repodata_record.json -exec grep -rnIH --color unknown {} \;`.
2. However, that wasn't enough to completely mitigate the problem. conda-store has locking, but it's only done during the `action_fetch_and_extract_conda_packages` step, per package. In other steps during the build process, such as installing the lockfile, we call into conda, which knows nothing about this lock and can cause issues. Even with the  `repodata_record.json` fix described in (1), there's still a potential race condition where conda might check the extracted package directory before the json file was created there. Or it might check whether the tarball is extracted before conda-store had a chance to extract it and start extracting it on its own. To mitigate this, I download the tarball, extract it, and create the json file in a temporary directory, which I then move into the shared package cache. There's still a tiny race window during the `rmtree` and `move` operations because it cannot be done with one atomic syscall. However, that's the best we can do for now. The said race window is tiny, so unlikely to cause serious issues in practice. I also had to be careful with how I use the `shutil` operations so that they wouldn't cause failures if a job fails and stale data is present in the cache. This was discovered on the Windows CI builder because it has the extended prefix test that's expected to fail, which was causing subsequent jobs to fail as well. This was reproducing all the time (or maybe most of the time) on the Windows builder.
3. Similarly, there was a race condition in the `symlink` helper, which conda-store calls at the end of the build process. Multiple build jobs were racing each other while trying to remove the existing target file and create the symlink, which caused build failures. Because it's only done by conda-store, this is easy to fix by making this whole process atomic by using a filelock. The filelock package is a spinlock, so builds will continue trying to create the symlink until they eventually succeed. This symlink issue was maybe reproducing once for every 10 jobs.
4. I also had to create the magic file in the cache because sometimes it's not yet created when the download action is running. Without the magic file, cache queries would fail even if `repodata_record.json` files have proper channels specified.

## Pull request checklist

## How to test

- To test parallel builds:
  - set `c.CondaStoreWorker.concurrency = 1` in `tests/assets/conda_store_config.py`
  - run `docker compose up --build`
  - once the server is running and accepting connections, run `PYTEST_ADDOPTS="-vvv -k test_create_specification_parallel_auth --capture=no" pytest tests/test_api.py` in a different terminal
  - the test should fail with an assert because builds take too long
  - set the `concurrency` value to `4` (the default), the test should succeed.

- To test the caching issue:
  - switch to the `main` branch
  - run the test from this PR with `c.CondaStoreWorker.concurrency = 4`
  - at least one of the builds should fail (status 'FAILED')
  - when I was testing this, I saw something like 50% of the builds being in the 'FAILED' state in the admin web UI
  - in the build logs, the errors should look similar to this:
```
ERROR:root:/opt/conda/pkgs/psutil-5.9.0-py310h5eee18b_0.tar.bz2
ERROR:root:Error with archive /opt/conda/pkgs/traitlets-5.7.1-py310h06a4308_0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message was:
ERROR:root:
ERROR:root:failed with error: [Errno 2] No such file or directory: '/opt/conda/pkgs/traitlets-5.7.1-py310h06a4308_0.tar.bz2'
ERROR:root:/opt/conda/pkgs/wheel-0.41.2-py310h06a4308_0.tar.bz2
ERROR:root:Error with archive /opt/conda/pkgs/comm-0.1.2-py310h06a4308_0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message was:
ERROR:root:
ERROR:root:failed with error: [Errno 2] No such file or directory: '/opt/conda/pkgs/comm-0.1.2-py310h06a4308_0.tar.bz2'
ERROR:root:/opt/conda/pkgs/jupyter_core-5.5.0-py310h06a4308_0.tar.bz2
ERROR:root:Compressed file ended before the end-of-stream marker was reached
ERROR:root:
ERROR:root:
ERROR:root:PathNotFoundError: /opt/conda/pkgs/psutil-5.9.0-py310h5eee18b_0.tar.bz2
ERROR:root:InvalidArchiveError("Error with archive /opt/conda/pkgs/traitlets-5.7.1-py310h06a4308_0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message was:\n\nfailed with error: [Errno 2] No such file or directory: '/opt/conda/pkgs/traitlets-5.7.1-py310h06a4308_0.tar.bz2'")
ERROR:root:PathNotFoundError: /opt/conda/pkgs/wheel-0.41.2-py310h06a4308_0.tar.bz2
ERROR:root:InvalidArchiveError("Error with archive /opt/conda/pkgs/comm-0.1.2-py310h06a4308_0.tar.bz2.  You probably need to delete and re-download or re-create this file.  Message was:\n\nfailed with error: [Errno 2] No such file or directory: '/opt/conda/pkgs/comm-0.1.2-py310h06a4308_0.tar.bz2'")
ERROR:root:PathNotFoundError: /opt/conda/pkgs/jupyter_core-5.5.0-py310h06a4308_0.tar.bz2
ERROR:root:EOFError('Compressed file ended before the end-of-stream marker was reached')
```

### DIFF
--- a/conda-store-server/conda_store_server/action/download_packages.py
+++ b/conda-store-server/conda_store_server/action/download_packages.py
@@ -44,14 +44,18 @@ def action_fetch_and_extract_conda_packages(
             with filelock.FileLock(str(lock_filename)):
                 # This magic file, which is currently set to "urls.txt", is used
                 # to check cache permissions in conda, see _check_writable in
-                # PackageCacheData. Sometimes this file is not yet created while
-                # this action is running. But without this magic file,
-                # PackageCacheData cache query functions, like query_all, will
-                # return nothing, so packages will be downloaded outside of this
-                # action, which will corrupt the cache. Without the magic file,
-                # this error might be thrown when installing the lockfile:
+                # PackageCacheData.
+                #
+                # Sometimes this file is not yet created while this action is
+                # running. Without this magic file, PackageCacheData cache
+                # query functions, like query_all, will return nothing.
+                #
+                # If the magic file is not present, this error might be thrown
+                # during the lockfile install action:
+                #
                 # File "/opt/conda/lib/python3.10/site-packages/conda/misc.py", line 110, in explicit
                 #   raise AssertionError("No package cache records found")
+                #
                 # The code below is from create_package_cache_directory in
                 # conda, which creates the package cache, but we only need the
                 # magic file part here:
@@ -72,6 +76,70 @@ def action_fetch_and_extract_conda_packages(
                     with file_path.open("wb") as f:
                         shutil.copyfileobj(conda_package_stream, f)
                     conda_package_handling.api.extract(file_path_str)
+
+                    # This code is needed to avoid failures when building in
+                    # parallel while using the shared cache.
+                    #
+                    # There are tarballs that only have the info/index.json file
+                    # and no info/repodata_record.json. The latter is used to
+                    # interact with the cache, so _make_single_record from
+                    # PackageCacheData would create the missing json file during
+                    # the lockfile install action.
+                    #
+                    # The code that does that in conda is similar to the code
+                    # below. However, there is an important difference. The
+                    # code in conda would fail to read the url and return None
+                    # here:
+                    #
+                    #   url = self._urls_data.get_url(package_filename)
+                    #
+                    # And that would result in the channel field of the json
+                    # file being set to "<unknown>". This is a problem because
+                    # the channel is used when querying cache entries, via
+                    # match_individual from MatchSpec, which would always result
+                    # in a mismatch because the proper channel value is
+                    # different.
+                    #
+                    # That would make conda think that the package is not
+                    # available in the cache, so it would try to download it
+                    # outside of this action, where no locking is implemented.
+                    #
+                    # As of now, conda's cache is not atomic, so the same
+                    # dependencies requested by different builds would overwrite
+                    # each other causing random failures during the build
+                    # process.
+                    #
+                    # To avoid this problem, the code below does what the code
+                    # in conda does but also sets the url properly, which would
+                    # make the channel match properly during the query process
+                    # later. So no dependencies would be downloaded outside of
+                    # this action and cache corruption is prevented.
+                    #
+                    # To illustrate, here's a diff of an old conda entry, which
+                    # didn't work, versus the new one created by this action:
+                    #
+                    # --- /tmp/old.txt        2024-02-05 01:08:16.879751010 +0100
+                    # +++ /tmp/new.txt        2024-02-05 01:08:02.919319887 +0100
+                    # @@ -2,7 +2,7 @@
+                    #    "arch": "x86_64",
+                    #    "build": "conda_forge",
+                    #    "build_number": 0,
+                    # -  "channel": "<unknown>",
+                    # +  "channel": "https://conda.anaconda.org/conda-forge/linux-64",
+                    #    "constrains": [],
+                    #    "depends": [],
+                    #    "features": "",
+                    # @@ -15,5 +15,6 @@
+                    #    "subdir": "linux-64",
+                    #    "timestamp": 1578324546067,
+                    #    "track_features": "",
+                    # +  "url": "https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2",
+                    #    "version": "0.1"
+                    #  }
+                    #
+                    # Also see the comment above about the cache magic file.
+                    # Without the magic file, cache queries would fail even if
+                    # repodata_record.json files have proper channels specified.
 
                     # Otherwise .conda, do nothing in that case
                     ext = ".tar.bz2"
--- a/conda-store-server/conda_store_server/action/download_packages.py
+++ b/conda-store-server/conda_store_server/action/download_packages.py
@@ -58,8 +58,6 @@ def action_fetch_and_extract_conda_packages(
                         extracted_dir = pathlib.Path(file_path_str.replace(ext, ""))
                         # This file is used to parse cache records via PackageCacheRecord in conda
                         repodata_file = extracted_dir / "info" / "repodata_record.json"
-                        index_file = extracted_dir / "info" / "index.json"
-                        assert index_file.exists()
 
                         if not repodata_file.exists():
                             raw_json_record = read_index_json(extracted_dir)
@@ -81,4 +79,3 @@ def action_fetch_and_extract_conda_packages(
                                 package_cache_record
                             )
                             write_as_json_to_file(repodata_file, repodata_record)
-                            assert repodata_file.exists()
--- a/conda-store-server/conda_store_server/action/download_packages.py
+++ b/conda-store-server/conda_store_server/action/download_packages.py
@@ -90,11 +90,14 @@ def action_fetch_and_extract_conda_packages(
                         # This code is needed to avoid failures when building in
                         # parallel while using the shared cache.
                         #
-                        # There are tarballs that only have the info/index.json file
-                        # and no info/repodata_record.json. The latter is used to
-                        # interact with the cache, so _make_single_record from
-                        # PackageCacheData would create the missing json file during
-                        # the lockfile install action.
+                        # Package tarballs contain the info/index.json file,
+                        # which is used by conda to create the
+                        # info/repodata_record.json file. The latter is used to
+                        # interact with the cache. _make_single_record from
+                        # PackageCacheData would create the repodata json file
+                        # if it's not present, which would happen in conda-store
+                        # during the lockfile install action. The repodata file
+                        # is not created if it already exists.
                         #
                         # The code that does that in conda is similar to the code
                         # below. However, there is an important difference. The
--- a/conda-store-server/conda_store_server/action/download_packages.py
+++ b/conda-store-server/conda_store_server/action/download_packages.py
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import shutil
+import tempfile
 import typing
 
 # This import is needed to avoid the following error on conda imports:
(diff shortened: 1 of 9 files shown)


tests/test_api.py
```diff
@@ -467,6 +469,100 @@ def test_create_specification_auth(testclient):
...
+    print("loop_delta", loop_delta)
+
+    # If there are jobs in the queue, the loop didn't complete in the allocated
+    # time. So something went wrong, like a build getting stuck or parallel
+    # builds not working
+    assert len(build_ids) == 0
+
+    # If all jobs are done twice as fast as they would do sequentially, then
+    # parallel builds are working
+    assert loop_delta < build_delta * (num_builds / 2)
```
nkaretnikov, line 563: This is a flawed assumption. A different check is needed. It doesn't work because there's some additional time needed before the first build actually starts executing. Subsequent builds (from build 2+) are much faster.

With concurrency=4, I get:
```
build_delta 178.358115
loop_delta 411.383843
```

(Note: my test VM only has 2 CPUs, exposed, which explains why this is only 2x faster than with concurrency=1.)

With concurrency=1, I get:
```
build_delta 427.717276
loop_delta 784.000384
```

Since this is an integration test, I cannot just measure against different concurrency parameters in the test. I think I might try comparing the deltas between subsequent jobs (1 vs 2, 2 vs 3, etc.), with the assumption that with concurrency=1 these will be roughly the same, and with concurrency=4 these will be done in batches.

tests/test_api.py
```diff
@@ -467,6 +470,158 @@ def test_create_specification_auth(testclient):
...
+    # So the test below looks at the value of the first quartile, which is the
+    # median of the lower half of the dataset, where all these small values will
+    # be located, and compares it to a certain threshold, which is unlikely to
+    # be reached based on how long it takes a single build to run
+    # non-concurrently on average:
+    threshold = 5
+    quartiles = statistics.quantiles(build_deltas, method='inclusive')
+    print("build_deltas", build_deltas)
+    print("stats", min(build_deltas), quartiles)
+    assert quartiles[0] < threshold
```
nkaretnikov, line 622: I ran benchmarks on a completely different machine with 4 CPUs, just to see how values would change. I think the above test is okay based on these:

concurrency = 4:
```
build_deltas [1.025565, 7.176997, 1.019974, 77.987058, 4.085229, 25.557227, 14.347968, 8.206779, 2.057178]
stats 1.019974 [2.057178, 7.176997, 14.347968]

build_deltas [1.031824, 1.020348, 45.126866, 37.897655, 17.460182, 8.242122, 65.296544, 1.02701, 1.049515]
stats 1.020348 [1.031824, 8.242122, 37.897655]

build_deltas [1.029401, 50.235828, 1.067922, 31.738958, 17.412726, 41.970197, 28.651753, 1.030208, 14.338367]
stats 1.029401 [1.067922, 17.412726, 31.738958]
```
(The last one has a bigger median because the retrieve packages job was running in parallel.
I wanted to see how the numbers would be affected by that. Note that the median of the lower half was not affected.)

----

concurrency = 1:
```
build_deltas [28.631925, 33.738711, 29.625621, 56.210018, 57.21373, 33.693227, 53.14048, 33.745332, 35.830624]
stats 28.631925 [33.693227, 33.745332, 53.14048]

>       assert quartiles[0] < threshold
E       assert 33.693227 < 5

build_deltas [28.665233, 33.790378, 29.662559, 56.242397, 57.218049, 33.758011, 105.30695, 33.760865, 35.784077]
stats 28.665233 [33.758011, 33.790378, 56.242397]

>       assert quartiles[0] < threshold
E       assert 33.758011 < 5

build_deltas [37.866588, 33.755057, 29.649771, 50.167768, 67.47894, 33.751589, 56.277974, 37.855099, 36.887129]
stats 29.649771 [33.755057, 37.855099, 50.167768]

>       assert quartiles[0] < threshold
E       assert 33.755057 < 5
```

nkaretnikov: And here's a [box plot](https://en.wikipedia.org/wiki/Box_plot#Elements) of the datasets from the test comment. Note how the Q1 value (the lower part of the box that represents the median of the smallest half of the dataset) is very similar between `c4_2cpu` and `ci` (which are running with concurrent builds on). Q1 is very close to the min value and the min is itself very small. Compare that to `c1`, which is what we're trying to detect, which is running non-concurrently.

tests

```py
import matplotlib
matplotlib.use('TkAgg')
import matplotlib.pyplot as plt

my_dict = {'c4_2cpu': c4_2cpu, 'c1': c1, 'ci': ci}
plt.boxplot(my_dict.values(), labels=my_dict.keys())
plt.show()
```

nkaretnikov: And based on the other benchmark values posted above, which are from a different machine, you can see that the smallest half doesn't vary much when running concurrently, so it should be reliable as an estimate.

nkaretnikov: The reason Q1 is used (as opposed to just comparing min values) is because it makes use of the smallest half of the dataset, which is what we're interested in, and the median value helps filter out outliers there.

.github/workflows/tests.yaml
```diff
@@ -110,7 +110,7 @@ jobs:
       - name: "Run integration tests ✅"
         run: |
           export PYTHONPATH=$PYTHONPATH:$PWD
-          pytest ../tests/test_api.py ../tests/test_metrics.py
+          PYTEST_ADDOPTS="-vvv --capture=no" pytest ../tests/test_api.py ../tests/test_metrics.py
```
nkaretnikov, line 113: TODO: Remove this (debug only)

tests/test_api.py
```diff
@@ -467,6 +470,158 @@ def test_create_specification_auth(testclient):
...
+            break
+
+        # Measures how long it takes to do a single build
+        if len(build_ids) < prev_builds:
+            if prev is not None:
+                build_delta = (now - prev).total_seconds()
+                print("build_delta", build_delta)
+                build_deltas.append(build_delta)
+            prev_builds = len(build_ids)
+            prev = now
```
nkaretnikov, line 530: This skips the very first build on purpose because it usually takes much longer than the other builds, due to the jobs waiting to be scheduled

tests/test_api.py
```diff
@@ -467,6 +470,158 @@ def test_create_specification_auth(testclient):
...
+
+    # If there are jobs in the queue, the loop didn't complete in the allocated
+    # time. So something went wrong, like a build getting stuck or parallel
+    # builds not working
+    assert len(build_ids) == 0
+
+    # Because this is an integration test, we cannot change the server
+    # c.CondaStoreWorker.concurrency value, which is set to 4 by default. But
+    # it's possible to devise a statistical test based on locally collected
+    # data, using 'build_deltas' above:
```
nkaretnikov, line 566: The reason for this being an integration test is that it'd test the code the same way users would in real world scenarios. I had a version of this as a unit-test before, but it was rejected due to not being realistic: https://github.com/conda-incubator/conda-store/pull/567

tests/test_api.py
```diff
@@ -467,6 +470,158 @@ def test_create_specification_auth(testclient):
...
+    #     1.02644,
+    #     33.591736,
+    #     1.016269,
+    #     19.299738,
+    #     7.115025,
+    #     20.342442,
+    #     12.222805,
+    #     6.103751,
+    #     1.016237,
+    # ]
```
nkaretnikov, line 606: This one is from a CI runner.

tests/test_api.py
```diff
@@ -467,6 +470,158 @@ def test_create_specification_auth(testclient):
...
+    #
+    # These values will vary depending on workload and the number of CPUs. But
+    # the main observation here is this: if parallel builds are working,
+    # there will be a number of values that are relatively small compared to the
+    # time it takes to run a single build when not running concurrently.
```
nkaretnikov, line 611: This is because these 1-5 second values are actually two builds running in parallel and completing almost at the same time after each other:

```
| build1 start -------------------------------- build1 done | <-> 1-5 second delta here
  | build2 start ----------------------------------- build2 done |
time ------------------------------------------------------------------------->
```

conda-store-server/conda_store_server/action/install_lockfile.py
```diff
@@ -16,6 +18,32 @@ def action_install_lockfile(
...
+    #
+    # For example, packages A and B use the same dependency. Package A starts
+    # the download process and puts the dependency in the cache. Package B sees
+    # that the dependency is in the cache and tries to use it, but the
+    # dependency can be in an invalid state, such as being not fully-copied or
+    # downloaded, because the cache operations are not atomic, so the build of
+    # package B fails.
+    #
+    # This is a known problem in conda and I'm not aware of any solution besides
+    # just disabling the cache by having a per-build temporary cache directory.
```
pearu, line 33: Another solution is to download the dependency to a temporary file (so that has different name from the dependency file) and when download is complete, rename the temporary file to the correct file name. In this way, the package B can see only the valid dependencies in the cache.

nkaretnikov: @pearu I don't think it's practical to do here. conda-store is a thin wrapper around conda that provides a user-friendly web UI for package management. This would essentially require me to reimplement conda's caching. So it seems to me it's something that should be solved upstream, probably using the symlink mechanism, like in Nix, because it's an atomic operation on Linux. But there's also Windows to think about as well, which might require a different solution.

pearu: @nkaretnikov yes, fixing this upstream makes sense.

nkaretnikov:
Chris mentioned to me offline that there's some locking code in `action_fetch_and_extract_conda_packages`, which runs before this. I'll investigate why that doesn't work and report back.

nkaretnikov:
I think we should merge this as is. There's no regression when using temporary per-build caches, which this PR introduces.
There are two problems with existing code in `main`:
- conda-lock always downloads and extracts dependencies regardless of whether they are in the cache or not, so caching was never working, and there's no way to control that
- this download and extraction process is not atomic in conda, so multiple builds write over each other, which corrupts packages in the cache causing build failures.

Here is an example chain of calls:
- `action_install_lockfile` (conda-store)
- `python -m conda_lock install ...` (calls into conda-lock)
- `click_install`
- `install`
- `do_conda_install`
- `mamba create ...` (calls into mamba)
- `/opt/conda/bin/mamba`
- `main` (from `mamba.mamba`)
- `_wrapped_main`
- `do_call`
- `create`
- `install`
- `explicit(args_packages, prefix, ...)` (from `conda.misc`)
- `pfe = ProgressiveFetchExtract(fetch_specs)` (from `.core.package_cache_data`)
- `pfe.execute()`
- where `fetch_specs` looks like this:
```
[MatchSpec("main/linux-64::_libgcc_mutex==0.1=main[md5=013d3f22cd3b87f71224bd936765bcad]"),
MatchSpec("main/linux-64::ca-certificates==2023.12.12=h06a4308_0[md5=78b3ba713700dac5e7ea2617e6d165d2]"),
MatchSpec("main/linux-64::ld_impl_linux-64==2.38=h1181459_1[md5=9508af80612e4fa1b5d1abb43ca7e63c]"),
MatchSpec("main/noarch::tzdata==2023d=h04d1e81_0[md5=f5eda80e3454526141a2246c21b060ee]"),
...
]
```
- `self.prepare()` (in `ProgressiveFetchExtract`)
- `self.make_actions_for_record` (based on the provided `fetch_specs`)
- `self.cache_actions` and `self.extract_actions` are populated and look like this:
```
CacheUrlAction<url='https://conda.anaconda.org/main/linux-64/python-3.11.7-h955ad1f_0.tar.bz2', target_full_path='/opt/conda/pkgs/python-3.11.7-h955ad1f_0.tar.bz2'>
CacheUrlAction<url='https://conda.anaconda.org/main/linux-64/attrs-23.1.0-py311h06a4308_0.tar.bz2', target_full_path='/opt/conda/pkgs/attrs-23.1.0-py311h06a4308_0.tar.bz2'>
...

ExtractPackageAction<source_full_path='/opt/conda/pkgs/python-3.11.7-h955ad1f_0.tar.bz2', target_full_path='/opt/conda/pkgs/python-3.11.7-h955ad1f_0'>
ExtractPackageAction<source_full_path='/opt/conda/pkgs/attrs-23.1.0-py311h06a4308_0.tar.bz2', target_full_path='/opt/conda/pkgs/attrs-23.1.0-py311h06a4308_0'>
```
- there are the same packages which were downloaded to cache by `action_fetch_and_extract_conda_packages` before
- `do_cache_action` and `do_extract_action` are called
- `cache_action.execute` and `extract_action.execute` are called
- `CacheUrlAction` and `ExtractPackageAction` are called (see [`conda/conda/core
/path_actions.py`](https://github.com/conda/conda/blob/cd1511dac91e16b56b31d731393baddda13365c9/conda/core/path_actions.py)
- `CacheUrlAction.execute`:
  - `target_package_cache = PackageCacheData(self.target_pkgs_dir)`
  - `self._execute_channel(target_package_cache, progress_update_callback)`
    - `download` (from `..gateways.connection.download`)
- `ExtractPackageAction.execute`:
  - `extract_tarball` (from `..gateways.disk.create`).

Looking at this chain of calls, I see no way to skip the download/extraction process.

nkaretnikov:
@jaimergp Tania asked that you review this one. Please take a look, thanks!

conda-store-server/conda_store_server/action/install_lockfile.py
```diff
@@ -16,6 +18,32 @@ def action_install_lockfile(
...
+    # requires having access to dependencies from the cache.
+    #
+    # Here are the relevant issues from the upstream repo, but they were closed
+    # without any solution:
+    # https://github.com/conda/conda/issues/9447
+    # https://github.com/conda/conda/issues/5997
+    # https://github.com/conda/conda/issues/7741
+    tmp_pkgs_dir = tempfile.TemporaryDirectory()
+    env = os.environ.copy()
+    env["CONDA_PKGS_DIRS"] = tmp_pkgs_dir.name
```
jaimergp, line 45: This is problematic. Having ephemeral caches might have unexpected outcomes since conda's central model is "download to a central cache and link packages from there". While we are okay with hardlinks (the underlying disk blocks just get a new pointer/inode), hard-copies with path replacement will also lose the original file.

You might also lose important metadata only available in the `info/` folder of the cache, and this is not necessarily linked to `conda-meta`. 

And of course, this disables the cache completely, which will incur in repeated package downloads, increased bandwidth usage, and overall waste of disk writes.

jaimergp (changes requested):
I don't think we should move forward with this fix. It's going to cause a huge performance impact on IO (downloads, disk writes).

Instead, I'd like to revisit the current workflow where we solve an environment. I see three steps:

1. Solve the environment (need to download repodata)
2. Update the cache as needed by downloading and extracting artifacts.
3. Linking artifacts to target environment.

For step 1, I'd like to have a service that periodically downloads and updates the repodata cache (let's say every 15 min). Then all the solves happen in `--offline --dry-run --json` mode (or whatever conda-lock requires). This should minimize locking problems there.

For step 2, I would set a download queue service. This service would have a queue based on conda's `ProgressiveFetchExtract`, and would be able to calculate which things need to be downloaded and which things are already in cache. If two jobs try to download the same thing, the queue should be smart enough to know whether they are already downloaded or in-queue.

Finally, for step 3, we would check if linking can happen based on the cache status of the required packages. Once all the items added to the download queue are downloaded and extracted, linking can start.

This might require a major overhaul of the build jobs, because they will become three separate steps that run in some kind of async fashion, but honestly, I don't see a better solution.

We could instead contribute per-artifact-locking for `conda` (this is somehow in their roadmap for 2024), but then we are not in control of the download queue and other optimizations we could from a "server mode" setup as we are trying to do here.

With respect to this PR, I'd leave the test with XFAIL, and then iterate as necessary in subsequent PRs. The temporary `CONDA_PKGS_DIRS` fix cannot make it into the codebase, imho.

Let me know if you have questions, but this is probably better discussed in the issue we are trying to close.

nkaretnikov:
Jaime and I talked about this offline:
- I need to look into why caching is not happening in `ProgressiveFetchExtract`, this shouldn't take more than 3 hours
	- there's some code that prepares the actions before `execute` is called, in `prepare`
	- `prepare` calls `make_actions_for_record` and  that's where the cache could have been used, but it doesn't happen for some reason
	  - see calls to `PackageCacheData` and md5 checks there, it should return `None, None` if the cache is working
- I explained what the test does and why it's this way, it was agreed that it can stay as is
- **I need to create an issue for long-term plans in conda-store. Jaime said it's not great that we rely on separate tools that know nothing about each other, which causes issues like this. I've been told there are plans to make conda more useable as a library**
- per-build temporary caches, like the ones proposed in this PR, is a no-go because it breaks how conda is supposed to be used, which will lead to problems
  - so the plan is to make global cache usable again. And if that's not possible, then the issues need to be addressed elsewhere, maybe upstream.

nkaretnikov:
Status update:
- I fixed the main issue that was causing 50% of builds to fail at random when running in parallel. That was due to tarballs that only have `info/index.json`, so `info/repodata_record.json` was created inside conda, with `channel` set to an invalid value, which never matched the real spec. That caused downloads to happen during lockfile installs, outside of the action where we have locking. And since there's no builtin locking in conda, builds having the same dependencies caused entries in the cache to be overwritten and corrupted.
- I fixed an issue that appeared right after I fixed the first one, due to the magic file sometimes missing in the cache.
- I fixed a race condition in the symlink helper that also appeared shortly and was repro'ing once in a while.
- I still need to test properly, so there might be other issues.

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -1,10 +1,24 @@
+import os
 import pathlib
 import shutil
 import typing
 
+# This import is needed to avoid the following error on conda imports:
+# AttributeError: 'Logger' object has no attribute 'trace'
```
jaimergp, line 7: Yes, this is really annoying! I've been meaning to fix that upstream for a while. There should be no global monkey patching of `logging` _just_ to add a utility function 😠

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -28,6 +42,29 @@ def action_fetch_and_extract_conda_packages(
...
+                # during the lockfile install action:
+                #
+                # File "/opt/conda/lib/python3.10/site-packages/conda/misc.py", line 110, in explicit
+                #   raise AssertionError("No package cache records found")
+                #
+                # The code below is from create_package_cache_directory in
+                # conda, which creates the package cache, but we only need the
+                # magic file part here:
+                cache_magic_file = pkgs_dir / PACKAGE_CACHE_MAGIC_FILE
+                if not cache_magic_file.exists():
```
jaimergp, line 63: This might still allow for a race condition, doesn't it? `try-except` constructions work better here. I wonder if we should fix this upstream too.

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -38,4 +75,96 @@ def action_fetch_and_extract_conda_packages(
...
                     with file_path.open("wb") as f:
                         shutil.copyfileobj(conda_package_stream, f)
-                    conda_package_handling.api.extract(str(file_path))
+                    conda_package_handling.api.extract(file_path_str)
+
+                    # This code is needed to avoid failures when building in
+                    # parallel while using the shared cache.
+                    #
+                    # There are tarballs that only have the info/index.json file
+                    # and no info/repodata_record.json. The latter is used to
```
jaimergp, line 84: I think `repodata_record.json` is _never_ in the remote tarball. It's actually created by `conda` upon extraction, taking `info.json` as the source, and extended with the actual info in `repodata.json`. See https://github.com/conda/conda/blob/6ec61a8445ae2663b134f8956104007827dafeab/conda/core/path_actions.py#L1430-L1441

conda-store-server/conda_store_server/action/download_packages.py
nkaretnikov: `if exists` might be prone to race conditions in some cases (but not always, see below). It's a classic [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue. `try/except` might be racy as well. It depends on where and how that's used. The general idea: you try using the shared resource right away without doing any checks and handle any failures if any later.

That said, I think it's actually fine to keep `if exists` here because I'm _assuming_ this file is never deleted once it's created. So this check is just to:
- avoid modifying file metadata such as creation/modification time for no reason (this might be useful later to understand when the cache was created or modified)
- avoiding unnecessary work and making it faster (I'm assuming the syscall that checks whether the file exists returns faster compared to the one that needs to populate the metadata).

No fix is needed upstream either because `create_package_cache_directory` has `try/except`. So it always calls `touch` on the file first before potentially throwing an error. Note, however, that the usecase is different here. I'm assuming `create_package_cache_directory` is called only once when initializing the cache while the code above is called for every build, so I want to avoid unnecessary work.

conda-store-server/conda_store_server/action/download_packages.py
nkaretnikov: FYI: It was discussed before in the context of a different issue and received a "won't fix" reply: https://github.com/conda/conda/issues/8919#issuecomment-510912585. So it might depend on what APIs are considered private in conda.

conda-store-server/conda_store_server/action/download_packages.py
nkaretnikov: Tested locally with an inverted condition just in case. That didn't find any matching packages, so removed this check.

nkaretnikov:
@jaimergp PTAL. See the updated comment in the PR description. I document all issues and fixes there.

conda-store-server/conda_store_server/action/download_packages.py
jaimergp: Yes, that was a different team in a different time. We will get this fixed in 2024, hopefully.

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -28,14 +43,149 @@ def action_fetch_and_extract_conda_packages(
...
+                        file_path_str = str(file_path)
+                        extracted_dir = pathlib.Path(
+                            strip_pkg_extension(file_path_str)[0]
+                        )
+                        extracted_dir_str = str(extracted_dir)
+                        context.log.info(f"DOWNLOAD {filename} | {count_message}\n")
+                        (
+                            filename,
+                            conda_package_stream,
+                        ) = conda_package_streaming.url.conda_reader_for_url(url)
```
jaimergp, line 83: Let's add `conda-package-streaming` to our environment files since we are using it directly.

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -28,14 +43,149 @@ def action_fetch_and_extract_conda_packages(
...
+                        # didn't work, versus the new one created by this action:
+                        #
+                        # --- /tmp/old.txt        2024-02-05 01:08:16.879751010 +0100
+                        # +++ /tmp/new.txt        2024-02-05 01:08:02.919319887 +0100
+                        # @@ -2,7 +2,7 @@
+                        #    "arch": "x86_64",
+                        #    "build": "conda_forge",
+                        #    "build_number": 0,
+                        # -  "channel": "<unknown>",
+                        # +  "channel": "https://conda.anaconda.org/conda-forge/linux-64",
```
jaimergp, line 141: Did you raise an issue upstream either in conda or conda-lock? I'd like this to be fixed properly at some point so we don't have to carry all this code around.

jaimergp: For example, one potential fix would be to tell `conda` to ignore `channel` if set to `<unknown>` (treat it like a wildcard, basically). Another fix would be stop conda-lock from dropping the channel information. I think I've seen this issue in the test suite too when working with conda-libmamba-solver, so it definitely needs a look.

conda-store-server/conda_store_server/action/download_packages.py
```diff
@@ -28,14 +43,149 @@ def action_fetch_and_extract_conda_packages(
             file_path = pkgs_dir / filename
             count_message = f"{packages_searched} of {total_packages}"
             with filelock.FileLock(str(lock_filename)):
```
jaimergp, line 45: I can't comment on the lines above, but I've seen instances where `package['url']` is None. You also have `fn` (filename) as a metadata field I think? In case you want to provide a fallback.

conda-store-server/conda_store_server/action/download_packages.py
nkaretnikov: Discussed offline. I'll open an issue in the conda-store repo first and then follow up.

conda-store-server/conda_store_server/action/download_packages.py
nkaretnikov: I'll keep this in mind, but I suggest we keep this as is for now. I was trying to copy the upstream code as close as possible and in `package_cache_data.py` it's always just the url:

```py
            url = self._urls_data.get_url(package_filename)
            package_cache_record = PackageCacheRecord.from_objects(
                raw_json_record,
                url=url,
                fn=basename(package_tarball_full_path),
                md5=md5,
                size=getsize(package_tarball_full_path),
                package_tarball_full_path=package_tarball_full_path,
                extracted_package_dir=extracted_package_dir,
            )

```

jaimergp (approved):
LGTM.

We are however using code from upstream projects with small modifications to work around some unfortunate issues for which we have little control over. This can result is added churn in the long run if we are not careful with how much the code deviates from upstream. For now it's ok, though, and maybe we are lucky and it's a relatively solid chunk of logic ;)

In the long term, I'd like to see "conda concurrency" solved elsewhere, though. Their PackageCache abstraction was not designed with parallel runners in mind, so there's only so much we can do before it bites us again. I recommend opening an issue with our latest findings, specially because there're tons of good info in this PR that would be good to summarize for actionability in the future.

PR #747, opened 2024-01-29
Add status `CANCELED`

### LINKED ISSUE #547: [ENH] - status for cancelled build
### Feature description

Once PR #531 is merged, we'll have the possibility to cancel builds.
In the current state of this feature, a cancelled build has a "Failed" status.

Implementing a clear "Cancelled" status would be clearer for the users. 
Implementing this would imply adding an Alembic migration

### Value and/or benefit

- Clearer build status for the users

### Anything else?

_No response_

nkaretnikov:
Fixes #547.

## Description

This pull request: adds status `CANCELED`, which is displayed in the web UI when the user cancels a build.

Screen Shot 2024-02-26 at 02 46 23

## Pull request checklist

## How to test

- Run via docker
- Cancel a build via the admin UI by clicking on the cancel button
- The status of the build should say "CANCELED".

### DIFF
--- a/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
+++ b/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
@@ -39,13 +39,3 @@ def upgrade():
 
 def downgrade():
     op.execute("DELETE FROM build WHERE status = 'CANCELED'")
-    with op.batch_alter_table(
-        "build",
-        schema=None,
-    ) as batch_op:
-        batch_op.alter_column(
-            "status",
-            existing_type=sa.VARCHAR(length=9),
-            type_=sa.VARCHAR(length=9),
-            existing_nullable=False,
-        )
--- a/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
+++ b/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
@@ -38,10 +38,7 @@ def upgrade():
 
 
 def downgrade():
-    # There is a foreign key constraint linking these two tables, so need to
-    # remove matching build artifacts first
-    op.execute(
-        "DELETE FROM build_artifact WHERE build_artifact.build_id IN "
-        "(SELECT id FROM build WHERE status = 'CANCELED')"
-    )
-    op.execute("DELETE FROM build WHERE status = 'CANCELED'")
+    # There are foreign key constraints linking build ids to other tables. So
+    # just mark the builds as failed, which was the status previously used for
+    # canceled builds
+    op.execute("UPDATE build SET status = 'FAILED' WHERE status = 'CANCELED'")
--- a/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
+++ b/conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
@@ -41,7 +41,7 @@ def downgrade():
     # There is a foreign key constraint linking these two tables, so need to
     # remove matching build artifacts first
     op.execute(
-        "DELETE FROM build_artifact BA WHERE BA.build_id IN "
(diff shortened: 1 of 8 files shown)


conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
```diff
@@ -0,0 +1,51 @@
...
+    with op.batch_alter_table(
+        "build",
+        schema=None,
+    ) as batch_op:
+        batch_op.alter_column(
+            "status",
+            existing_type=sa.VARCHAR(length=9),
+            type_=sa.VARCHAR(length=9),
+            existing_nullable=False,
+        )
```
nkaretnikov, line 51: The `batch_alter_table` block is not needed on downgrade. It'll be a no-op in both cases because the enum has been already converted to a VARCHAR at that point.

nkaretnikov: The whole migration is very similar to what we have for the installer, but it needs to be tested just in case:
- SQLite/Postgres
- downgrade/upgrade with existing data in the DB
- table schema changes (type) in Postgres
- entries are deleted on downgrade.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -986,8 +986,9 @@ async def api_put_build_cancel(
         tasks.task_cleanup_builds.si(
             build_ids=[build_id],
             reason=f"""
-    build {build_id} marked as FAILED due to being canceled from the REST API
+    build {build_id} marked as CANCELED due to being canceled from the REST API
     """,
+            is_canceled=True,
```
nkaretnikov, line 991: Test that this is printed to logs

conda-store-server/conda_store_server/worker/tasks.py
```diff
@@ -106,10 +106,15 @@ def task_update_storage_metrics(self):
...
+def task_cleanup_builds(
+    self,
+    build_ids: typing.List[str] = None,
+    reason: str = None,
+    is_canceled: bool = False,
+):
     conda_store = self.worker.conda_store
     with conda_store.session_factory() as db:
-        build_cleanup(db, conda_store, build_ids, reason)
+        build_cleanup(db, conda_store, build_ids, reason, is_canceled)
```
nkaretnikov, line 117: Test if this status also shows up and works in the new web UI.

conda-store-server/conda_store_server/server/views/api.py
nkaretnikov: Add a test for this

conda-store-server/conda_store_server/worker/tasks.py
nkaretnikov: conda-store-ui will need to be updated after this lands to support this better. Currently, it shows undefined for these builds because it doesn't know about this new status:
Screen Shot 2024-02-26 at 00 41 12

conda-store-server/conda_store_server/alembic/versions/03c839888c82_add_canceled_status.py
```diff
@@ -0,0 +1,41 @@
...
+    if not str(op.get_bind().engine.url).startswith("sqlite"):
+        op.execute("DROP TYPE IF EXISTS buildstatus")
+
+
+def downgrade():
+    op.execute("DELETE FROM build WHERE status = 'CANCELED'")
```
nkaretnikov, line 41: This migration does the same thing as the one we have for the installer:
https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/alembic/versions/57cd11b949d5_add_installer.py

The reason the `downgrade` part is different is because it's a no-op here with canceled, since the field size is the same.

dcmcand (approved):
Looks good. Thanks @nkaretnikov :rocket:

trallard:
@gabalafou I have not verified if this is working as expected, could you check please?

Issue #561, opened 2023-09-06
META - Populate documentation with initial content


pavithraes:
This is the list of new documentation pages that need to be created. 

Convert the items to issues as you work on them to avoid overloading the issue tracker.

```[tasklist]
### conda-store
- [x] Introduction
- [ ] https://github.com/conda-incubator/conda-store/issues/566
- [ ] https://github.com/conda-incubator/conda-store/issues/568
- [x] Tutorial - Using shebang (https://conda-incubator.github.io/conda-store-ui/?path=/docs/docs-configuration--page)
- [x] How-to - Configure conda-store (https://conda.store/en/latest/administration.html#configuration)
- [ ] https://github.com/conda-incubator/conda-store/issues/569
- [x] Explanation - Resources and performance (https://conda.store/en/latest/administration.html#resource-requirements, https://conda.store/en/latest/administration.html#performance)
- [x] Explanation - Architecture (https://conda.store/en/latest/contributing.html#architecture)
- [x] Explanation - Databases (https://conda.store/en/latest/contributing.html#database)
- [x] https://github.com/conda-incubator/conda-store/issues/754
```

> **Note**
> **conda-store-ui tutorial section is blocked** on having automated screenshots using Playwright to keep the docs in sync with the rapid changes to the UI.

```[tasklist]
### conda-store-ui
- [x] Introduction
- [x] Tutorial - Install & Setup
- [x] Tutorial - Login/Logout
- [x] Tutorial - Create and edit envs
- [x] Tutorial - Delete envs - #563
- [x] Tutorial - Version control (switch to diff version)
- [x] How-to - Configure conda-store-ui (https://conda-incubator.github.io/conda-store-ui/?path=/docs/docs-configuration--page)
- [x] References - Storybook components
```

```[tasklist]
### jupyterlab-conda-store
- [x] Introduction - #752
- [x] Install & Setup - #752
```

```[tasklist]
### Community
- [ ] https://github.com/conda-incubator/conda-store/issues/399
- [x] Design assets -- xref: https://github.com/conda-incubator/conda-store-ui/issues/344
- [x] Contributing docs
- [x] Maintainer docs
```

pavithraes:
All tasks except the REST API (already have a PR open) are done, so calling this META complete :)

PR #752, opened 2024-01-30
[DOC] Add JupyterLab Extension docs page

pavithraes:
Addresses #561 

## Description

This pull request:

- Creates a landing page for JLab extension with installation instructions.

## Pull request checklist

- [ ] Did you add/update relevant tests for this change (if required)? N/A


docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
 ---
-sidebar_position: 1
 title: Introduction
 description: Introduction to JupyterLab Extension.
 ---
 
 # conda-store JupyterLab extension
+
+A extension to use the [conda-store UI][conda-store-ui] - a React-based frontend for conda-store, within JupyterLab.
```
gabalafou, line 8:
```suggestion
An extension to use the [conda-store UI][conda-store-ui] - a React-based frontend for conda-store, within JupyterLab.
```

docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
...
+
+## Install 📦
+
+1. Pre-requisites: `conda-store-server`, JupyterLab `>=` 3.0 and `<=` 4.0, and Python `>=` 3.8 installed.
```
gabalafou, line 12: Is it necessary to put the comparison operators in backticks?

docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
...
+
+2. Install the extension:
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
```
gabalafou, line 17: Is it standard to colocate imports with call sites in these kinds of docs?

docusaurus-docs/jupyterlab-conda-store/introduction.md
gabalafou, line 1: Should this extension be .mdx, since it uses JSX?

docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
...
+
+<Tabs>
```
gabalafou, line 19: Oh neat. What are these `<Tabs>` and `<TabItem>`? Are these a MUI thing?

docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
...
+
+```bash
+pip uninstall jupyterlab-conda-store
+```
+</TabItem>
+</Tabs>
+
+## Usage
+
+In the JupyterLab window, click on `conda-store` menu-bar item to open the UI in a new window within JupyterLab.
```
gabalafou, line 64:
```suggestion
In the JupyterLab window, click on the `conda-store` menu bar item to open the UI in a new window within JupyterLab.
```

gabalafou, line 64: Would be nice to have a screen shot here of opening the conda-store from within JupyterLab

docusaurus-docs/jupyterlab-conda-store/introduction.md
```diff
@@ -1,7 +1,71 @@
 ---
-sidebar_position: 1
```
gabalafou, line 2: Seems a little odd to show the sidebar at all:

image

docusaurus-docs/docusaurus.config.js
```diff
@@ -183,6 +183,11 @@ const config = {
           '⚠️ We are in the process of revamping our docs, some pages may be incomplete or inaccurate. ⚠️',
         isCloseable: false,
       },
+      docs: {
+        sidebar: {
+          hideable: true,
```
gabalafou, line 188: Did you intend to hide the sidebar for the page you added? Why is this change in this PR?

gabalafou (changes requested):
This looks nice! I added a few comments, I think at least a couple of them will need to be addressed

docusaurus-docs/docusaurus.config.js
pavithraes: Yes, this is intentional. We don't have any more pages for the JLab extension (all usage actions are similar to conda-store-ui), so the sidebar will be empty - hence hiding it.

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: Yeah, I agree, but Docusaurus currently doesn't have a way for us to do this with our setup.

First, there is no way to "hide" the page fully. We can also set the sidebar as "collapsed" by default.
BUT - that's only possible to do globally. We can't do this because we're using the [multi-instance](https://docusaurus.io/docs/docs-multi-instance) feature to have all docs in the same place, and thesidebar is useful for conda-store & conda-store-ui.

Does this makes sense?

If we feel strongly, one alternative solution is to have a custom static page instead of using Docusaurus-generated pages.

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: Yes, with the more recent MDX versions, the `<, >` characters cause issues. Ref: https://docusaurus.io/blog/releases/3.0#markdown

docusaurus-docs/docusaurus.config.js
pavithraes: xref: https://github.com/conda-incubator/conda-store/pull/752/files#r1504642431

There is actually no way to completely hide it, hence this option to make the experience slightly nicer.

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: I'm not sure how these are implemented. Docusaurus provides this by default, ref docs: https://docusaurus.io/docs/markdown-features/tabs#displaying-a-default-tab

docusaurus-docs/jupyterlab-conda-store/introduction.md
gabalafou: Oh right. I wonder if you can do could use HTML entities like `&gt;` (greater than) and `&lt;` (less than), which I believe is supported by GitHub flavored markdown so might be supported in these docs but I don't know for sure. But this is non-blocking ;)

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: Well, I believe so? I've been using this style convention for a while now across projects. But'll try to verify.

docusaurus-docs/docusaurus.config.js
gabalafou: Ah, I think I see now. 

As far as I understand, technically this change isn't related to the other changes in this PR. My preference would be to put this change in a separate PR, just to keep things neat and tidy, but not going to block on it

docusaurus-docs/jupyterlab-conda-store/introduction.md
gabalafou: No, don't feel strongly, just wasn't sure if it had been noticed.  :)

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: Oh nice, I'll update, thanks!

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: Indeed, added!

docusaurus-docs/jupyterlab-conda-store/introduction.md
pavithraes: I was wrong! There are some colocated imports in the Docusaurus docs itself, but looks like most users define it at the top. I'll move this import to the topm and make a fresh PR to update this everywhere else in the docs.

docusaurus-docs/docusaurus.config.js
pavithraes: I'd actually prefer to include it here because this docs page is what prompted the update.

pavithraes:
@gabalafou - Thanks for a thorough review, I've addressed all your comments. :)

gabalafou (approved):

PR #772, opened 2024-03-03
Add ability to create environment from lockfile

### LINKED ISSUE #676: [ENH] - Add ability to create environment from lockfile
### Feature description

The only(ish) way to guarantee a fully reproduced environment is through lockfiles. We produce them, but that's only truly useful if we can USE them. 

I propose that conda-store add the ability to create environments from lockfiles. 

### Value and/or benefit

Reproducibility! 

### Anything else?

_No response_

nkaretnikov:
Fixes #676.

## Description

This pull request:

- adds a way to create an environment from a lockfile
- adds a test.

## Pull request checklist

## Additional information

- conda-store-ui will need to be extended to support this feature because it has its own YAML parser
- I didn't add documentation for this reason as well. Despite being in this repo, most of our docs are about conda-store-ui. I think it needs to be implemented on the UI side before we document it in user-visible docs, otherwise it'll be confusing.

### Automated testing

- Run the server via docker:

```bash
docker compose up --build
```

- Once the server is running, the following command should succeed:

```py
pytest -rfxs tests/test_api.py::test_create_lockfile_specification_auth -vvv --capture=no
```

### Manual testing

- Submit a specification like the following:

Screen Shot 2024-03-04 at 04 26 22

- A whitespace-sensitive version is also accepted:

Screen Shot 2024-03-04 at 04 07 34

- **Do not forget to click on the checkbox**

- The keys are: `name`, `description`, and `lockfile`. conda-store requires that a specification has a name, which is why this wraps an upstream conda-lock lockfile format instead of passing it directly as a specification. From `conda-store-server/conda_store_server/schema.py`:

```py
class LockfileSpecification(BaseModel):
    name: constr(regex=f"^[{ALLOWED_CHARACTERS}]+$")  # noqa: F722
    description: Optional[str] = ""
    lockfile: Lockfile
```

-   `lockfile` corresponds to a lockfile that conda-store produces, which is the [v1 conda-lock lockfile](https://github.com/conda/conda-lock/blob/38ec3955c52ef05ca8d6cd47a4c46e0624372f95/conda_lock/lockfile/v1/models.py#L296):

```py
class Lockfile(StrictModel):
    version: ClassVar[int] = 1

    package: List[LockedDependency]
    metadata: LockMeta
```


conda-store-server/conda_store_server/schema.py
```diff
@@ -463,6 +465,49 @@ def parse_obj(cls, specification):
...
+            )
+
+        return super().parse_obj(specification)
+
+    def dict(self):
+        res = super().dict()
+        # The dict_for_output method includes the version field into the output
+        # and excludes unset fields. Without the version field present,
+        # conda-lock would reject a lockfile during parsing, so it wouldn't be
+        # installable, so we need to include the version
```
nkaretnikov, line 501:
```suggestion
        # installable, that's why we need to include the version
```

conda-store-server/conda_store_server/orm.py
```diff
@@ -142,20 +143,22 @@ class Specification(Base):
...
         self.spec = specification
         self.sha256 = utils.datastructure_hash(self.spec)
+        self.is_lockfile = is_lockfile
 
     id = Column(Integer, primary_key=True)
     name = Column(Unicode(255), nullable=False)
     spec = Column(JSON, nullable=False)
     sha256 = Column(Unicode(255), unique=True, nullable=False)
     created_on = Column(DateTime, default=datetime.datetime.utcnow)
+    is_lockfile = Column(Boolean, default=False)
```
nkaretnikov, line 161:
```suggestion
    is_lockfile = Column(Boolean, nullable=False)
```

conda-store-server/conda_store_server/app.py
```diff
@@ -663,15 +673,23 @@ def register_environment(
...
         return build.id
 
-    def create_build(self, db: Session, environment_id: int, specification_sha256: str):
+    def create_build(
+        self,
+        db: Session,
+        environment_id: int,
+        specification_sha256: str,
+        is_lockfile: bool = False,
+    ):
```
nkaretnikov, line 692:
```suggestion
    def create_build(self, db: Session, environment_id: int, specification_sha256: str):
```

Revert this back, `is_lockfile` is not used later (it's read from the DB in that function).

conda-store-server/conda_store_server/action/generate_lockfile.py
```diff
@@ -61,3 +61,17 @@ def action_solve_lockfile(
...
     with lockfile_filename.open() as f:
         return yaml.safe_load(f)
+
+
+@action.action
+def action_save_lockfile(
+    context,
+    specification: schema.LockfileSpecification,
+):
+    lockfile = specification.dict()["lockfile"]
```
nkaretnikov, line 71: Add a comment here explaining that it's necessary to call it this way to serialize lockfile properly, otherwise version will not be in the output.

.github/workflows/tests.yaml
```diff
@@ -85,6 +85,15 @@ jobs:
         shell: bash -el {0}
         working-directory: conda-store-server
     steps:
+      # https://github.com/marketplace/actions/debugging-with-tmate
```
nkaretnikov, line 88: Remove this block

nkaretnikov:
I've just noticed this fails for existing non-lockfile environments if you attempt a rebuild. For some reason, it attempts to rebuild as a lockfile env. I'll look into it.

nkaretnikov:
@jaimergp A gentle reminder that this needs review.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -20,13 +21,20 @@ def get_installer_platform():
     return context.subdir
 
 
+# TODO: Ideally, this would just use the lockfile instead of reading data
+# from the specification. But we currently also need to support legacy lockfiles
+# generated by conda-store, which were not generated by conda-lock, and those
+# don't have enough information for constructor to generate an installer. In
```
jaimergp, line 27: I don't think this is true. You can pass the `explicit-list.txt` file (they contain an `@EXPLICIT` header) to [`environment_file`](https://conda.github.io/constructor/construct-yaml/#environment-file) and it should work. This is a relatively new feature introduced last year so it might need some polishing, but it's [part of the test suite](https://github.com/conda/constructor/blob/28309ac0d03093101de222d445c3150c1498ba90/tests/test_examples.py#L530-L541).

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -52,11 +60,28 @@ def write_file(filename, s):
...
+    if is_lockfile:
+        # Adds channels
+        channels = [c.url for c in specification.lockfile.metadata.channels]
+
+        # Adds dependencies
+        for p in specification.lockfile.package:
+            if p.manager == "pip":
+                pip_dependencies.append(f"{p.name}=={p.version}")
+            else:
+                dependencies.append(f"{p.name}=={p.version}")
```
jaimergp, line 74: What is this block for? Are we adding _all_ the packages listed to the database? These are not necessarily the "input packages" for the environment. It's also true that there are no easy ways to fetch that from the `metadata.sources` info, though.

Regardless, this conda spec should also contain the build string, which can be parsed from the URL:

```suggestion
        # Adds dependencies
        for p in specification.lockfile.package:
            if p.manager == "pip":
                pip_dependencies.append(f"{p.name}=={p.version}")
            else:
                ext = ".tar.bz2" if p.url.endswith(".tar.bz2") else ".conda"
                build_string = p.url.split("/")[-1][:-len(ext)].rsplit("-", 2)[-1]
                dependencies.append(f"{p.name}=={p.version}={build_string)")
```

(Review that code, just wrote it on the spot, no testing).

conda-store-server/conda_store_server/server/templates/create.html
```diff
@@ -26,6 +26,8 @@
         <label for="specification">Specification:</label>
         <div id="editor" style="height: 400px;">{{ specification or "" }}</div>
         <textarea name="specification" id="specification" style="display: none;"></textarea>
+        <input type="checkbox" name="is-lockfile" id="is-lockfile" />
+        <label for="is-lockfile">This specification contains a <a href="https://github.com/conda-incubator/conda-lock">Conda-Lock</a> lockfile</label>
```
jaimergp, line 30: I don't like this checkbox. This should be a dropdown with all possible inputs (more are possibly coming, like `explicit.txt` lockfiles). For now: `environment.yml`, `conda-lock.yml` are the options.

conda-store-server/conda_store_server/server/templates/create.html
```diff
@@ -64,6 +66,7 @@
      let url = "{{ url_for('api_post_specification') }}";
      let namespaceInput = document.querySelector('#namespace');
      let specificationInput = document.querySelector('#specification');
+     let isLockfileInput = document.querySelector('#is-lockfile');
```
jaimergp, line 69: Adjust for dropdown.

conda-store-server/conda_store_server/server/templates/create.html
```diff
@@ -73,6 +76,7 @@
          body: JSON.stringify({
              'namespace': namespaceInput.value,
              'specification': specificationInput.value,
+             'is_lockfile': isLockfileInput.checked,
```
jaimergp, line 79: This should just be string field like `input_type`.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -790,6 +790,7 @@ async def api_post_specification(
     entity=Depends(dependencies.get_entity),
     specification: str = Body(""),
     namespace: Optional[str] = Body(None),
+    is_lockfile: Optional[bool] = Body(False, embed=True),
```
jaimergp, line 793: Adjust for dropdown menu.

conda-store-server/conda_store_server/orm.py
```diff
@@ -142,20 +143,22 @@ class Specification(Base):
...
         self.spec = specification
         self.sha256 = utils.datastructure_hash(self.spec)
+        self.is_lockfile = is_lockfile
 
     id = Column(Integer, primary_key=True)
     name = Column(Unicode(255), nullable=False)
     spec = Column(JSON, nullable=False)
     sha256 = Column(Unicode(255), unique=True, nullable=False)
     created_on = Column(DateTime, default=datetime.datetime.utcnow)
+    is_lockfile = Column(Boolean, nullable=False)
```
jaimergp, line 161: Same here, store the input type instead.

conda-store-server/conda_store_server/dbutil.py
```diff
@@ -71,7 +73,11 @@ def upgrade(db_url, revision="head"):
     revision: str [default: head]
         The alembic revision to upgrade to.
     """
-    engine = create_engine(db_url)
+    engine = create_engine(
+        db_url,
+        # See the comment on the CustomJSONEncoder class on why this is needed
+        json_serializer=lambda x: json.dumps(x, cls=utils.CustomJSONEncoder),
```
jaimergp, line 79: I think this could also be a `functools.partial` like

```suggestion
        json_serializer=partial(json.dumps, cls=utils.CustomJSONEncoder),
```

... in case the lambda feels awkward.

conda-store-server/conda_store_server/server/templates/create.html
```diff
@@ -26,6 +26,8 @@
         <label for="specification">Specification:</label>
         <div id="editor" style="height: 400px;">{{ specification or "" }}</div>
         <textarea name="specification" id="specification" style="display: none;"></textarea>
```
jaimergp, line 28: I don't like how we are using the same textarea for two different things:

- When the input is `environment.yml`, we have all the information we need. This is good.
- When the input is a conda-lock, we need `name` and `description`. Since these fields are not part of the conda-lock schema, we embed the lockfile in a `lockfile` key. This is weird for the user because they will want to assume this collects the contents of the file. But it's not, they will be forced to edit after copy pasting.

The best solution IMO would be to revamp the UI a bit:

- As mentioned before, place a dropdown _above_ the text area so users can choose the type of input. The dropdown will have some events attached:
  - If the input file is a regular environment.yml, present the normal textarea
  - If the input file is a lockfile, then add two input fields with `name` and `description`. The content of the lockfile will be added to the textarea. For the backend, these three fields can be merged together in the data model you created (name, description and lockfile), but for the user shouldn't have to carry that abstraction burden.

That said, is `description` part of the `environment.yml` specification, or are we just adding it as a free form key because conda doesn't complain? Because the cheap way out of this for now would be, if conda-lock allows that, to add `name` and `description` top-level keys, and some sort of validation. But I don't like that hack.

jaimergp (changes requested):
I haven't tested this live, but I already have a few comments about the initial implementation. See inlines for more details, but to sum up:

- Use dropdown menu instead of a checkbox
- Don't push data model abstractions to the UI (the user should not need to manipulate their lockfile when copy-pasting). Instead, add extra fields as needed.
- constructor does support `explicit.txt` lockfiles which can be easily generated from the conda-lock.yml.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
nkaretnikov: Resolved. I've looked more into this. Turns out supporting explicit-lockfiles is not necessary here. First, they don't include pip dependencies, so the installer would miss those. Additionally, environments generating constructor artifacts will never have explicit-lockfiles, because builds using constructor would use new-style env-lockfiles, as env-style is the new default method (see `schema.BuildArtifactType.LOCKFILE` in `build_conda_environment`).

So I've changed the code that when a conda-store spec is provided, it tries to look up the lockfile and use that. There's a try/catch because it's in the external storage and is generated as part of a different task, but it works for me locally. On an exception, it just falls back to using the specification (non-pinned).

nkaretnikov:
@jaimergp PTAL. Will rebase after this is approved.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -53,11 +56,38 @@ def write_file(filename, s):
...
+        channels = [c.url for c in specification.lockfile.metadata.channels]
+
+        # Adds dependencies
+        for p in specification.lockfile.package:
+            # Ignores packages not matching the current platform. Versions can
+            # be different between platforms or a package might not support all
+            # platforms. constructor is cross-friendly, but we're currently
+            # building only for the current architecture, see the comment in
+            # get_installer_platform
+            if p.platform != get_installer_platform():
```
jaimergp, line 72: Is this excluding `noarch` too? It shouldn't.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -53,11 +56,38 @@ def write_file(filename, s):
...
+                build_string = p.url[: -len(ext)].rsplit("-", maxsplit=1)[-1]
+                dependencies.append(f"{p.name}=={p.version}={build_string}")
+
+    else:
+        # Adds channels
+        channels = specification.channels
+
+        # Adds dependencies
+        for d in specification.dependencies:
+            if type(d) is schema.CondaSpecificationPip:
```
jaimergp, line 87:
```suggestion
            if isinstance(d, schema.CondaSpecificationPip):
```

Any reason to not use the more idiomatic `isinstance`?

conda-store-server/conda_store_server/action/generate_constructor_installer.py
```diff
@@ -53,11 +56,38 @@ def write_file(filename, s):
     # conda and pip need to be in dependencies for the post_install script
     dependencies = ["conda", "pip"]
     pip_dependencies = []
-    for d in specification.dependencies:
-        if type(d) is schema.CondaSpecificationPip:
-            pip_dependencies.extend(d.pip)
-        else:
-            dependencies.append(d)
+    channels = []
```
jaimergp, line 59: We don't need this, don't we? We define `channels` in both branches.

conda-store-server/conda_store_server/build.py
```diff
@@ -416,13 +432,38 @@ def build_constructor_installer(db: Session, conda_store, build: orm.Build):
...
+                    specification = schema.LockfileSpecification.parse_obj(
+                        {
+                            "name": build.specification.name,
+                            "lockfile": json.loads(
+                                conda_store.storage.get(build.conda_lock_key)
+                            ),
+                        }
+                    )
+                    is_lockfile = True
+                except Exception:
```
jaimergp, line 456:
```suggestion
                except Exception as exc:
                    logger.debug("Exception while obtaining lockfile", exc_info=exc)
```

Maybe we should log the exception for easier debuggability down the line.

conda-store-server/conda_store_server/server/views/api.py
```diff
@@ -811,7 +814,14 @@ async def api_post_specification(
 
         try:
             specification = yaml.safe_load(specification)
-            specification = schema.CondaSpecification.parse_obj(specification)
+            if is_lockfile:
+                lockfile_spec = {}
+                lockfile_spec["name"] = environment_name
+                lockfile_spec["description"] = environment_description
+                lockfile_spec["lockfile"] = specification
```
jaimergp, line 821:
```suggestion
                lockfile_spec = {
                    "name": environment_name,
                    "description": environment_description,
                    "lockfile": specification,
                }
```

jaimergp (commented):
Some more comments for now. I need to take a look at this on the UI to see how the UX feels, but we can start addressing the simple changes while I do that. Thanks!

jaimergp:
Good news ✅ Tests pass locally as instructed in the PR description.

Bad news 😢 I observed some surprising behavior during manual tests. What I did:

1. I started the server with `docker compose up --build`
2. Connected to http://localhost:8080 via the browser
3. Logged with `jaime` / `password` (this info could be added to the "running in localhost server, I guess?)
4. Started to create a new environment with name `lockfiletest`
5. Changed the editor to YAML mode and pasted the [JSON sample lock file](https://github.com/conda-incubator/conda-store/blob/588e86403aefb2d829d400468b04078e4ac11150/conda-store-server/tests/assets/conda-lock.zlib.flask.yaml) there.
6. Waited for the build to finish successfully, and so it did (at least no errors were reported).

Things I noticed:

- The editor uses a toggle `GUI <=> YAML`. However it allows JSON too. I think `YAML` is misleading and should just say `Text` or `Editor` or something like that.
- I didn't receive any feedback about having detected a lockfile or a YAML. If we are just letting things through and the checks are done server side, at least we should have a checkbox above the editor saying which file formats are accepted there.
- While building, the reported input specs only covered the usually-injected `ipykernel`, which was surprising. I was expecting all the specs from the lockfile here. I thought this was just a UI glitch, because no packages should have been injected in the lockfile, but...
  - It was not! The lockfile ipykernel wanted [6.28](https://github.com/conda-incubator/conda-store/blob/588e86403aefb2d829d400468b04078e4ac11150/conda-store-server/tests/assets/conda-lock.zlib.flask.yaml#L205), yet we received 6.29.3 !!! See attached [`calculated-lockfile.json`](https://github.com/conda-incubator/conda-store/files/15099225/calculated-lockfile.json). This defeats the purpose of the lockfile, imo.
  - My take: We should never alter the lockfile. If the namespace configuration requires injected specs, what we should do is simply check if the lockfile contents satisfy those injected specs and error out otherwise. If satisfied, then the UI should say something like "Satisfied" or "Found". We should never attempt to modify the input lockfile contents.
    - If we do want to support that, then we should explicitly name it "Import from lockfile". This was very surprising imo.
    - Or instead of erroring out, warn that the injected specs altered the lockfile contents in a certain way, but it will most probably change if we touch it so I don't see the point on this angle.
  - We should add tests for the above behaviour. Input lockfile and output lockfile should have the same `content_hash` for the tested platforms. Right now the hashes are different, probably because of the injected spec issue above.
- Once the environment was built, the UI didn't report any elements that told the user this env was imported from a lockfile. The input lockfile was not stored anywhere so I couldn't retrieve it either. 

So all in all, I don't think this PR is ready for merge yet. We need to discuss which behavior is expected and properly report it on the UI. Namely, I was expecting this flow:

1. User imports lockfile.
2. The UI reports if it complies with the configured constraints.
3. If it does, go ahead and save it / install it. 
  - The output lockfile should be the same or equivalent (same content_hash for that platform), meaning it wasn't modified.
  - The UI reports this env was imported from a lockfile, which should tell the user that is better to not modify it too much.
4. Optional convenience: If it doesn't, offer to import it / convert it to environment.yaml and modify til constraints are satisfied (I guess that by removing version pins on the affected specs).


Thanks for the patience and sorry about the delays in my reviews!

nkaretnikov:
@jaimergp 

> Bad news 😢 I observed some surprising behavior during manual tests. What I did:
> [...] Connected to http://localhost:8080 via the browser

You're accessing a different URL. conda-store is split into the UI part (different repo) and the server part (this repo). What you're trying to access above is the UI code, which is not here. Based on your feedback in this PR, I added code in this repo to allow users to paste lockfiles via the admin interface. You need to do `docker compose up --build` and go to `http://localhost:8080/conda-store/admin/`. There you'll see the UI for adding lockfiles.

Once this is approved and merged, we can add support for lockfiles in the UI repo. I indicated this in the first message: "conda-store-ui will need to be extended to support this feature because it has its own YAML parser." Our usual process is: server -> design (optional) -> UI.

jaimergp:
Ok, I see, thanks! I edited the above comment so it makes sense on its own now, this is the review for the admin UI:

- Lockfile identity is kept between input and output.
- Dropdown shows extra fields as needed.
- The admin UI is missing the pip packages in the HTML table. I can only see the conda ones.
- When I pasted [the lockfile](https://github.com/conda-incubator/conda-store/blob/588e86403aefb2d829d400468b04078e4ac11150/conda-store-server/tests/assets/conda-lock.zlib.flask.yaml), the resulting pinned environment.yml includes `defaults` in the channels, but it should only have `main`. See [pinned-env.yml.txt](https://github.com/conda-incubator/conda-store/files/15100142/pinned-env.yml.txt).
- Only `conda-lock.yml` files are supported. `@EXPLICIT`-style txt lockfiles are not. I think we should use a more precise term instead of `lockfile` in the dropdown. Maybe `conda-lock` is enough. In the future we could have `explicit` as an additional option if necessary.
- If I purposefully introduced broken JSON syntax, I get an error saying "Unable to parse. Invalid YAML". However, it's not YAML. I think we should just say "invalid contents", and --optionally-- either link to the documentation page (if any) that says which formats are supported, or a brief description in-place.


So yea, that's all I think. Mostly trivial things and hopefully easy to fix. A bit unsure about the weird `defaults` in the rendered environment.yaml though. Let me know if you need assistance there.

conda-store-server/conda_store_server/action/generate_constructor_installer.py
nkaretnikov: Updated. However, it might be unnecessary here. I've tried installing two `noarch` packages. They still have the platform set to `linux-64` despite coming from a `noarch` url. I think it's because of how we call `action_solve_lockfile`, which always passes `conda_utils.conda_platform()` to conda-lock.

<details>
<summary>Repro</summary>

spec:
```
channels:
- conda-forge
dependencies:
- pytest
- pip:
  - flask
- conda-forge/noarch::pytest-timeout
- conda-forge/noarch::pytest-mock
description: ''
name: test
prefix: null
variables: null
```

lockfile:
```
        {
            "name": "pytest-mock",
            "version": "3.14.0",
            "manager": "conda",
            "platform": "linux-64",
            "dependencies": {
                "python": ">=3.8",
                "pytest": ">=5.0"
            },
            "url": "https://conda.anaconda.org/conda-forge/noarch/pytest-mock-3.14.0-pyhd8ed1ab_0.conda",
            "hash": {
                "md5": "4b9b5e086812283c052a9105ab1e254e",
                "sha256": "f7cd8910ce3690a9189483f4b01dde1970b2b86ea853a03b1225f344f988045d"
            },
            "category": "main",
            "optional": false
        },
        {
            "name": "pytest-timeout",
            "version": "2.3.1",
            "manager": "conda",
            "platform": "linux-64",
            "dependencies": {
                "python": ">=3.7",
                "pytest": ">=7.0.0"
            },
            "url": "https://conda.anaconda.org/conda-forge/noarch/pytest-timeout-2.3.1-pyhd8ed1ab_1.conda",
            "hash": {
                "md5": "27860d8e6cc9e14da4ea900788693e40",
                "sha256": "501e2417f23596815aeb41ad3bc69016a59c9f5dab62860b8be41e006d538fe8"
            },
            "category": "main",
            "optional": false
        },
```

</details>

nkaretnikov:
@jaimergp 

Updated, PTAL. Replies:

> The admin UI is missing the pip packages in the HTML table. I can only see the conda ones.

This is a pre-existing issue, created https://github.com/conda-incubator/conda-store/issues/815. I suggest to postpone and fix this separately since (1) it's not related, (2) the admin UI is not something end-users are interacting with anyway. There was also a discussion about removing the admin UI, so I wouldn't invest a lot of energy into it.

> When I pasted the lockfile, the resulting pinned environment.yml includes defaults in the channels, but it should only have main. See pinned-env.yml.txt.

Also pre-existing, created https://github.com/conda-incubator/conda-store/issues/814. Suggest to postpone/fix later for the same reasons.

> Only conda-lock.yml files are supported. @EXPLICIT-style txt lockfiles are not. I think we should use a more precise term instead of lockfile in the dropdown. Maybe conda-lock is enough. In the future we could have explicit as an additional option if necessary.

Renamed to "unified lockfile" according to https://conda.github.io/conda-lock/output/#unified-lockfile. We have no docs for the admin UI page and there was a discussion about potentially removing the admin UI. I'll document lockfiles when I add the conda-store-ui PR since that's the user-facing part and it has the doc page.

> If I purposefully introduced broken JSON syntax, I get an error saying "Unable to parse. Invalid YAML". However, it's not YAML. I think we should just say "invalid contents", and --optionally-- either link to the documentation page (if any) that says which formats are supported, or a brief description in-place.

There's no JSON. Lockfiles are YAML: they even have the yml extension. It's just that they can be written with curly braces (JSON-style) or use indentation, [see here](https://medium.com/@kasunbg/write-yaml-without-indentation-via-curly-braces-3c05ae8700ce). They are parsed by the YAML parser in `api_post_specification`. The proposed error message sounds way too generic IMO: "contents" can mean anything on the entire HTML page. This is also an existing error message that was there before, so I'm hesitant to change it.

jaimergp (approved):

Issue #669, opened 2023-11-22
Integration Test: Login and delete shared environment


dcmcand:
### Context

Part of #662

Integration test for the following user story:

#### User Story
Persona: I'm a tech lead that managed envs for my team
Need: I want to delete an environment

steps:
1. Login
2. Delete a shared environment

#### Test
TODO: @dcmcand work out api calls and setup needed
##### Setup
1.
2.
##### API Calls
1.
2.

### Value and/or benefit

This will ensure that new releases of conda-store will not break critical user journeys.

### Anything else?

Note: This integration test only covers conda-store-server and this workflow should be tested at the api level. UI logic will be tested in the [conda-store-ui](https://github.com/conda-incubator/conda-store-ui) repo.

Issue #662, opened 2023-11-14
META - testing improvements


dcmcand:
### Context

This is a meta issue to track improvements that need to be made to conda-store's testing.

### Value and/or benefit

Our test suite should provide the following:

1. Rapid feedback to developers - we want failures to occur as early as possible in the pipeline so developers can get feedback on their changes rapidly.
2. Relevance - We should test things that we control. In practice this means not testing other people's code. We should be using pinned versions of dependencies for integration tests, mocked responses for end to end tests, and only testing our own code for unit tests.
3. Consistency - We need to ensure that all of our tests are consistent and deterministic. Flaky tests are annoying, slow down feedback loops, and undermine confidence in the system.
4. Confidence - By the end of our test pipeline we should have high confidence that what we release will not break things for users and that our code works as we intend. This includes both for functional requirements, and nonfunctional requirements such as security and accessibility.
5. Comprehensive coverage - all important user journeys and code paths should be comprehensively tested.

```[tasklist]
### Tasks
- [ ] https://github.com/conda-incubator/conda-store/issues/619
- [ ] https://github.com/conda-incubator/conda-store/issues/668
- [ ] https://github.com/conda-incubator/conda-store/issues/669
- [ ] https://github.com/conda-incubator/conda-store/issues/670
- [ ] https://github.com/conda-incubator/conda-store/issues/671
- [ ] https://github.com/conda-incubator/conda-store/issues/672
- [ ] https://github.com/conda-incubator/conda-store/issues/673
- [ ] https://github.com/conda-incubator/conda-store/issues/674
```

PR #776, opened 2024-03-08
Add user journey for logging in and deleting shared environment

peytondmurray:
Fixed https://github.com/conda-incubator/conda-store/issues/669.

## Description

This PR implements an integration test which checks that an admin can create an environment in a shared namespace and delete both the environment and the namespace.

This pull request:

- Adds an integration test which logs a user in and deletes an environment in a shared namespace and the namespace itself
- Modifies the existing `test_admin_user_can_create_environment` test; previously, that test was doing something very similar to what the new test implemented in this PR does. With this change, the existing test now tests that an admin can create an environment in the `default` namespace, more closely matching the existing docstring.
- This PR refactors the `user_journeys.utils.API` class to wait for responses from the conda-store server so that we don't need to do response handling in every test we write.

## Pull request checklist

## Additional information

- See the https://github.com/conda-incubator/conda-store/issues/662 meta-issue for additional context.

## How to test

1. Load the conda-store server with `docker compose up`
2. Run the test with `pytest conda-store-server/tests/user_journeys/test_user_journeys.py`

Alternatively, these tests are run as part of CI.

### DIFF
--- a/conda-store-server/tests/user_journeys/test_user_journeys.py
+++ b/conda-store-server/tests/user_journeys/test_user_journeys.py
@@ -56,7 +56,9 @@ def test_admin_login_and_delete_shared_environment(
     environment = api.create_environment(
         namespace,
         specification_path,
-    ).json()["data"]["specification"]["name"]
+    ).json()["data"][
+        "specification"
+    ]["name"]
 
     api.delete_environment(namespace, environment)
     api.delete_namespace(namespace)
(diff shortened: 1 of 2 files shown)


conda-store-server/tests/user_journeys/test_user_journeys.py
```diff
@@ -29,15 +29,32 @@ def test_admin_user_can_create_environment(
...
+@pytest.mark.parametrize(
+    "specification_path",
+    [
+        ("tests/user_journeys/test_data/simple_environment.yaml"),
+    ],
+)
+def test_login_and_delete_shared_environment(
+    base_url: str, specification_path: str
+) -> None:
+    """Test that a user can login and delete an environment in a shared namespace."""
```
dcmcand, line 50: This is specifically testing an admin user correct? would it be worth adding another pytest parameter for the token and creating a non admin token to test this for both?

dcmcand (approved):
Looks great :rocket:

PR #777, opened 2024-03-11
Support hash-only build paths

### LINKED ISSUE #678: [ENH] - Support hash-only build paths
### Feature description

Introduce "Nix-like" hashes that are fixed-size.
We currently have two build-key (build-key is the environment build part of the build-path) types: the legacy one and truncated (with a timestamp).
In this new scheme, it'll be a completely new build-path, which is fixed-size and not depend on user-provided data, like in Nix.
See here https://github.com/conda-incubator/conda-store/pull/652#discussion_r1390512691

### Value and/or benefit

Not running into build_path limits, better control over build_path length (always fixed-size, only depends on conda-store root).

### Anything else?

This is a big change. It requires thinking about collisions. Ideally, we want to share the same envs between multiple people, so this requires thinking about resource sharing. I have details on this in the link above. We might want to hash the lockfile as opposed to CondaSpecification as currently.

The proposed format is:
```
store_directory / hash / <env contents>
```

nkaretnikov:
Fixes #678.

## Description

This pull request:

- adds hash-only build paths (v3)
- updates a test
- updates documentation.

## Pull request checklist

#### Automated testing:

See tests changed in this PR.

#### Manual testing:

- [x] with `~/.conda-store` containing builds created on main, switch to this branch and create a new build: old and new builds should be accessible in the UI (lockfile, archive, and log links should work; marking one of the old builds as default should work)
- [x] same, but also explicitly set `build_key_version = 3` in the config
- [x] same, but also explicitly set `build_key_version = 2` in the config (make sure the previous default works with this PR)
- [x] make sure two exact same specifications can be created in different namespaces and one of them is preserved when the other one is deleted.


conda-store-server/tests/test_actions.py
```diff
@@ -330,6 +330,7 @@ def test_add_lockfile_packages(
         (False, 0),  # invalid
         (False, 1),  # long (legacy)
         (False, 2),  # short (default)
+        (False, 3),  # short (default)
```
nkaretnikov, line 333: TODO: typo in the comment

nkaretnikov, line 333:
```suggestion
        (False, 2),  # shorter hash
        (False, 3),  # hash-only (default)
```

conda-store-server/conda_store_server/orm.py
```diff
@@ -283,6 +290,18 @@ def environment_path(self, conda_store):
         path
 
         """
+        # Uses local import to make sure BuildKey is initialized
+        from conda_store_server import BuildKey
+
+        # This is not used with v3 because the whole point of v3 is to avoid any
+        # dependence on user-provided variable-sized data in the filesystem and
```
nkaretnikov, line 297:
```suggestion
        # dependence on user-provided variable-sized data on the filesystem and
```

docusaurus-docs/conda-store/references/faq.md
```diff
@@ -94,19 +94,50 @@ It consists of:
...
+:::note
+When version 3 is used, `Build.build_path` will not include the namespace name,
+because it's not fixed size, so all builds will be placed right into
+`CondaStore.store_directory`. Additionally, `CondaStore.environment_directory`
+will be completely ignored, so no symlinks will be created, because the
+environment directory format also includes variable-size data (the namespace and
+environment names). This shouldn't impact anything because the generated
+artifacts rely on storage or use the database, and those are not affected by
+these changes to the state directory.
+:::
```
nkaretnikov, line 126: > so no symlinks will be created [...] This shouldn't impact anything [...]

So, I'm actually concerned about this. And I think it *will* impact Nebari/jhub integration because I see the following in `examples/docker/assets/jupyterhub_config.py`, but I have no way to test this right now:

```py
with open(os.path.join(home, '.condarc'), 'w') as f:
    f.write(json.dumps({
        'envs_dirs': [
            '/opt/conda-store/conda-store/default/envs',
            '/opt/conda-store/conda-store/filesystem/envs',
            f'/opt/conda-store/conda-store/{self.user.name}/envs',
        ]
    }))
```

The question is how to proceed. As explained in the doc note above, `environment_directory` is ignored because there's variable-size data, which is incompatible with the purpose of v3. So I see these options:
- Keep v2 the default because it's a good balance (no big hashes and supports env symlinking), v3 will be there for those who don't need to rely on envs_dirs
- Set v2 only in the docker config files bundled with conda-store, which is what Nebari uses.

nkaretnikov:
@jaimergp PTAL. Keeping as a draft for now because this needs to be addressed first: https://github.com/conda-incubator/conda-store/pull/777#discussion_r1519165915. Will do final testing once that is resolved as well.

docusaurus-docs/conda-store/references/faq.md
nkaretnikov: @jaimergp Let's do the final review later. First, I want to agree on the plan here. You can ignore everything else.

nkaretnikov: Update: I've scheduled a meeting with Chuck to discuss how Nebari uses this.

jaimergp: Tag me again after the meeting and I'll review.

nkaretnikov:
@jaimergp This is tested and ready for review, PTAL

docusaurus-docs/conda-store/references/faq.md
```diff
@@ -94,22 +94,60 @@ It consists of:
 3. the id of a build
 4. the environment name.
 
-The version 2 format is now the default. Environments created using the version
-1 format will continue to be accessible in the UI, but new builds will use the
-version 2 format. No changes are needed for existing deployments of conda-store.
+However, version 2 build paths don't solve the problem completely because they
+include user-provided data, like the environment name, and that data can be
+arbitrary large.
```
jaimergp, line 99:
```suggestion
arbitrarily large.
```

docusaurus-docs/conda-store/references/faq.md
```diff
@@ -94,22 +94,60 @@ It consists of:
...
+created, because the environment directory format also includes variable-size
+data (the namespace and environment names).
+
+The lack of symlinks doesn't prevent server artifacts from being generated,
+which are available for download via the UI (lockfiles, archives, etc.), because
+those rely on storage or use the database.
+
+But it does impact conda integration or tools that rely on it, like when
+conda-store is used with JupyterLab as part of a Nebari deployment. Without
+environment symlinks, there'll be no way to tell conda where to look for
```
jaimergp, line 133: Are we creating these symlinks for the end user? Should we? Maybe an example here would be more illustrative.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -46,10 +49,28 @@ def _version2_fmt(build: "Build") -> str:  # noqa: F821
         name = build.specification.name
         return f"{hash}-{timestamp}-{id}-{name}"
 
+    def _version3_fmt(build: "Build") -> str:  # noqa: F821
+        # Adds namespace here to separate builds performed by different users
+        # since all builds are stored in the same directory in v3, see
+        # Build.build_path in orm.py. Additionally, this also hashes the
+        # timestamp and build id just to make collisions very unlikely
```
jaimergp, line 56: Do we want to hash all these components? My first impression is that hashing the lockfile contents would be enough. If that hash is the same for two environments, then why recreate them, right?

If we consider ACLs and concurrent execution from different users, I guess I can see why we add `namespace` to the mix. Security, yadda yadda. However, why do we want to prevent collisions with timestamp and build_id? Can't we just say "oh, this thing already exists in your namespace, there you go" and save a few bytes and inodes?

conda-store-server/conda_store_server/__init__.py
```diff
@@ -46,10 +49,28 @@ def _version2_fmt(build: "Build") -> str:  # noqa: F821
...
+        namespace_name = build.environment.namespace.name
+        specification_hash = build.specification.sha256
```
jaimergp, line 58: Can you remind how is this sha256 computed? From the input textarea, or from the output lockfile?

nkaretnikov: See this code and where it's used in the codebase:

```py
def datastructure_hash(v):
    json_blob = json.dumps(recursive_sort(v))
    return hashlib.sha256(json_blob.encode("utf-8")).hexdigest()
```

It doesn't use the lockfile, it uses the specification (the user-provided text input).

jaimergp: Ah sorry, just saw above.

conda-store-server/conda_store_server/__init__.py
jaimergp: I feel this discussion should involve more members of the team and shouldn't be based on implementation details alone. Pinging @conda-incubator/conda-store for awareness.

nkaretnikov: Okay, we can wait until the next team meeting.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -86,14 +107,28 @@ def get_build_key(cls, build: "Build") -> str:  # noqa: F821
...
         if build_key[cls._version2_hash_size] == "-":  # v2
             return int(parts[2])  # build_id
+        elif "-" not in build_key:  # v3
+            with conda_store.get_db() as db:
+                builds = db.query(orm.Build).all()
+                # Note: this uses a separate for-loop instead of using filter
+                # because the latter doesn't allow for passing lambdas
+                for build in builds:
+                    if cls._version3_fmt(build) == build_key:
+                        return build.id
```
jaimergp, line 131: Are there any performance threats here? Should we return _all_ builds ever or maybe do it in chunks?

jaimergp:
After the discussion in today's weekly:

Some background:

* I am concerned with adding more and more versions to the path settings. That's why I was pushing on "solving this once and for all". But that's just an opinion. If the team is ok with having several path versions, then my asks in items below are not so important.

My asks:

* Build IDs used to identify each attempt in the database shouldn't be so tied to the final path. Instead, we should resolve the lockfile first, and then compute the filesystem path from that output.
* Since that will involve more work, the current compromise of `build-id + timestamp + username` is ok. I do insist that this should just be an internal identifier in the database. The filesystem path could be computed later once input data is available.

As things stand, this is a fixed-length implementation of `v2` right? We could try a hashed-lockfile in a future v4.

nkaretnikov:
@jaimergp and I had a 15 min call about this.

To get this merged:
- [ ] investigate the SQL query perf issue mentioned above
- [ ] add a `$PREFIX/conda-meta/conda-store` file
  - it must **not** have an extension to avoid collisions with other files in that directory
  - the format can be arbitrary JSON and should just provide a way for external (arbitrary) tools to map: `<namespace>/<env>` -> `hash` (if there're multiple generations, these should be stored in a sorted list by creation time)
  - this will provide a path forward for conda-store-ui and its Jupyter integration to use v3 while avoiding filesystem length issues
- update docs to mention this new file and its format.

jaimergp:
I talked with @trallard and we agree that we are good to merge, with the following reservations:

* Mark v3 as experimental with no guarantees of hash stability across conda-store versions (for now).
* Document that this is going to receive iterations where we experiment with different ingredients for the hash. In the future we might want to revisit the lockfile hashing until we arrive to a more robust solution.

How does that sound? Thanks!

nkaretnikov:
@jaimergp PTAL. Will rebase after this is approved.

jaimergp (approved):

PR #786, opened 2024-03-17
Set `CONDA_STORE_DIR` to `platformdirs.user_data_path`

### LINKED ISSUE #648: [ENH] - Consider using OS-specific standards for `conda-store` data
### Feature description

As part of #513, it was suggested to use `~/.conda-store` as the default data directory. This was implemented successfully in https://github.com/conda-incubator/conda-store/pull/639.

I am suggesting we stick to well defined standards for this kind of decisions, instead of polluting the home directory with one more dot-folder. Namely:

- On Linux, follow [XDG Base Directory Specification
](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
- On macOS, it seems that CLI apps still follow XDG standards as Linux. GUI stuff drops things in `~/Library`. See this [SO thread](https://stackoverflow.com/questions/3373948/equivalents-of-xdg-config-home-and-xdg-data-home-on-mac-os-x).
- On Windows, we should probably use [`%LOCALAPPDATA%`](http://msdn.microsoft.com/en-us/library/windows/desktop/dd378457%28v=vs.85%29.aspx) (i.e. non-roaming).

### Value and/or benefit

Following well established standards in each Operating System allow us to be better citizens, and navigate better known spaces (e.g. permissions, lack of user profiles in some Windows installations).

A nice side effect, and one of the reasons these standards exist, is to avoid polluting the user profile with dot-directories.

### Anything else?

Comes from https://github.com/conda-incubator/conda-store/pull/639#issuecomment-1787555517

nkaretnikov:
Fixes #648.

## Description

This pull request:

- sets `CONDA_STORE_DIR` to `platformdirs.user_data_path`
- updates docs
- adds a test.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/__init__.py
+++ b/conda-store-server/conda_store_server/__init__.py
@@ -1,13 +1,13 @@
 import datetime
 import typing
 
-from pathlib import Path
+import platformdirs
 
 
 __version__ = "2024.3.1"
 
 
-CONDA_STORE_DIR = Path.home() / ".conda-store"
+CONDA_STORE_DIR = platformdirs.user_data_path(appname="conda-store")
 
 
 class BuildKey:
--- a/conda-store-server/environment.yaml
+++ b/conda-store-server/environment.yaml
@@ -18,7 +18,7 @@ dependencies:
   - redis-py
   - sqlalchemy<=1.4.47
   - alembic
-  - platformdirs
+  - platformdirs >=4.0,<5.0a0
   - psycopg2
   - pymysql
   - requests
--- a/conda-store-server/environment.yaml
+++ b/conda-store-server/environment.yaml
@@ -18,6 +18,7 @@ dependencies:
   - redis-py
   - sqlalchemy<=1.4.47
   - alembic
+  - platformdirs
   - psycopg2
   - pymysql
   - requests
--- a/docusaurus-docs/conda-store/references/configuration-options.md
+++ b/docusaurus-docs/conda-store/references/configuration-options.md
@@ -30,7 +30,7 @@ The `CONDA_STORE_DIR` Python variable specifies the conda-store data directory,
 which is used by some of the configuration options mentioned below, like
 `CondaStore.store_directory` and `LocalStorage.storage_path`. This variable
 relies on the [`platformdirs`][platformdirs] library to select the recommended user data
-location on each platform:
+location on each platform. On most systems, this will default to:
 
 - Linux: `/home/<USER>/.local/share/conda-store`
 - Windows: `C:\Users\<USER>\AppData\Local\conda-store\conda-store`
--- a/docusaurus-docs/conda-store/references/configuration-options.md
+++ b/docusaurus-docs/conda-store/references/configuration-options.md
@@ -44,7 +44,7 @@ should correspond to the following environment variables:
 - macOS: no dedicated environment variable.
 
 Note that whether these environment variables are actually used by
-[platformdirs] is up to the library authors and can be changed at any time.
+[`platformdirs`][platformdirs] is up to the library authors and can be changed at any time.
 Please use the conda-store configuration options mentioned below instead.
 
 [platformdirs]: https://github.com/platformdirs/platformdirs
--- a/docusaurus-docs/conda-store/references/configuration-options.md
+++ b/docusaurus-docs/conda-store/references/configuration-options.md
@@ -29,7 +29,7 @@ conda-store-worker --config <path-to-conda-store-config.py>
 The `CONDA_STORE_DIR` Python variable specifies the conda-store data directory,
 which is used by some of the configuration options mentioned below, like
 `CondaStore.store_directory` and `LocalStorage.storage_path`. This variable
-relies on the [platformdirs] library to select the recommended user data
+relies on the [`platformdirs`][platformdirs] library to select the recommended user data
 location on each platform:
 
 - Linux: `/home/<USER>/.local/share/conda-store`
--- a/docusaurus-docs/conda-store/references/configuration-options.md
+++ b/docusaurus-docs/conda-store/references/configuration-options.md
@@ -4,6 +4,8 @@ description: conda-store configuration options
 
 # Configuration options
 
+### Traitlets
+
 :::warning
 This page is in active development, content may be inaccurate and incomplete.
 :::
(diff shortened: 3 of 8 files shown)


nkaretnikov:
@jaimergp This is ready and tested, PTAL. CI failures are unrelated.

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -22,7 +24,30 @@ conda-store-server --config <path-to-conda-store-config.py>
 conda-store-worker --config <path-to-conda-store-config.py>
 ```
 
-Below we outline the options for conda-store.
+### Data directory
+
+The `CONDA_STORE_DIR` Python variable specifies the conda-store data directory,
+which is used by some of the configuration options mentioned below, like
+`CondaStore.store_directory` and `LocalStorage.storage_path`. This variable
+relies on the [platformdirs] library to select the recommended user data
```
jaimergp, line 32:
```suggestion
relies on the [`platformdirs`][platformdirs] library to select the recommended user data
```

I think this is the right syntax, isn't it?

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -22,7 +24,30 @@ conda-store-server --config <path-to-conda-store-config.py>
...
+location on each platform:
```
jaimergp, line 33:
```suggestion
location on each platform. On most systems, this will default to:
```

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -22,7 +24,30 @@ conda-store-server --config <path-to-conda-store-config.py>
...
+
+- Linux: `/home/<USER>/.local/share/conda-store`
+- Windows: `C:\Users\<USER>\AppData\Local\conda-store\conda-store`
+- macOS: `/Users/<USER>/Library/Application Support/conda-store`.
+
+The platform user data directory prefix, which is before `conda-store` above,
```
jaimergp, line 39:
```suggestion
The platform user data directory prefix, which is the parent of the `conda-store` directory above,
```

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -22,7 +24,30 @@ conda-store-server --config <path-to-conda-store-config.py>
...
+should correspond to the following environment variables:
+
+- Linux: `$XDG_DATA_HOME`
+- Windows: `%LOCALAPPDATA%`
+- macOS: no dedicated environment variable.
+
+Note that whether these environment variables are actually used by
+[platformdirs] is up to the library authors and can be changed at any time.
```
jaimergp, line 47:
```suggestion
[`platformdirs`][platformdirs] is up to the library authors and can be changed at any time.
```

conda-store-server/environment.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   - redis-py
   - sqlalchemy<=1.4.47
   - alembic
+  - platformdirs
```
jaimergp, line 21: I wonder if we should pin this to at least current major. The library has [received some breaking updates in the past](https://platformdirs.readthedocs.io/en/latest/changelog.html) that would have affected our configuration.

```suggestion
  - platformdirs >=4.0,<5.0a0
```

We can then unpin as needed if there are major releases that do not break.

jaimergp:
This looks good. Let's discuss the pinning item briefly and then we can make a decision.

nkaretnikov:
@jaimergp Updated, PTAL.

The only concern I have is that the library version bounds might not be strict enough. The authors introduced a breaking change between 2.5.4 and 2.6.0. But this can also be updated later if we run into problems.

jaimergp:
> The only concern I have is that the library version bounds might not be strict enough. The authors introduced a breaking change between 2.5.4 and 2.6.0. But this can also be updated later if we run into problems.

Yea, I saw that, but hopefully they have learnt from that oversight. Also [this kind of comment](https://github.com/platformdirs/platformdirs/issues/265#issuecomment-1937111568) makes me believe they are being more careful now and not too keen on breaking anything else.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -1,11 +1,12 @@
...
 import typing
-from pathlib import Path
+
+import platformdirs
 
 __version__ = "2024.3.1"
 
 
-CONDA_STORE_DIR = Path.home() / ".conda-store"
+CONDA_STORE_DIR = platformdirs.user_data_path(appname="conda-store")
```
jaimergp, line 9:
```suggestion
CONDA_STORE_DIR = Path(platformdirs.user_data_path(appname="conda-store"))
```

Just realized `platformdirs` returns `str` objects, so let's keep the same type around just in case.

conda-store-server/conda_store_server/__init__.py
```diff
@@ -1,11 +1,12 @@
 import datetime
 import typing
-from pathlib import Path
```
jaimergp, line 3: Bring this back for the change below 😬

jaimergp (changes requested):
Just one more thing.

Also, since this adds a new dependency, what's the process for letting the release manager know there are new deps to be added to the conda recipe? Will they look at a specific file? Maybe [pyproject.toml](https://github.com/conda-incubator/conda-store/blob/fe6b45a7f722ffc5f9e9b4f3669724441032b263/conda-store-server/pyproject.toml#L28-L48) also needs adjusting?

conda-store-server/conda_store_server/__init__.py
nkaretnikov: No change is necessary here, the `*_path` functions already return a `Path`.

conda-store-server/conda_store_server/__init__.py
nkaretnikov: See above.

nkaretnikov:
@jaimergp I've added `platformdirs` to `pyproject.toml`. I've never released conda-store, let's ask Tania.

@trallard How do new dependencies affect the release process? Anything to do besides updating `pyproject.toml` and yaml files? I've read `release.md` -- just making sure.

trallard:
Only the `pyprojecr.toml` needs updating

jaimergp (approved):

PR #790, opened 2024-03-20
Make conda_flags configurable

patrix58:
Currently the `conda_flags` parameter in https://github.com/conda-incubator/conda-store/blob/2024.3.1/conda-store-server/conda_store_server/action/generate_lockfile.py#L20 but this is not configurable.

This PR includes changes to have the `conda_flags` parameter configurable from the config py script.

## Description

We are using conda-store within nebari in one of our projects. We have a reason why we don't need strict channel priority while building the environments. The default is still `--strict-channel-priority` to be backwards compatible, but it makes the option to the user to configure different conda flags e.g. `--no-channel-priority`.

## Pull request checklist

- running tests locally
- was deployed and tested with nebari on a k8s cloud

## How to test

Add e.g. `c.CondaStore.conda_flags = "--no-channel-priority"` in the config py script.

### DIFF
--- a/conda-store-server/conda_store_server/app.py
+++ b/conda-store-server/conda_store_server/app.py
@@ -147,6 +147,12 @@ class CondaStore(LoggingConfigurable):
         config=True,
     )
 
+    conda_flags = Unicode(
+        "--strict-channel-priority",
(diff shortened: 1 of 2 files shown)


nkaretnikov (approved):
@patrix58 Thanks, LGTM!

PR #792, opened 2024-03-20
DEV - Clean and update dependencies

trallard:
Thanks to @pavithraes we realised that now runtime dependencies for conda-store(-server) are not up to date nor included in distributed versions of the package. 

This is in part due to having bloated `environment.yaml` files that contain both dev and runtime dependencies (that should only de specified in the `pyproject.toml`). Since these are both used in local development and CI missing dependencies have gone unnoticed. This PR aims to clean the dependencies and enviroments.

## Description

This pull request:

- Adds conda-store(-server) runtime dependencies to the corresponding `pyproject.toml`
- Keeps dev only dependencies in `environment.yaml`
- Bumps GH actions and pre-commit versions

## Pull request checklist


conda-store/pyproject.toml
```diff
@@ -13,8 +13,10 @@ keywords = ["conda"]
 authors = [
   { name = "Christopher Ostrouchov", email = "chris.ostrouchov@gmail.com" },
 ]
-maintainers = [{ name = "Tania Allard", email = "trallard@bitsandchips.me" }]
-# TODO: update maintainers
+maintainers = [{ name = "Tania Allard", email = "trallard@bitsandchips.me" },
+    {name= "Pavithra Eswaramoorthy", email = " pavithraes@outlook.com"},
```
peytondmurray, line 17:
```suggestion
    {name= "Pavithra Eswaramoorthy", email = "pavithraes@outlook.com"},
```
(not sure if this matters)

conda-store/pyproject.toml
```diff
@@ -65,10 +66,9 @@ conda-store = "conda_store.__main__:main"
 
 [tool.black]
 line-length = 88
-target-version = ['py38', 'py39', 'py310']
+target-version = ['py38', 'py39', 'py310', 'py311']
```
peytondmurray, line 69: I guess 3.12 probably makes sense to add here as well?

trallard: +1 on adding it. The thing is that black "tries" to infer the versions, so I prefer explicitly having these per

> You should include all versions that your code supports. If you support Python 3.8 through 3.11, you should write:...

pavithraes (approved):
Thank you, this looks great!

I took a look at the docs updates. I also tried local+dev install conda-store against this branch, and all seemed good.

The suggestions are typo-fixes, so I'll approve.

conda-store-server/environment-dev.yaml
```diff
@@ -7,57 +7,22 @@ dependencies:
...
-  - celery
   - flower
-  - redis-py
-  - sqlalchemy<=1.4.47
-  - psycopg2
-  - pymysql
-  - requests
-  - uvicorn
-  - fastapi
-  - pydantic < 2.0
```
nkaretnikov, line 26: Same here. The dependencies need to be reviewed carefully. For instance, this file is used by integration tests to run the test code. Currently, the integration tests don't run at all - they fail during the collect step because a different version of pydantic is being used (instead of the <2 version specified above).

From the test output:

```
==================================== ERRORS ====================================
______________________ ERROR collecting tests/test_api.py ______________________
../tests/test_api.py:25: in <module>
    from conda_store_server import schema
conda_store_server/schema.py:50: in <module>
    class AuthenticationToken(BaseModel):
conda_store_server/schema.py:55: in AuthenticationToken
    role_bindings: Dict[constr(regex=ARN_ALLOWED), List[str]] = {}
E   TypeError: constr() got an unexpected keyword argument 'regex'
=========================== short test summary info ============================
ERROR ../tests/test_api.py - TypeError: constr() got an unexpected keyword argument 'regex'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.34s ===============================
```

The workflow code:

```yaml
  integration-test-conda-store-server:
...
        with:
          environment-file: conda-store-server/environment-dev.yaml
```

I suspect there will be other failures once you add pydantic back because the tests import `conda_store_server`.

conda-store-server/pyproject.toml
```diff
@@ -25,55 +29,77 @@ classifiers = [
...
+  "uvicorn",
   "yarl",
-  "filelock",
-  "itsdangerous",
-  "jinja2",
-  "python-multipart",
-  "python-docker"
-  # conda (which should not be included here)
+  # installer
+  "constructor",
```
nkaretnikov, line 65: constructor is a runtime dependency and needs to be in the environment for it to work, which is why you're seeing these unit test failures:

```
tests/test_actions.py::test_generate_constructor_installer
```

nkaretnikov (changes requested):
@trallard Thanks for your contribution! I think we need to rethink the approach here.

To summarize:
- This PR makes too many unrelated changes: whitespace, documentation, version bumps, formatting, which makes it harder to review
- I suggest keeping only absolutely essential things here (for the purposes of adding missing dependencies to the pyproject file): env yaml files, workflows, pyproject update, and splitting out everything else into one or more self-contained PRs
- Moreover, I think the current approach needs rethinking altogether. The initial goal of this PR is to add missing dependencies to the pyproject file. But as is, it tries to change the existing testing infra, which relies heavily on env yaml files in several places, which is also what devs use to test and run the code locally (as opposed to using hatch to build via pyproject)
- Changing too many things at once is error-prone. So I suggest this instead: add missing dependencies to the pyproject file as needed, add _a completely separate workflow that would use the pyproject file_ and would run unit tests against conda-store installed via hatch, keep everything else as is. I understand this is not ideal because of duplication. But this way we keep the good known baseline around until we have a working solution with hatch. After a while, once we're comfortable with new testing infra, we can retire the old one.

I'm happy to have an offline chat about this or take over parts of this PR if needed.

.github/workflows/tests.yaml
```diff
@@ -54,15 +48,16 @@ jobs:
 
       - name: "Linting Checks 🧹"
         run: |
-          hatch env run -e dev lint
+          hatch env run -e lint lint
```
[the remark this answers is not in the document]
trallard: One refers to the environment name (lint) and the other the script (lint)

conda-store-server/environment.yaml
```diff
@@ -6,33 +6,5 @@ dependencies:
...
-  - yarl
-  - pyjwt
-  - filelock
-  - itsdangerous
-  - jinja2
-  - python-multipart
-  # artifact storage
-  - minio
-  # installer
-  - constructor
```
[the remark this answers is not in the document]
trallard: `constructor` is a runtime dependency of conda-store itself so it should be already installed.

We do not need to keep double installing dependencies - runtime and dev. 
This is the whole purpose of the PR separating runtime and dev concerns as currently they are all mushed up (runtime dependencies are in dev so when they were not added to the pyproject.toml it was always obscured as they were installed with the dev env file)

conda-store-server/pyproject.toml
trallard: Exactly is a runtime dependency not a dev dependency. So it does not need to be in the environment.

trallard:
Will close this as all the additions here have been split across other PRs

PR #796, opened 2024-03-25
🔧 Update .pre-commit-config

trallard:
Split from #792 

## Description

- **:wrench: Update .pre-commit-config**
- **:wrench: Update pre-commit config in pyproject.toml**
- **Update docs and actions to use lint**

## Pull request checklist

## How to test

- Run pre-commit locally with `pre-commit run all-files`
- Run with hatch with `hatch env run -e lint lint`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,34 +7,28 @@
 #
 # - Register git hooks: pre-commit install --install-hooks
 # - Run on all files:   pre-commit run --all-files
+ci:
+  autoupdate_schedule: monthly
+  autofix_commit_msg: |
+    [pre-commit.ci] Apply automatic pre-commit fixes
 
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 24.3.0
     hooks:
       - id: black
         exclude: "examples|tests/assets"
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
     # Ruff version.
-    rev: "v0.0.289"
+    rev: "v0.3.4"
     hooks:
       - id: ruff
         exclude: "examples|tests/assets"
         args: ["--fix"]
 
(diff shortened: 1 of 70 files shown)


peytondmurray (approved):
Looks great - I especially like that pre-commit.ci will autoupdate. :rocket: 

Something weird is up with one of the tests, though it doesn't look like anything related. Is this flaky?

trallard:
Seems like one of those network issues. I tried and re run the workflow multiple times with no avail and  there is nothing in this PR that would affect tests. 

Since it's a parallel test one will tag @nkaretnikov

.pre-commit-config.yaml
```diff
@@ -7,34 +7,28 @@
...
-  - repo: https://github.com/pycqa/isort
-    rev: 5.12.0
-    hooks:
-      - id: isort
-        name: isort
-        additional_dependencies: [toml]
-        files: \.py$
-        args: ["--profile", "black"]
-        exclude: "conda-store-server/conda_store_server/action/__init__.py"
-
```
nkaretnikov, line 35: Why is this removed? Is it assumed that pre-commit will handle this?

I reordered some imports in `conda-store-server/conda_store_server/api.py`, but `hatch env run -e lint lint` didn't fix that.

nkaretnikov (approved):
@trallard That was a flaky test. Approving to unblock, but see the inline question about import sorting.

.pre-commit-config.yaml
trallard: I had swapped to the `isort` in `ruff` but it makes little difference to be honest and if it makes it easier/less confusing to keep the sort pre-commit it is ok

trallard:
Thanks @nkaretnikov - those flaky tests are an absolute pain. 

Reverted to vanilla `isort` btw

PR #797, opened 2024-03-25
Add user journey test to check that logs can be found for failed builds

### LINKED ISSUE #672: Integration Test: access logs from failed buld
### Context

Part of https://github.com/conda-incubator/conda-store/issues/662

#### User Story

Persona: I'm a software engineer who manages my own envs
Need: I have an env that failed to build, I need to investigate the logs to determine the cause of the failure.

#### Test
TODO: @dcmcand work out api calls and setup needed
##### Setup
1.
2.
##### API Calls
1.
2.

### Value and/or benefit

This will ensure that new releases of conda-store will not break critical user journeys.

### Anything else?

Note: This integration test only covers conda-store-server and this workflow should be tested at the api level. UI logic will be tested in the [conda-store-ui](https://github.com/conda-incubator/conda-store-ui) repo.

peytondmurray:
Fixes #672.

## Description

This PR implements a user journey test that checks that a user can access logs for a failed build.

This pull request:

- Adds a simple broken environment spec
- Implements a `get_logs` function for the user journey API utils
- Adds a test for the user journey mentioned above

## Pull request checklist

## How to test

Run the conda-store server:

```bash
docker compose up
```

Then run the test:

```bash
cd conda-store-server
pytest tests/user_journeys/test_user_journeys.py::test_failed_build_logs
```

### DIFF
--- a/conda-store-server/tests/user_journeys/test_user_journeys.py
+++ b/conda-store-server/tests/user_journeys/test_user_journeys.py
@@ -128,3 +128,20 @@ def test_admin_delete_environment(base_url: str):
 
     assert len(api.list_environments(namespace).json()["data"]) == 0
     api.delete_namespace(namespace)
+
+
+@pytest.mark.user_journey
+def test_failed_build_logs(base_url: str):
+    """Test that a user can access logs for a failed build."""
+    api = utils.API(base_url=base_url)
+    namespace = "default"
+    build_request = api.create_environment(
+        namespace,
+        "tests/user_journeys/test_data/broken_environment.yaml",
+    ).json()
+
+    assert build_request["data"]["status"] == "FAILED"
+    assert (
+        "invalidpackagenamefaasdfagksdjfhgaskdf"
+        in api.get_logs(build_request["data"]["id"]).text
+    )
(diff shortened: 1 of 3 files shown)


conda-store-server/tests/user_journeys/test_user_journeys.py
```diff
@@ -100,3 +100,18 @@ def test_user_login_and_create_shared_environment(
...
+    api = utils.API(base_url=base_url)
+    namespace = "default"
+    build_request = api.create_environment(
+        namespace,
+        "tests/user_journeys/test_data/broken_environment.yaml",
+    ).json()
+
+    assert build_request["data"]["status"] == "FAILED"
+    logs = api.get_logs(build_request["data"]["id"]).text
+    assert "could not be installed" in logs
```
dcmcand, line 117: I generally don't like testing for specific strings since that feels a bit brittle. Is there another way you can think of to validate the logs? If not I will still approve, just a reservation about this method.

dcmcand (approved):

conda-store-server/tests/user_journeys/test_user_journeys.py
peytondmurray: I'll check instead that the fake package name is in the logs. Thanks for the feedback!

peytondmurray:
Test failures appear to be unrelated.

PR #798, opened 2024-03-25
MAINT - Add conda-store dependencies to pyproject.toml

trallard:
Split from #792 

## Description

- **BumpJLab** - now conda-store-ui is JLab >= 4.0 compatible
- **Update dev and runtime dependencies - server**
- **Update runtime dependencies - core**
- **Update dev dependencies - core**

## Pull request checklist

## How to test

This can be tested by building the packages with `hatch build` both for `conda-store` and `conda-store-server`

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -54,7 +54,7 @@ jobs:
 
       - name: "Linting Checks 🧹"
         run: |
-          hatch env run -e dev lint
+          hatch env run -e lint lint
 
       - name: "Build package 📦"
         run: |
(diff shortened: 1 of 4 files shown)


peytondmurray (approved):
I can't comment on the individual dependencies, but I like that more python deps are moving into `pyproject.toml`.

One thing that might be helpful for future maintainers would be a comment at the top of the `environment.yaml` files that are in various places in this repository with a brief note about what dependencies are okay to add to that file and which need to be added to `pyproject.toml`. Something like

> This environment file is for \<whatever purpose\>; please put python project dependencies in pyproject.toml rather than here

or similar would be useful.

trallard:
Thanks! Yes I plan to add such notes and docs as I wrap up the changes that were in that original PR.

Issue #803, opened 2024-03-28
[META] Test failures on CI


peytondmurray:
### Context

This issue is to collect currently failing CI test reports.

- [`test_create_specification_parallel_auth`](https://github.com/conda-incubator/conda-store/actions/runs/8472738567/job/23215487457?pr=797)
- [`test_install_specification`](https://github.com/conda-incubator/conda-store/actions/runs/8472738567/job/23215487457?pr=797)
- [`test_conda_store_register_environment_workflow`](https://github.com/conda-incubator/conda-store/actions/runs/8472160234/job/23213684751?pr=797)
- [`test_api_cancel_build_auth`](https://github.com/conda-incubator/conda-store/actions/runs/8472160234/job/23213684079?pr=797)
- [`test_generate_constructor_installer[simple_specification]`](https://github.com/conda-incubator/conda-store/actions/runs/8425949310/job/23073173256?pr=797) - [See issue](https://github.com/conda-incubator/conda-store/issues/799)
- [`test_solve_lockfile_multiple_platforms[simple_specification]`](https://github.com/conda-incubator/conda-store/actions/runs/8288925479/job/22684452762?pr=776#step:7:70) - [See issue](https://github.com/conda-incubator/conda-store/issues/783)

### Value and/or benefit

Increasing confidence in test suite will allow us to move more quickly.

nkaretnikov:
I've added a PR that helps mitigate issues caused by flaky tests: https://github.com/conda-incubator/conda-store/pull/813

trallard:
Not a fail, but I have started to notice a warning about stuff that recently changed so we might want to update https://github.com/conda-incubator/conda-store/actions/runs/9490164999/job/26153052479#step:7:78

peytondmurray:
Will go back and look at the current state of the unit tests.

PR #813, opened 2024-04-18
Run tests as separate jobs

nkaretnikov:
## Description

This pull request:
- splits up tests into multiple workflows
- splits up integration tests into multiple jobs
- fixes some flaky tests.

## Pull request checklist

## Additional information

- Now, if integration tests fail, you don't need to wait for unit tests to finish before restarting them, and vice versa.
- Integration tests are split into multiple jobs, so it's easier to restart each separately (there's less chance of a job failing because it involves less tests now).

Also see #803.


tests/test_api.py
```diff
@@ -275,14 +275,24 @@ def test_api_get_build_one_unauth(testclient):
...
+        assert r.data.specification.name == "python-flask-env"
+        if r.data.status in [
+            schema.BuildStatus.QUEUED.value,
+            schema.BuildStatus.BUILDING.value,
+        ]:
+            time.sleep(10)
+            continue
+        assert r.data.status == schema.BuildStatus.COMPLETED.value
+        success = True
+    assert success
```
nkaretnikov, line 295: This and the other test were modified because they failed often, due to checking the build before it's available.

.github/workflows/test_build_docker_image.yaml
```diff
@@ -0,0 +1,62 @@
...
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+
+# ensuring only one instance is running at a given time
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
nkaretnikov, line 16: This now has to include the workflow. Otherwise, it would not be possible to run multiple workflows concurrently.

.github/workflows/test_conda_store_server_integration.yaml
```diff
@@ -0,0 +1,82 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  integration-test-conda-store-server:
+    name: "integration-test conda-store-server"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        test-type: ["playwright", "integration", "user-journey"]
```
nkaretnikov, line 25: This now runs each test type as a separate job, so you can restart them individually if they fail.

.github/workflows/test_conda_store_server_integration.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+      - name: "Deploy docker-compose"
+        run: |
+          docker-compose up -d
+          docker ps
+
+          wait-for-it localhost:5432 # postgresql
+          wait-for-it localhost:9000 # minio
+          wait-for-it localhost:8080 # conda-store-server
+          sleep 10 # give the server more time to initialize
```
nkaretnikov, line 53: This gives time for envs that come with the server to initialize (default, filesystem).

tests/test_api.py
```diff
@@ -275,14 +275,25 @@ def test_api_get_build_one_unauth(testclient):
...
+        if r.data.status in [
+            schema.BuildStatus.QUEUED.value,
+            schema.BuildStatus.BUILDING.value,
+        ]:
+            time.sleep(10)
+            continue
+        assert r.data.status == schema.BuildStatus.COMPLETED.value
+        success = True
+        break
+    assert success
```
peytondmurray, line 296: Okay, let's hope we don't need 50 retries here. A slightly cleaner way of doing this is to just `return` when the build is completed rather than the success/break/assert:

```python
def test_api_get_build_one_auth(testclient):
    testclient.login()
    for _ in range(50):
        response = testclient.get("api/v1/build/1")
        response.raise_for_status()
                r = schema.APIGetBuild.parse_obj(response.json())
        assert r.status == schema.APIStatus.OK
        assert r.data.id == 1
        assert r.data.specification.name == "python-flask-env"
        if r.data.status in [
            schema.BuildStatus.QUEUED.value,
            schema.BuildStatus.BUILDING.value,
        ]:
            time.sleep(10)
            continue
        assert r.data.status == schema.BuildStatus.COMPLETED.value
        return
```

but it's up to you if you think it worth the change.

.github/workflows/test_conda_store.yaml
```diff
@@ -0,0 +1,84 @@
...
+          wait-for-it localhost:8080 # conda-store-server
+
+      - name: "Install conda-store for tests 📦"
+        run: |
+          pip install .
+
+      - name: "Run basic tests -  not authenticated"
+        run: |
+          sleep 20
+          ./tests/unauthenticated-tests.sh
```
peytondmurray, line 67: I see this was in the workflow that was removed too, so I'm fine with it. Do you have any idea though why do we need this wait? We have `wait-for-it` running above that should be waiting for the `conda-store-server`...?

.github/workflows/test_conda_store_server_integration.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - name: "Run integration tests ✅"
+        run: |
+          export PYTHONPATH=$PYTHONPATH:$PWD
+          pytest ../tests/test_api.py ../tests/test_metrics.py
+        if: matrix.test-type == 'integration'
+
+      - name: "Run user journey tests ✅"
+        run: |
+          pytest -m "user_journey"
+        if: matrix.test-type == 'user-journey'
```
peytondmurray, line 77: As a future goal it might be nice to let `pytest` handle this; we could just pass in the `test-type` from the matrix above:

```
- name: Run ${{ test-type }} tests:
  run: |
    pytest -m "${{ test-type }}"
```

but we'd need the different tests to be marked properly for this to work.

.github/workflows/test_conda_store_server_unit.yaml
```diff
@@ -0,0 +1,79 @@
...
+        uses: conda-incubator/setup-miniconda@v2
+        with:
+            environment-file: ${{ matrix.environment-file }}
+            miniforge-version: latest
+
+      # This fixes a "DLL not found" issue importing ctypes from the hatch env
+      - name: "Reinstall Python 3.10 on Windows runner"
+        uses: nick-fields/retry@v2.8.3
+        with:
+          timeout_minutes: 9999
```
peytondmurray, line 51: I know this isn't your change, but does it make sense to put something reasonable here?

peytondmurray (approved):
Just a few comments, otherwise this looks like it will really help the CI be more useful. Thanks for this :rocket:

tests/test_api.py
nkaretnikov: The suggested change is problematic: if the build is stuck in the queued/building state for 50 iterations, the loop will exit silently and the test will succeed. I want to detect this as well, which is why I check for success outside of the loop.

.github/workflows/test_conda_store_server_unit.yaml
nkaretnikov: I remember when this was added and had concerns as well. But I think it's fine. The goal of this timeout is to make sure it never hits before `max_attempts: 6` below executes, so it's less flaky this way. It's been working fine, so I see no reason to change this.

.github/workflows/test_conda_store.yaml
nkaretnikov: Nope, I have no context on this one. But I suspect it's the same reason why I added another sleep myself -- to make sure envs are initialized properly after the server has started.

.github/workflows/test_conda_store_server_integration.yaml
nkaretnikov: `wait-for-it` just checks whether a URL is accessible. But some additional work is needed on the server side to initialize the bundled envs (default, filesystem), which are used for testing. So `wait-for-it` alone is not sufficient.

.github/workflows/test_conda_store_server_integration.yaml
nkaretnikov: Yep, something to think about for later. I was hesitant to introduce too many changes since I wanted to keep this well-scoped.

.github/workflows/test_build_docker_image.yaml
trallard, line 1: We should perhaps rename it to `build_docker.yaml` as it is not testing the Docker image per se

.github/workflows/test_build_docker_image.yaml
```diff
@@ -0,0 +1,62 @@
+name: "Test build Docker image"
```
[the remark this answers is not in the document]
nkaretnikov: Had a similar idea as well, but decided against it to avoid confusion. We already have a job called `build_and_push_docker_image` that's part of `release.yaml`. This job was part of `tests.yaml`, so I prefixed it with "Test" to see which is which. AFAICT, it also doesn't push anything because of `push: false`, so looks like the intention here is to just test whether a docker container can be built.

In the interest of time, I'm going to merge this as is since it's already approved, but this can always be resolved in followup PRs.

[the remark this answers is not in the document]
trallard: But it is not testing the docker image just building it the fact that we have a step/job named build and push is inconsequential to having a workflow named build or build and push.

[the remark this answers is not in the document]
nkaretnikov: Yeah, I think a better name might be "Build Docker image", but we already have a file called `build_docker_image.yaml` with the name "Publish Dev Docker Images".

I think my preferred solution would be making all these consistent, so that the name matches the filename and there are no clashes between files. Something like this (untested):
- "Publish dev Docker image" -> publish_dev_docker_image.yaml (renamed from "Publish Dev Docker Images")
- "Build Docker image" -> build_docker_image.yaml (the workflow discussed in this thread)

@peytondmurray Hey Peyton, I'm currently on PTO, would you be comfortable taking over this in a separate PR? The task would be coming up with a consistent scheme for names and filenames of docker workflows.

PR #820, opened 2024-05-06
Privatize internal conda-store API

### LINKED ISSUE #723: Privatize `conda-store` and `conda-store-server` internals
### Context

This project is expected only to be used through the REST API (provided by `conda-store-server`) or CLI (provided by `conda-store`). Currently, we have several Python functions implicitly "public facing", even though we don't expect users to use conda-store through a "Python API". 

We can consider being explicit about the public API and moving everything else to a private space/directory (conventionally, prefixed with an underscore). 

### Value and/or benefit

Cleaner codebase and avoid any unintended usage/maintenance over the long-term.

### Anything else?

We need to make sure this change is aligned with the backwards compatibility policy.

peytondmurray:
Closes #723.

## Description

This PR privatizes most of conda-store and conda-store-server in preparation for implementing an upcoming backwards compatibility policy.

This pull request:

- Moves everything into `_internal` subpackages, except for a select few outward-facing APIs.

## Pull request checklist

## How to test

No special testing requirements because new behavior has not been introduced; standard CI tests are sufficient.


dcmcand (approved):
@peytondmurray, I am assuming that every function you left in the pubblic namespace either corresponds to a public rest api route or is used by the conda-store-ui, is that correct. I think this PR is great, I just want to double check that there aren't any little helper functions or anything that we could privatize. They wouldn't need to move to the `_internal` namespace imo, just have prepend them with a `_`. I am approving this based on the above assumption, but you might want to take another look through to see if any of the functions that are public could be privatized.

PR #823, opened 2024-05-13
Add `conda config` warning docs/message; add duration log for actions

### LINKED ISSUE #810: [ENH] - Make logs more user-friendly
### Feature description

This has two subtasks:

- [ ] Explain how to interpret the output of `conda config` and the strict channel priority. Users only look at the config output and overlook `CONDA_FLAGS=--strict-channel-priority`. The latter is what actually determines the channel priority in conda-store. Add a very visible message before the config output is printed in the logs explaining this.

- [ ] Print action end time and time delta. This probably should be done with a context manager where `ActionContext` is defined, so it automatically applies to all actions.

### Value and/or benefit

Improved UX.

### Anything else?

_No response_

peytondmurray:
Fixes #810.

## Description

This PR adds:

1. A warning message about environment variables used by `conda-store` to override user conda configuration, which should help clear up user uncertainty about the output of `conda config`.
2. Some information in the documentation about the above :point_up:
3. An info-level log message is now emitted displaying the amount of time each action takes.

## Pull request checklist

- [x] Did you test this change locally? **Yes, but not all tests pass locally on `main`. The ones I expected to pass locally have passed.**

## Notes

- To see that this works locally, run any action. Alternatively, run `pytest conda_store_server/test/test_actions.py::test_action_decorator` - that test specifically checks for the run time information output to stdout that this PR implements.


gabalafou (approved):
Did not test but the code looks good to me. 

The log message may be ambiguous, not sure, but at any rate it does the job of letting the user know that something is up with conda config, which is the main point.

conda-store-server/conda_store_server/_internal/action/generate_lockfile.py
```diff
@@ -27,6 +27,13 @@ def action_solve_lockfile(
     with environment_filename.open("w") as f:
         json.dump(specification.dict(), f)
 
+    context.log.info(
+        "Note that the output of `conda config --show` displayed below only reflects "
+        "settings in the conda configuration file, which might be overridden by "
+        "variables required to be set by conda-store via the environment. Overridden "
+        f"settings: CONDA_FLAGS={conda_flags}"
```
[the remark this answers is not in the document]
peytondmurray: Yes, these were the three things I was trying to convey. As far as I know these are all of the overridden settings as well.

docusaurus-docs/conda-store/explanations/conda-concepts.md
```diff
@@ -80,3 +80,15 @@ could be updated the next minute the same solve for the same
...
+
+Because conda-store needs to configure some parts of conda without modifying
+the user's conda configuration file, internally conda-store sets some conda
+configuration variables using environment variables. The impact of this is that
+if a user tries to print their conda configuration with `conda config`, some of
+the configuration settings displayed by that command will not reflect the values
+that are actually used by conda-store. In particular, `conda-store` internally
+sets `CONDA_FLAGS=--strict-channel-priority`, overriding the channel priority in
+the conda configuration file. Please keep this in mind when using `conda config`
+to inspect your conda configuration.
```
kcpevey, line 94: This is also (and perhaps more frequently) an issue when you view the logs for conda-store since the config is printed there as well.

kcpevey, line 94:
```suggestion
to inspect your conda configuration and when viewing the build logs.
```

kcpevey (approved):
I just had a minor suggestion, otherwise, this looks great! Thanks @peytondmurray !

PR #801, opened 2024-03-26
DEV - Update dev workflows and CI

trallard:
Split from #792 

## Description

- Remove OS-specific `environment.yaml` files as no longer needed
- Update env files - server
- Update environment file - core
- Add note about updating dependencies
- Update CI - remove unnecessary `yaml` files

## Pull request checklist

- [x] Did you test this change locally?
- [x] Did you update the documentation (if required)?
- [ ] Did you add/update relevant tests for this change (if required)?

## Additional information

With this PR, at most, contributors/maintainers will need to regenerate/update their local conda environment. 

## How to test

💻 Locally:

1. Regenerate your local `conda` development environment (should use `conda-store-server/environment-dev.yaml`
2. Perform your usual development tasks such as running tests, building conda-store, etc.

To test the changes to the test GH workflows, you can use https://github.com/nektos/act and run the workflow or job accordingly. For example, to only run the conda-store integration tests: `act -j "integration-test-conda-store-server"`

PR #827, opened 2024-05-30
Update runtime dependencies

peytondmurray:
## Description

This pull request:

This PR updates runtime dependencies in preparation for the upcoming release.

See #801 for the related PR that this one was based on. The only difference between the dependencies changed here and the ones in that PR is that I removed `pytest-mock`, which isn't being used as far as I can tell.

## Pull request checklist

- [x] Did you test this change locally? **Yes - I tried installing using fresh conda environments, only installing `conda-store/environment.yaml` and `conda-store-server/environment.yaml`. Then I did `docker compose up --build`, and was able to successfully access the UI and build an environment with `numpy` as a dependency:

image

### DIFF
--- a/conda-store-server/pyproject.toml
+++ b/conda-store-server/pyproject.toml
@@ -50,12 +50,10 @@ dependencies = [
   "requests",
   "pydantic<2",
   "python-multipart",
-  "sqlalchemy<=1.4.47",
+  "sqlalchemy<2",
   "traitlets",
   "uvicorn",
   "yarl",
-  # installer
-  "constructor",
   # artifact storage
   "minio",
   "platformdirs >=4.0,<5.0a0"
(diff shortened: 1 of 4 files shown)


conda-store-server/pyproject.toml
```diff
@@ -75,15 +74,12 @@ dependencies = [
...
-  "flower",
-  "playwright",
   "pre-commit",
   "pytest",
   "pytest-celery",
-  "pytest-mock",
   "pytest-playwright",
   "twine>=5.0.0",
-  "pkginfo>=1.10",
+  "pkginfo>=1.10", # Needed tos upport metadata 2.3
```
trallard, line 82:
```suggestion
  "pkginfo>=1.10", # Needed to support metadata 2.3
```

trallard (approved):
LGTM just had to make a tiny change so will merge

Issue #830, opened 2024-06-10
[Meta] Test optimizations


peytondmurray:
### Context

I'm running tests locally, following the steps in `test_conda_store_server_unit.yaml` workflow. The unit tests for conda-store-server take a long time. Inside `conda-store-server/`, I ran `pytest --durations=0 -m "not extended_prefix and not user_journey" tests` and recorded the following durations:

<details><summary>Click to expand output</summary>
<p>

```
588.13s call     tests/test_actions.py::test_generate_constructor_installer[simple_lockfile_specification]
194.93s call     tests/test_actions.py::test_generate_constructor_installer[simple_specification]
172.98s call     tests/test_actions.py::test_generate_constructor_installer[simple_specification_with_pip]
118.69s call     tests/test_actions.py::test_generate_constructor_installer[simple_lockfile_specification_with_pip]
73.38s call     tests/test_actions.py::test_solve_lockfile[simple_specification_with_pip]
29.00s call     tests/test_actions.py::test_solve_lockfile_multiple_platforms[simple_specification]
24.80s setup    tests/test_actions.py::test_api_get_build_lockfile[conda_prefix1-False-3]
23.47s setup    tests/test_actions.py::test_api_get_build_lockfile[conda_prefix1-False-1]
23.11s setup    tests/test_actions.py::test_api_get_build_lockfile[conda_prefix1-True-1]
22.98s setup    tests/test_actions.py::test_add_conda_prefix_packages[conda_prefix1]
22.81s setup    tests/test_actions.py::test_api_get_build_installer[conda_prefix1]
22.74s setup    tests/test_actions.py::test_generate_conda_pack[conda_prefix1]
22.60s setup    tests/test_actions.py::test_api_get_build_lockfile[conda_prefix1-False-2]
22.41s setup    tests/test_actions.py::test_api_get_build_lockfile[conda_prefix1-False-0]
22.19s setup    tests/test_actions.py::test_generate_conda_export[conda_prefix1]
21.56s setup    tests/test_actions.py::test_generate_conda_docker[conda_prefix1]
17.62s call     tests/test_actions.py::test_solve_lockfile_multiple_platforms[simple_specification_with_pip]
15.39s call     tests/test_actions.py::test_install_specification
13.53s call     tests/test_actions.py::test_fetch_and_extract_conda_packages
12.25s call     tests/test_actions.py::test_solve_lockfile[simple_specification]
[...]

PR #833, opened 2024-06-12
MAINT - Mock out call to env solve in `test_generate_constructor_installer`

peytondmurray:
Partially addresses #830.

## Description

Currently, `test_generate_constructor_installer` accounts for 64% of the time spent on unit tests of `conda-store-server`. This is because the test checks whether specifications can be built into individual installers using [`constructor`](https://github.com/conda/constructor), which in turn asks `conda` to download `repodata.json` and solve the environment, which takes a long time and is susceptible to network disruptions. I don't think it's actually necessary for us to solve the environment each time here, since we're not interested in testing `conda` (or `constructor`) in this test.

This pull request modifies the test to still produce the input files needed to call `constructor`, but then mocks out the call to `constructor` itself, avoiding the environment solve entirely. Some simple checks of the call to `constructor` are made instead, and some quick verification of the files that are meant to be used as input are made using internal `constructor` functions.

### Other changes

- Added some type annotations in places that I needed to review to understand the code
- Fixed a DeprecationWarning for a bad escape sequence in a regex that I came across

## Pull request checklist


trallard (approved):

trallard (approved):
Thanks @peytondmurray, I had a quick look and this looks good to me, at least from a logic POV. 
I do want, however @jaimergp to have a quick look into this _just in case_ there is something missing but otherwise I will go ahead and approve.

trallard:
Also this PR needs updating as it has some merge conflicts @peytondmurray

jaimergp:
The mocking is solid here and lgtm.

As a side note, I think we are dealing with a symptom of something bigger: why are we running the solver for constructor? Don't we have a lockfile at this point? I think constructor will still create the environment locally even if passed an explicit lockfile, but if we do have a solved environment, the environment itself can also be used as the input, which is a bit more accurate in terms of bundling what we want to bundle.

trallard:
The CI is failing, so I re-ran this in case it was a flaky test run. 

> As a side note, I think we are dealing with a symptom of something bigger: why are we running the solver for constructor? Don't we have a lockfile at this point?

Ideally, we should have a `lockfile` at this point. @peytondmurray do you have more insights?

trallard:
Hmmm, still failing - per the logs, there is something iffy trying to find `conda_store_server._internal.action` as `conda_store_server.action`

peytondmurray:
Yep, this isn't a flaky test, it's a bad mock - the mock target changed since merging the privatization PR; should be fixed now.

peytondmurray:
> Ideally, we should have a lockfile at this point. @peytondmurray do you have more insights?

I'm not sure exactly how this would work; this action takes a specification (i.e. a python object representing an `environment.yml` or a `conda-lock.yml`) and formats that as a recipe (i.e. a `construct.yaml` and some post-install files) for `constructor`. Then it actually calls `constructor` in a subprocess.

The only thing this new test does is parameterize some specifications and some lockfiles, pass them to the underlying action, and mock out the subprocess call to `constructor`. Doesn't `constructor` _always_ solve specs?

PR #834, opened 2024-06-12
MAINT - Mock out calls to `conda-lock` for `test_solve_lockfile`

peytondmurray:
Partially addresses #830.

## Description

Currently `test_solve_lockfile` calls out to conda-lock and also invokes `conda info`, both of which are quite slow. This PR mocks out calls to `conda-lock` and `conda` in `conda_store_server/tests/test_actions.py::test_solve_lockfile`, and only checks that the invocations are correctly formed, not that conda-lock successfully executes. This reduces the runtime of the test to near zero.

## Pull request checklist


conda-store-server/tests/test_actions.py
```diff
@@ -70,14 +74,56 @@ def test_function(context):
...
-    assert len(context.result["package"]) != 0
+
+    # Check that the call to `conda_lock` is correctly formed
+    mock_run_lock.assert_called_once()
+    call_args = mock_run_lock.call_args_list[0][1]
+    assert str(call_args["environment_files"][0]).endswith("environment.yaml")
+    assert call_args["platforms"] == platforms
+    assert str(call_args["lockfile_path"]).endswith("conda-lock.yaml")
+    assert call_args["conda_exe"] == conda_store.conda_command
+    assert call_args["with_cuda"] == cuda_version
```
trallard, line 124: Nice! I like that not only this focuses on what we own, and, therefore should be testing, but these asserts seem a lot more robust 🚀

trallard (approved):
Thank you!!!! This looks pretty good and it makes sense to remove any third-party verification and focus only on what matters to conda-store. At least on unit tests.

Merging now!

PR #836, opened 2024-06-18
MAINT - Pin pydantic version 

trallard:
We started seeing CI failures with pydantic `1.10.15` so needed to add a pin to avoid such a version

## Pull request checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -22,7 +22,7 @@ outputs:
         - conda-store = conda_store.__main__:main
     requirements:
       host:
-        - python >=3.8
+        - python >=3.8,<=3.12.4
         - pip
         - hatchling >=1.14.0
         - hatch-vcs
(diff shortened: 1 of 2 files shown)


trallard:
@peytondmurray this is working now
Note that there has not been a release for pydantic on conda-forge yet so the alternative there is to pin the Python version (less than ideal but a quick fix for now)

Ready for review

recipe/meta.yaml
```diff
@@ -69,14 +70,17 @@ outputs:
...
         - conda-pack
         - conda-lock
+        - constructor
         - fastapi
         - filelock
         - itsdangerous
         - jinja2
         - minio
+        # @trallard there is no updated version in conda-forge
+        # need to come back and pin =1.10.16,<2.0a0
```
jaimergp, line 80:
```suggestion
        # need to come back and pin >=1.10.16,<2.0a0
```

jaimergp:
https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/779 should make this unnecessary. Also https://github.com/conda-forge/pydantic-feedstock/pull/125 is in, so 1.10.6 should be available soon.

PR #832, opened 2024-06-12
ENH - Update workflows triggers

trallard:
Since the CI takes so long, it does not make sense to run all the CI workflows when making docs-only changes.

This PR:
- **:construction_worker: CI should not run on docs**
- **:wrench: Update conda recipe** (forgot to update it recently when I updated the feedstock; please do not hate me)

PR #837, opened 2024-06-18
BUG - Fix triggers

trallard:
## Description

It seems the triggers updates in #832  are not currently working using the `!` ignore pattern in paths.
This PR changes such triggers to an explicit `include` paths pattern.

🚨 The build conda package workflow will fail until we merge #836 

## Pull request checklist

### DIFF
--- a/.github/workflows/test_build_docker_image.yaml
+++ b/.github/workflows/test_build_docker_image.yaml
@@ -6,9 +6,13 @@ env:
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


trallard:
@peytondmurray seems the triggers are working now as they should (the conda recipe check is still failing but should be fixed in #836)

peytondmurray (approved):
Thanks for this! :ship:

PR #840, opened 2024-07-02
restart workers after executing 10 tasks to mitigate memory leaks

Adam-D-Lewis:
Fixes #https://github.com/nebari-dev/nebari/issues/2418

## Description

This pull request:

- restarts celery workers after they execute 10 tasks (docs [here](https://docs.celeryq.dev/en/stable/userguide/workers.html#max-tasks-per-child-setting)).  This mitigates memory leaks which are occurring.

## Pull request checklist

## How to test

@trallard  edited from: https://github.com/conda-incubator/conda-store/pull/840#issuecomment-2207828243

To test, you could check out this branch, then run conda standalone [conda.store/conda-store/how-tos/install-standalone](https://conda.store/conda-store/how-tos/install-standalone) or you can run it in [docker containers](docker-compose up --build -d as explained [here](https://conda.store/community/contribute/contribute-code)). After you startup, you can visit localhost:8080/conda-store/admin and rebuild the environment it comes with a few times. If you watch the docker memory usage (docker stats) you'll see that the memory usage does not continue to climb with each new build as it does without this option set.


Adam-D-Lewis:
@peytondmurray @kcpevey @trallard Could I get a review on this?

Adam-D-Lewis:
I'm happy to make this configurable via a traitlet as well if desired, but the worker restarts seem happen quickly so I can't see too much down side to just applying it for all conda store deployments.

conda-store-server/conda_store_server/_internal/worker/app.py
```diff
@@ -72,6 +72,7 @@ def start(self):
         argv = [
             "worker",
             "--loglevel=INFO",
+            "--max-tasks-per-child=10",  # mitigate memory leaks
```
aktech, line 75: Looks good to me. I am assuming there is no limit by default.

Adam-D-Lewis: Yes, that's right.  By default, the workers will never restart due to finishing a certain number of tasks.

kcpevey:
I can't speak to the technical aspect of this, but I'm very glad to see it. I see the memory leakage on all my deployments. As far as I can tell it leads to an erroneously failed build due to OOM - which also generally fails with errors that make no sense. All around, this is a big admin management and end user UX improvement.

peytondmurray (commented):
Huh, with this option it looks like celery will just automatically restart the worker after 10 tasks are completed. My one complaint with this is it doesn't strictly speaking fix https://github.com/nebari-dev/nebari/issues/2418 as we are only treating a symptom not the memory usage itself.

But in any case from what I have read it seems like this is exactly the kind of use case the option was intended for, so I guess let's just use it. Is there an easy way to test this? If so it would be nice to ensure that it works.

Adam-D-Lewis:
> Huh, with this option it looks like celery will just automatically restart the worker after 10 tasks are completed. My one complaint with this is it doesn't strictly speaking fix [nebari-dev/nebari#2418](https://github.com/nebari-dev/nebari/issues/2418) as we are only treating a symptom not the memory usage itself.

I agree it does just treat the symptom without fixing the underlying issue.  That said, I think this will work robustly.

> But in any case from what I have read it seems like this is exactly the kind of use case the option was intended for, so I guess let's just use it. Is there an easy way to test this? If so it would be nice to ensure that it works.

You could check out this branch, then run conda standalone https://conda.store/conda-store/how-tos/install-standalone or you can run it in [docker containers](docker-compose up --build -d as explained [here](https://conda.store/community/contribute/contribute-code)).  After you startup, you can visit localhost:8080/conda-store/admin and rebuild the environment it comes with a few times.  If you watch the docker memory usage (`docker stats`) you'll see that the memory usage does not continue to climb with each new build as it does without this option set.

peytondmurray:
> To test, you could check out this branch, then run conda standalone https://conda.store/conda-store/how-tos/install-standalone or you can run it in [docker containers](docker-compose up --build -d as explained [here](https://conda.store/community/contribute/contribute-code)). After you startup, you can visit localhost:8080/conda-store/admin and rebuild the environment it comes with a few times. If you watch the docker memory usage (docker stats) you'll see that the memory usage does not continue to climb with each new build as it does without this option set.

Ahh, what I meant was is there a simple way of adding automated testing?

trallard:
> Ahh, what I meant was is there a simple way of adding automated testing?

I am +1 with Peyton here. It is good that we can manually verify (once or so), but I would prefer having some sort of automated tests (akin to load tests) to ensure this is, in fact, a useful fix.
Those tests could later be used to also do some benchmarking/memory checks. 
So while the fix itself seems fine I would suggest holding on merging until the tests have been added.

Adam-D-Lewis:
I've followed up on where the memory leak is occurring using memray.  I'm new to memary, but the flamegraph seems to point to line 20 in list_conda_prefix_packages at /opt/conda-store-server/conda_store_server/_internal/action/add_conda_prefix_packages.py which is `prefix_data.load()` [here](    prefix_data.load()).

I'm not sure what that code is doing or why that would cause a memory leak, but I'm recording it here for now.  I'll try to find out more later or how to do a more targeted fix.

image

peytondmurray:
Thanks, this is actually enough to start properly narrowing this down to something we can properly troubleshoot. I still think it would be worth merging this feature as it doesn't seem like there's a real cost to this. I'll open another issue with the goal of actually eliminating the memory leak.

trallard:
While I can see why `add_conda_prefix_packages.py` could lead to memory growth (and potential issues if the json.load() and its in-mem representations are not handled accordingly, I cannot see a definite indication that this _is indeed_ the culprit.
`add_conda_prefix_packages.py ` definitely spends a lot of time in CPU but that does not necessarily mean it is leading to memory leaks or bloat. 

So we'd definitely need to do more profiling and mem analysis over time.
@Adam-D-Lewis can you share a minimally reproducible example of how you got to the flame graph you shared.

As per merging, since this PR has been marked as needing tests I think we should still hold until we have a test in place.

Adam-D-Lewis:
I plan to write what I did to generate the memray flamegraph soon.

As far as a test showing the leak, what I'm planning on doing is running a test which builds many conda envs and show that the memory used by the conda store worker docker container continues to rise with each build.

peytondmurray (approved):
After a discussion with @Adam-D-Lewis and @trallard, it seems like this is potentially only observable in docker. For now we will just merge this now without a test as it is low risk, and come back to testing memory consumption later on once we know more.

Issue #842, opened 2024-07-03
[MAINT] Bump all python versions to 3.12


peytondmurray:
### Context

The dev environment, CI etc are all running python 3.10. Active support ended over a year ago, so let's bump python to the latest supported version across the board.

### Value and/or benefit

We'll get all the benefits of updating python!

peytondmurray:
Closed by #844.

PR #841, opened 2024-07-03
DEV - Simplify Docker images

trallard:
This is a stack on top of #838 

## Description

This PR continues to simplify the development and test setup for conda-store

- Removes unnecessary `environment.yaml` files (less files to keep in check/updated)
- Simplifies and adds comments to Dockerfiles
- Bumps the base image to `[24.3.0-0](https://hub.docker.com/layers/condaforge/mambaforge/24.3.0-0/images/sha256-21f7908a6c984ae164cb51c8af71ef9901693630f4b744038e5df6ef968d9b05?context=explore)`

## Pull request checklist

- [x] Did you test this change locally?
- [ ] Did you update the documentation (if required)?
- [ ] Did you add/update relevant tests for this change (if required)?

## Additional information

Since this stacks on another PR, some integration and unit tests might fail (and will be fixed in #838). For now, the important parts are :
* Docker-related workflows, which are working.
* Integration tests (conda-store and conda-store-server) - in the sense that the container should be building and `docker compose up` should be working (which it is)

I built the Docker images locally, then used these ones for `docker-compose`, and all seemed to work.

PR #838, opened 2024-07-01
DEV - Update dev environments

trallard:
Supersedes https://github.com/conda-incubator/conda-store/pull/801

## Description

This PR should remove some unnecessary dependencies and environment file. This was originally done in https://github.com/conda-incubator/conda-store/pull/801 but with all the recent changes it was easier to clean this up in a separate PR

## Pull request checklist

- [x] Did you test this change locally?
- [x] Did you update the documentation (if required)?
- [x] Did you add/update relevant tests for this change (if required)?

PR #844, opened 2024-07-04
DEV -  Set a canonical default Python version (3.12)

trallard:
This is another stack to #838  and intended to be merged _after_ #841 

## Description

While reviewing #841, @peytondmurray suggested bumping the default Python version used for CI, Docker images, and the like.

This PR bumps us from `3.10` to `3.12`:

* Adds a `.python-version-default` file to serve as the canonical source of truth across all the places where we set Python versions
* Updates `./github/workflows` to use this new file to get the canonical Python version
* Bumps the Python version in other places

## Pull request checklist


peytondmurray (approved):
I was wondering if there was any other way to solve this, but this seems reasonable.

The only ones I could find that were missed are the entries in `pyproject.toml`. Do we need to worry about those?

Also the tests are failing, but I think that's from the previous PR that got merged.

jaimergp:
This is a better way of consolidating the information. My two cents:

- There's no standard filename or location for this type of file, but I do see some projects using this filename. There are others like simply `.python-version` (no `-default`).
- Parsing from a plain text file might introduce accidental trailing whitespace. Usually tricky to debug. We could make sure that the file is always whitespace stripped with a pre-commit rule for that file only. Or if there's none, a simply Python step in the workflow:

```yaml
- name: Make sure python-version has no whitespace
  shell: python
  run: |
    from pathlib import Path
    assert Path(".python-version-default").read_text().strip() == Path(".python-version-default").read_text()
```

- The environment.yml file python version might get out of date. There's a comment about it but if someone updates the python-version-default file they might not see the environment.yml one. So I suggest we add a little check step like above:

```yaml
- name: Make sure python-version has no whitespace
  shell: python
  run: |
    from pathlib import Path
    assert f'- python={Path(".python-version-default").read_text().strip()}' in Path("environment.yml").read_text()
```

peytondmurray:
Yep, now that you mention it pre-commit isn't hitting this file with the current configuration. Can we change it to strip trailing whitespace everywhere? I don't actually know of a single situation in which I'd _want_ trailing whitespace.

```yaml
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.6.0
    hooks:
      - id: trailing-whitespace
        files: ".*\\.py"
      - id: check-added-large-files
      - id: check-toml
      - id: end-of-file-fixer
      - id: trailing-whitespace
```

jaimergp:
And maybe make sure `end-of-file-fixer` doesn't add the extra newline?

trallard:
> Also the tests are failing, but I think that's from the previous PR that got merged.

@peytondmurray No this is the same original error in my PR about the `pip` key rather than from the previous PR

> The environment.yml file python version might get out of date. There's a comment about it but if someone updates the python-version-default file they might not see the environment.yml one. So I suggest we add a little check step like above:

My goal would be to eventually just get rid of `environment.yaml` files; right now, the only one that might be going stale would be the `dev-environment.yaml` as we are otherwise using the `python-default-version` file to set the Python version elsewhere. 

Now if someone uses that locally then that might be out of date 🤷🏽‍♀️  but I see the proposed workflow as already more robust _sans_ the end of file/whitespace thingy.

trallard:
> The only ones I could find that were missed are the entries in pyproject.toml. Do we need to worry about those?

Why would we need to change these? These are the Python requires `3.8` so unless we are dropping `3.8-11` we do not need to change these

trallard:
I fixed the pre-commit hooks, so there should be no need to add extra steps.

> There's no standard filename or location for this type of file, but I do see some projects using this filename. There are others, such as simply .python-version (no default).

Since there is no standard, I used `python-version-default` for symmetry with what we were using in some GH workflows (env variable), but I am not privy to keeping or changing the name in case either of you have a preference @jaimergp @peytondmurray

peytondmurray:
I'm totally fine with `.python-version-default`, and the changes `.pre-commit-config.yml` look good to me. if @jaimergp is okay with it, let's merge. :rocket:

trallard:
Ok will merge.

peytondmurray:
Closes https://github.com/conda-incubator/conda-store/issues/842.

PR #847, opened 2024-07-08
MAINT - Add dependabot config

trallard:
## Description

We already use dependabot for `npm` package updates (mostly docs) so I decided to add a `.dependabot.yaml` to also:

- include Docker images bases to the updates
- include github actions to the updates
- provide better commit messages and reuse our labels
- set the updates to `monthly` to avoid unnecessary noise

## Pull request checklist


peytondmurray (approved):
Great idea to introduce this. Let's merge it! :ship:

PR #864, opened 2024-08-16
[ENH] Allow fetching of arbitrary user environments

### LINKED ISSUE #857: [ENH] - Add endpoint for listing all envs a user has access to
### Feature description

This is a feature request to add an endpoint which would list all the environments a user has access to. 

### Value and/or benefit

This feature would enable platforms like Nebari to present JupyterHub/Lab interfaces with a reduced list of "available" environments. Currently, when conda-store is used in jhub-apps, all environments are returned and with many people on the platform. Not only is this unwieldy, users dont actually have permission to access those environments. 

### Anything else?

[The nebari issue where this arose. ](https://github.com/nebari-dev/nebari/issues/2193)

peytondmurray:
Fixes #857.

## Description

This pull request:

- Adds the ability to specify a JSON web token (JWT) as a query parameter when making a GET request to the `/environment/` endpoint. The response will include only environments visible to the specified JWT.
- Adds the ability to specify a `role_mapping` parameter for the corresponding Python API, `conda_store_server.api.list_environments`. The role mapping should be a dictionary of the same type used to configure role mappings in the `conda_store_config.py`, e.g. passing:

```python
role_mapping={
  "*/*": ["editor"],
  "default/env": ["viewer"]
}
```

to `list_environments` would return only environments that match either of the two dictionary keys; the keys mean the same thing as they do in the `conda_store_config.py`:

`"*/*"` = match all namespace names and match all environment names
`"default/env"` = match any namespace named "default" and match any environment named "env"

- To enable this, I moved `conda_store_server.server.auth.AuthenticationBackend.compile_arn_sql_like`, which turns a role mapping key into a SQL-compatible regex that can be used to filter namespaces and backends, to `conda_store_server._internal.utils.compile_arn_sql_like` so that I could use it outside of the `AuthenticationBackend`. Doing this required updating a few different files, and I took the opportunity to define some useful type aliases, leave some type annotations, and add some docstrings.

## Pull request checklist

## How to test

Tests of new functionality in both the REST API and the Python API were added, so CI should take care of this. You can also run them locally with `pytest`.


conda-store-server/conda_store_server/_internal/environment.py
```diff
@@ -141,3 +144,40 @@ def _append_pip_packages(specification, packages):
         )
 
     return specification
+
+
+def filter_environments(
+    query: Query,
+    role_bindings: schema.RoleBindings,
+) -> Query:
+    """Filter a query containin environments and namespaces by a set of role bindings.
```
dcmcand, line 153:
```suggestion
    """Filter a query containing environments and namespaces by a set of role bindings.
```

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -620,25 +631,77 @@ async def api_list_environments(
     status: Optional[schema.BuildStatus] = None,
     packages: Optional[List[str]] = Query([]),
     artifact: Optional[schema.BuildArtifactType] = None,
-    auth=Depends(dependencies.get_auth),
-    entity=Depends(dependencies.get_entity),
-    paginated_args=Depends(get_paginated_args),
-    conda_store=Depends(dependencies.get_conda_store),
+    jwt: Optional[str] = None,
+    auth: Authentication = Depends(dependencies.get_auth),
```
trallard, line 635: Parameters with default values (even if optional) should come after those without.
Similiarly, optional should come after required ones.

conda-store-server/conda_store_server/_internal/environment.py
```diff
@@ -141,3 +144,40 @@ def _append_pip_packages(specification, packages):
...
+    query: Query,
+    role_bindings: schema.RoleBindings,
+) -> Query:
+    """Filter a query containing environments and namespaces by a set of role bindings.
+
+    Parameters
+    ----------
+    query : Query
+        Query containing both environments and namespaces
+    role_bindings : schema.RoleBindings
```
trallard, line 159: What are the allowed `RoleBindings`? Going by https://github.com/peytondmurray/conda-store/blob/fad647e86f9b9acd8ae4b9148f1cf8e1e970b5ff/conda-store-server/conda_store_server/_internal/schema.py#L33 It would appear that anything that matches `namespace` and `environment` bindings per https://github.com/peytondmurray/conda-store/blob/fad647e86f9b9acd8ae4b9148f1cf8e1e970b5ff/conda-store-server/conda_store_server/_internal/schema.py#L36 would be acceptable values?

conda-store-server/conda_store_server/_internal/schema.py
```diff
@@ -23,12 +23,15 @@ def _datetime_factory(offset: datetime.timedelta):
 # this is a more restrictive list
 ALLOWED_CHARACTERS = "A-Za-z0-9-+_@$&?^~.="
 ARN_ALLOWED = f"^([{ALLOWED_CHARACTERS}*]+)/([{ALLOWED_CHARACTERS}*]+)$"
+ARN_ALLOWED_REGEX = re.compile(ARN_ALLOWED)
```
trallard, line 26: What does ARN stand for? seems to be either `name` `namespace` or a combination of these?

trallard: perhaps best to add a comment is needed

conda-store-server/conda_store_server/_internal/utils.py
```diff
@@ -147,3 +149,33 @@ def callable_or_value(v, *args, **kwargs):
...
+
+
+def compile_arn_sql_like(
+    arn: str, allowed_regex: re.Pattern[AnyStr]
+) -> tuple[str, str]:
+    """Turn an arn into a string suitable for use in a SQL LIKE statement.
+
+    Parameters
+    ----------
+    arn : str
```
trallard, line 161: Ah this is the response to my earlier question, let's add a comment in the schema then

conda-store-server/conda_store_server/_internal/utils.py
```diff
@@ -147,3 +149,33 @@ def callable_or_value(v, *args, **kwargs):
...
+    Returns
+    -------
+    tuple[str, str]
+        (namespace regex, environment regex) to match in a sql LIKE statement.
+        See conda_store_server.server.auth.Authentication.filter_environments
+        for usage.
+    """
+    match = allowed_regex.match(arn)
+    if match is None:
+        raise ValueError(f"invalid arn={arn}")
```
trallard, line 179: Let's make this a more user-friendly message. Otherwise it is ambiguous re wether the query was invalid/malformed or if there was no match.
Could even go as far as being `raise ValueError(f"Could not find a match for the requested namespace/environment={arn}")`
```suggestion
        raise ValueError(f"Could not find a match for the requested arn={arn}")
```

conda-store-server/conda_store_server/server/auth.py
```diff
@@ -219,7 +216,7 @@ def compile_arn_regex(arn: str) -> re.Pattern:
           - "example-asdf"
           - "example-asdf/example-qwer"
         """
-        if not ARN_ALLOWED_REGEX.match(arn):
+        if not schema.ARN_ALLOWED_REGEX.match(arn):
```
trallard, line 219: For consistency with `utils.py` and since it's better to check against what we expect vs what we do not
```suggestion
        if schema.ARN_ALLOWED_REGEX.match(arn) is None:
```

conda-store-server/environment-dev.yaml
```diff
@@ -32,3 +31,4 @@ dependencies:
   - pip
   - pip:
       - pytest-playwright
+      - build
```
trallard, line 34: Added this to kick the CI back again

trallard (changes requested):
Thanks @peytondmurray I left a number of suggestions and questions.

Other questions as this was not entirely clear to me from the PR description and after seeing the tests.

1. As an end user if I want to use the REST API to get a list of envs I need to pass then
```
role_mapping={
  "*/*": ["editor"],
  "default/env": ["viewer"]
}
```
as a query parameter like `environments?role_mapping`? or how would you expect this to work?

2. if so, what happens if the user does not provide a role for example
```
role_mapping={
  "*/*",
}
```
is that even possible?

peytondmurray:
> Thanks @peytondmurray I left a number of suggestions and questions.
> 
> Other questions as this was not entirely clear to me from the PR description and after seeing the tests.
> 
>     1. As an end user if I want to use the REST API to get a list of envs I need to pass then
> 
> 
> ```
> role_mapping={
>   "*/*": ["editor"],
>   "default/env": ["viewer"]
> }
> ```
> 
> as a query parameter like `environments?role_mapping`? or how would you expect this to work?

No, that's only for the Python API. For the REST API, you pass a JWT of the user you want to fetch environments for, e.g. `/environment/?jwt=<jwt-of-whatver-user>`.

>     2. if so, what happens if the user does not provide a role for example
> 
> 
> ```
> role_mapping={
>   "*/*",
> }
> ```
> 
> is that even possible?

Role mappings are dictionaries, so this isn't currently possible - you need associated roles in order for this to work.

conda-store-server/conda_store_server/_internal/environment.py
peytondmurray: Allowed values for `RoleBindings` are dictionaries mapping strings to lists of strings, just like you'd find in the role bindings configuration inside `conda_store_config.py`. Here are some examples:

```python
{
    "foo*/bar*": ['viewer'],
}
```

Describes `viewer` level access for any environments starting with `bar` (the `*` is wildcard) in any namespace starting with `foo`.


```python
{
    "*/*": ['admin'],
}
```

Describes `admin` level access for any environments in any namespace.


```python
{
    "test/env": ['editor'],
    "*/python": ['viewer'],
}
```

Describes `editor` level access for the environment named `env` in the namespace named `test`. It also describes `viewer` level access for any environment named `python` in any namespace.

conda-store-server/conda_store_server/_internal/schema.py
peytondmurray: I don't know what ARN stands for, but it seems to mean "a key in a RoleMapping". It's the part of a role mapping that describes what namespace/environment combination a role is mapped to.

conda-store-server/conda_store_server/_internal/server/views/api.py
peytondmurray: Whoops, I'll move these optional args below the `Depends` args.

conda-store-server/conda_store_server/_internal/utils.py
peytondmurray: Thanks, this is far better than the original.

trallard:
So for the REST API the default behavior is to just return all the environments a user has some access to. 
No filtering through role or something else.

peytondmurray:
No, sorry, I should have made that more clear - for the REST API, any user `A` who requests the environments that `B` has access to must also have access to those environments, following the discussion we had previously.

trallard:
Right I did see the filtering when doing the code review this

> No, that's only for the Python API. For the REST API, you pass a JWT of the user you want to fetch environments for, e.g. /environment/?jwt=<jwt-of-whatever-user>.

was what confused me. So all good then.
Now, does the whole `environment` fetching need to be documented somewhere? I see the comments, which are good, but I'm thinking of end-users who might want to use this new functionality.

peytondmurray:
100% agreed. Once the openapi.json reference is updated we'll have that. But maybe it make sense to start a document where some common uses cases are shown in examples?

trallard (approved):

PR #865, opened 2024-08-27
MAINT - Miscellaneous maintenance/governance tasks

trallard:
A few tasks were missing from when we moved the repo to the conda-incubator organisation.

- **Add CoC**: point to the conda CoC
- **Add licence header** to all files

This PR touches a lot of files but adds a standard license header, so there is no need for a thorough check just an approval and merge, provided I did not break anything on CI.


LICENSE
```diff
@@ -10,11 +10,11 @@ met:
 1. Redistributions of source code must retain the above copyright
 notice, this list of conditions and the following disclaimer.
 
-2. Redistributions in binary form must reproduce the above copyright
+1. Redistributions in binary form must reproduce the above copyright
 notice, this list of conditions and the following disclaimer in the
 documentation and/or other materials provided with the distribution.
 
-3. Neither the name of the copyright holder nor the names of its
+1. Neither the name of the copyright holder nor the names of its
```
peytondmurray, line 17: How come these were changed from `2.` and `3.` to `1.` and `1.`?

conda-store-server/LICENSE
```diff
@@ -7,11 +10,11 @@ met:
 1. Redistributions of source code must retain the above copyright
 notice, this list of conditions and the following disclaimer.
 
-2. Redistributions in binary form must reproduce the above copyright
+1. Redistributions in binary form must reproduce the above copyright
 notice, this list of conditions and the following disclaimer in the
 documentation and/or other materials provided with the distribution.
 
-3. Neither the name of the copyright holder nor the names of its
+1. Neither the name of the copyright holder nor the names of its
```
peytondmurray, line 17: Same note as for the `LICENSE` file.

conda-store-server/environment-dev.yaml
```diff
@@ -32,3 +35,4 @@ dependencies:
   - pip
   - pip:
       - pytest-playwright
+      - build
```
peytondmurray, line 38: :+1:

conda-store/LICENSE
```diff
@@ -7,11 +10,11 @@ met:
 1. Redistributions of source code must retain the above copyright
 notice, this list of conditions and the following disclaimer.
 
-2. Redistributions in binary form must reproduce the above copyright
+1. Redistributions in binary form must reproduce the above copyright
 notice, this list of conditions and the following disclaimer in the
 documentation and/or other materials provided with the distribution.
 
-3. Neither the name of the copyright holder nor the names of its
+1. Neither the name of the copyright holder nor the names of its
```
peytondmurray, line 17: Same note as for `LICENSE`.

examples/docker/assets/overrides.json
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) conda-store development team. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
peytondmurray, line 3: This breaks JSON compatibility. Should we now rename this to `overrides.json5`?

peytondmurray (commented):
I left a few comments/questions. One other question: why do we need the license in every single file? I've seen organizations do this and I've never been clear on exactly why it is necessary.

LICENSE
trallard: It seems it was the pre-commit hooks. It should render ok, but I can roll back to 1,2,3

conda-store-server/LICENSE
trallard: Yep I copy-pasted across the files so that they were the same

examples/docker/assets/overrides.json
trallard: Could do

trallard:
Depending on the license and its attribution requirements, source code should retain the copyright notice and license. For example, if someone reuses a file from this project, they have to maintain the license/copyright. Hence, it makes it easy for them and also for folks who need to audit OSS use and the like.

peytondmurray:
Looks fine to merge as-is - the pre-commit issue is with the autogenerated `openapi.json`. We can either make a new PR to ignore that or fix the autogeneration workflow.

peytondmurray (approved):

trallard:
The pre-commit thing should be fine as is I suppose, it will just push the fix when the docs are added.  So will merge now and if we struggle we can change it in a separate PR.

PR #881, opened 2024-09-17
Sort envs returned by REST API by current build's scheduled_on time

### LINKED ISSUE #859: [ENH] - Ensure completeness when fetching all pages using REST API
### Feature description

Currently the pagination API uses page-based pagination without sorting by time of creation/modification. This leads to incomplete results when iterated over the pages in dynamic systems with multiple modifications happening at the same time when results are fetched.

For example, if during the process of iterating over the pages of environments (`api/v1/environment`) an additional environment is created, either:
- the new environment will be omitted, or
- a different random environment will be omitted

this is because the environments are sorted by [namespace and name](https://github.com/conda-incubator/conda-store/blob/1df30a2132228d9ac40ed7230e01c5b625d8e7c8/conda-store-server/conda_store_server/_internal/server/views/api.py#L651). For a more concrete example see the details below.

<details>

Say we have five environments A, B, D, E, F, and page size equals 2.
User performs two actions:
- requested creating environment called "C", and then
- opens the form with a list of all environments

If the database commit happens after the second page is fetched, this could lead to the form getting replies:

- A, B - first page (ok)
- D, E - second page (ok at the time)
- E, F - third page - oops!

If many users are creating environments, the admin would be randomly missing some environments.

If the results were sorted by date of environment was created (/last modified if environments can be renamed), then this is not a problem, because the replies could be (assuming A, B, D, E, F were created in this order):

- A, B
- D, E
- F, C

Originally posted in https://github.com/nebari-dev/nebari/issues/2599#issuecomment-2260552463

</details>

Instead, one of pagination implementations which guarantees data completeness should be used across conda-store REST API:
- cursor-based pagination
- page pagination with a sort order guaranteeing that newly added records are on the last page (e.g. by date of creation/modification for environments)
- time-based pagination

### Value and/or benefit

No randomly missing items (e.g. environments) in paginated replies.

### Anything else?

_No response_

peytondmurray:
Fixes #859.

## Description

This PR sorts environments returned when a GET is sent to `/environment/` by the time the environments' current builds were submitted. In systems where users are creating environments while environments are being queried, this ensures that _all_ results are returned; other sorting methods (e.g. by name) can yield incomplete results.

## Pull request checklist


peytondmurray:
After a discussion, it seems like no matter how we solve this there's going to be a breaking change to the API. Cursor-based pagination will allow for any sorting order, so I'll look into implementing that, probably through `fastapi-pagination`.

peytondmurray:
Adding pagination via `fastapi-pagination` seems simple on the surface, but they don't publish an API reference in the docs, so I'm not sure how it affects what query parameters are accepted by our API. So in the process of trying to implement this, I started working on adding hot reloading, but due to the fact that we're using `traitlets` to load server configuration (and instantiate the FastAPI app) and that `uvicorn` has specific requirements about how the FastAPI app is instantiated when hot reloading, it's trickier to get this working than I thought. See #901 for details.

peytondmurray:
Hot reloading's done, but the `fastapi_pagination` docs aren't enough to get this working easily, and in the time it would take to actually read through the source for the project I think we can just implement a simple cursor-based paginator instead :/

peytondmurray:
Not sure if this is actually ready to be merged, as it is not currently compatible with the frontend, and a discussion about REST API versioning needs to be had before this approach can be merged. But otherwise, tests are passing and the implementation is I think mostly complete :) @soapy1

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -15,13 +15,37 @@
...
+    paginate,
+)
 from conda_store_server.conda_store import CondaStore
 from conda_store_server.exception import CondaStoreError
 from conda_store_server.server import schema as auth_schema
 from conda_store_server.server.auth import Authentication
 from conda_store_server.server.schema import AuthenticationToken, Permissions
 
 
+def get_cursor(cursor: Optional[str] = None) -> Cursor:
```
soapy1, line 32: Whats the motivation for separating the the `cursor` from the rest of the pagination args?

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -15,13 +15,37 @@
...
+    return Cursor.load(cursor)
+
+
+def get_cursor_paginated_args(
```
soapy1, line 36: It looks like most of these `depends` type functions are in the [`conda_store_server/_internal/server/dependencies` ](https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/_internal/server/dependencies.py) module. I think it would be good to consolidate these into that module.

soapy1 (changes requested):
Looking good! Thank you for adding such beautiful formatted and thorough docs to a lot of these functions :star_struck: 

Just a few things to sort out. 

I took notes on what I tested manually in this gist https://gist.github.com/soapy1/beae935f78725cfd1dd0897c1ce016d4

conda-store-server/conda_store_server/_internal/server/views/api.py
peytondmurray: There's really no reason other than that the cursor based pagination takes a lot more code than limit/offset, so I've broken that off into a separate module, so that `views/api.py` is mostly endpoints. If it makes more sense to keep this as part of `views/api.py` I'm happy to do so, however.

conda-store-server/conda_store_server/_internal/server/views/api.py
peytondmurray: Totally agree, thank you!

soapy1:
One more question (related to how we want to approach versioning for this change): how do we want to approach pagination for the rest of the api? Do we want all the other endpoints to also adopt this type of pagination?

peytondmurray:
I vote yes - as far as I can tell the advantages of limit/offset are that it's easy to implement, but the cost of it is exactly what is described in #859. Cursor based pagination addresses that.

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -708,21 +736,29 @@ async def api_list_environments(
...
-            orm_environments,
-            paginated_args,
-            schema.Environment,
-            exclude={"current_build"},
-            allowed_sort_bys={
-                "namespace": orm.Namespace.name,
-                "name": orm.Environment.name,
-            },
-            default_sort_by=["namespace", "name"],
+        paginated, next_cursor, count = paginate(
```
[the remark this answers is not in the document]
peytondmurray: The `current_build` field is an optional field in the `Environment` model, so originally I thought there's no harm in allowing it to be returned here unless there's some specific reason to strip it out. Do you think this was intended to make the response smaller?

Anyway, the `current_build_id` is kept here, and FastAPI has a way of excluding fields so I've just made use of that to capture the old behavior.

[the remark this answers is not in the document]
soapy1: > Do you think this was intended to make the response smaller?

could be?

conda-store-server/conda_store_server/_internal/server/views/pagination.py
```diff
@@ -0,0 +1,333 @@
...
+        Returns
+        -------
+        list[ColumnClause]
+            A list of sqlalchemy columns corresponding to the order_by values passed
+            as a query parameter
+        """
+        columns = []
+        if order_by:
+            for order_name in order_by:
+                idx = self.order_names.index(order_name)
```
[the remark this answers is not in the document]
peytondmurray: Ahh, thanks for this - I've added error handling and a test for bad query parameters!

peytondmurray:
The cursor-based pagination is now on `/api/v2/environment/`, and `/api/v1/environment/` has been restored as well. Ready for a second look!

conda-store-server/conda_store_server/_internal/schema.py
```diff
@@ -509,8 +517,8 @@ class APIDeleteNamespaceRole(BaseModel):
 
 
 # GET /api/v1/environment
-class APIListEnvironment(APIPaginatedResponse):
-    data: List[Environment]
+class APIListEnvironment(APICursorPaginatedResponse):
```
soapy1, line 520: It looks like the response when I hit the `/v1/environments` api endpoint has changed to

```
$ curl --header "Authorization: bearer ${CS_TOKEN}" http://localhost:8080/conda-store/api/v1/environment/\?size\=2 | jq

. . .
  "status": "ok",
  "message": null,
  "cursor": null,
  "count": 6
}
```

I think this response from the v1 api should be unchanged (wrt to main).  So we'll need to add a new type for the v2 of the api. Something like:

```
class APIV2ListEnvironment(APICursorPaginatedResponse):
     data: List[Environment] = []

class APIListEnvironment(APIPaginatedResponse):
   data: List[Environment]
```

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -15,45 +15,27 @@
...
-        size = server.max_page_size
-    size = min(size, server.max_page_size)
-    offset = (page - 1) * size
-    return {
-        "limit": size,
-        "offset": offset,
-        "sort_by": sort_by,
-        "order": order,
-    }
+router_api_v2 = APIRouter(
```
soapy1, line 35: Can the v2 router and associated endpoints go in it's own module (maybe like `api_v2.py`??). I think that'll make it easier to understand that there are multiple living versions of the api and make it easier to make changes in the future.

soapy1 (commented):
Looks great! Just 2 small changes suggested

conda-store-server/conda_store_server/_internal/schema.py
peytondmurray: You're right - will update now.

soapy1 (approved):
Looks great :page_with_curl:

PR #885, opened 2024-09-19
BUG - patch release workflow

trallard:
## Description

It seems our latest release workflow failed. This PR should fix the issues finding the distribution artifacts.

## Pull request checklist


trallard:
This seems to be working now for ref here is a run with the env branch protection disabled to verify https://github.com/conda-incubator/conda-store/actions/runs/10944608931?pr=885

peytondmurray (approved):
Thanks for the fast fix and for testing that the workflow runs :rocket:

PR #898, opened 2024-10-18
REACT_APP_URL_BASENAME

gabalafou:
## Description

Companion PR to https://github.com/conda-incubator/conda-store-ui/pull/431

## Pull request checklist


gabalafou:
I realized while working on this that we need to have a strategy for how to handle what happens when a user goes directly to an React app-defined route, such as: "/default/new-environment".

Currently, the server will return a 404 instead of returning the [conda-store-ui.html](https://github.com/conda-incubator/conda-store/blob/7d39efa9c708f71875dc48d0dc2800df860b5703/conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html#L4) page. 

I don't know all the ins and outs of the server, but we should probably redirect all unknown routes to the React app.

gabalafou:
Created an issue:

- https://github.com/conda-incubator/conda-store/pull/898

trallard (approved):
Thanks @gabalafou  I had not noticed you opened this PR, thank you!!!

PR #907, opened 2024-10-21
[ENH] Add hot reloading for development

### LINKED ISSUE #901: [ENH] - Enable hot reloading when developing
### Feature description

Currently, development with `docker compose up --build` can be slow because making a small change requires the developer to `docker compose down`, then `docker compose up --build`, which redownloads the entire environment.

`uvicorn` has support for hot reloading (although there are [caveats](https://github.com/encode/uvicorn/issues/687)), and if we can mount the current filesystem into the container read/write, we should be able to develop and have the server hot reload when changes are made.

### Value and/or benefit

Faster development.

### Anything else?

_No response_

peytondmurray:
Fixes #901.

## Description

This pull request enables hot reloading of the `conda-store-server` webserver in both standalone or in containers. If running in standalone mode, `reload = True` by default.

## Pull request checklist


soapy1 (approved):
Looks great! Very excited to not be spinning up and down containers. 

Would be great if you could also drop a note in the contributing guide on how to use this setup.

peytondmurray:
Thanks! I added a note about hot reloading to this section: https://github.com/peytondmurray/conda-store/blob/901-hot-reloading/docusaurus-docs/community/contribute/contribute-code.md#local-development-without-docker---conda-store-core

PR #912, opened 2024-10-22
[DOCS] Describe new config vars for React router

### LINKED ISSUE #435: Emit logs frequently for debugging

gabalafou:
Closes #435.


peytondmurray (approved):
Thank you, this context is super helpful :writing_hand:

Issue #914, opened 2024-10-24
Improve testing - developer experience


soapy1:
### Context

Currently the test coverage on conda-store is minimal. This makes it difficult to identify when changes might have unintended consequences (like breaking some functionality in another part of the application). For example, it would be nice if issues like #615 could be caught by automated tests.

It would be nice to:
* [ ] have more test coverage in conda-store
* [x] remove tests that are not helpful (if possible)
* [x] have the ability to see code coverage information for each pr (eg. integration with code cov, or a github action)
* [x] be able to run unit tests locally (eg. the unit tests are fast)

### Value and/or benefit

This will give developers more confidence that their proposed changes will not have unintended consequences in other parts of the software.

### Anything else?

Currently the code coverage is:
```
Name                                                                                                    Stmts   Miss  Cover
---------------------------------------------------------------------------------------------------------------------------
conda_store_server/__init__.py                                                                             71     30    58%
conda_store_server/_internal/__init__.py                                                                    0      0   100%
conda_store_server/_internal/action/__init__.py                                                            14      0   100%
conda_store_server/_internal/action/add_conda_prefix_packages.py                                           30      1    97%
conda_store_server/_internal/action/add_lockfile_packages.py                                               22      1    95%
conda_store_server/_internal/action/base.py                                                                54      3    94%
conda_store_server/_internal/action/download_packages.py                                                   55      0   100%
conda_store_server/_internal/action/generate_conda_docker.py                                                5      1    80%
conda_store_server/_internal/action/generate_conda_export.py                                               10      1    90%
conda_store_server/_internal/action/generate_conda_pack.py                                                  6      0   100%
conda_store_server/_internal/action/generate_constructor_installer.py                                      64      4    94%
conda_store_server/_internal/action/generate_lockfile.py                                                   36      6    83%
conda_store_server/_internal/action/get_conda_prefix_stats.py                                               9      1    89%
conda_store_server/_internal/action/install_lockfile.py                                                    12      0   100%
conda_store_server/_internal/action/install_specification.py                                               10      0   100%
conda_store_server/_internal/action/remove_conda_prefix.py                                                  9      0   100%
conda_store_server/_internal/action/set_conda_prefix_permissions.py                                        19      4    79%
conda_store_server/_internal/action/utils.py                                                                2      0   100%
conda_store_server/_internal/alembic/env.py                                                                25      5    80%
[...]

PR #919, opened 2024-10-25
Test improvements - part 1

soapy1:
partially addresses https://github.com/conda-incubator/conda-store/issues/914

## Description

This PR makes a few improvements to the tests for conda-store-server:
* restructures the `conda-store-server/test/` directory to more closely match the directory structure of the conda-store-server app
* ~~fixes the utc `datetime.datetime.utcfromtimestamp()` deprecation warning~~
  * cancel build actions fail as a result of fixing this warning. Presumably because a dependency that conda-store relies on does not support this
* adds a pytest marker for slow tests to allow developers to not run slow tests locally (this is just a dev experience improvement,  all tests should continue to be run in CI). In order to use this, try `python -m pytest -m "not extended_prefix and not user_journey and not long_running_test" tests/`

As a result of the updates to the fixtures, running tests should be much faster. Before this change, running locally
```
$ python -m pytest -m "not extended_prefix and not user_journey and not long_running_test" --durations=0 tests/
. . .
======================== 1 failed, 225 passed, 9 deselected, 2 xfailed, 1338 warnings in 1201.96s (0:20:01) ==============================
```

After,
```
==== 1 failed, 198 passed, 38 deselected, 816 warnings in 389.80s (0:06:29) =====
```

Note, this change is mostly because of marking some particularly long running tests.

## Pull request checklist


peytondmurray (approved):
Thanks for this - this improved organization make it easier for folks to come in and contribute to, and the new marks will be useful for tracking the worst offenders in terms of execution time.

PR #932, opened 2024-10-29
Create catch-all/fallback route for UI app

### LINKED ISSUE #899: Valid sub-routes for the UI app result in 404
Should we redirect all unknown routes to the React (UI) app?

When a user goes directly to a React app-defined sub-route (by sub-route I mean any route that is not the root route, "/"), such as "/:namespace/new-environment", the server will return a 404 instead of returning the [conda-store-ui.html](https://github.com/conda-incubator/conda-store/blob/7d39efa9c708f71875dc48d0dc2800df860b5703/conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html) page. 

We should probably **redirect all unknown routes to the React (UI) app**, [conda-store-ui](https://github.com/conda-incubator/conda-store-ui).

That way a user can copy-paste a link like "https://example.com/conda-store/default/new-environment" directly into their browser, as opposed to always having to start from "https://example.com/conda-store". Being able to load non-root routes was the whole point of [introducing URL routing into the front end](https://github.com/conda-incubator/conda-store-ui/pull/389).

gabalafou:
Fixes #899.

## Pull request checklist

- n/a Did you update the documentation (if required)?

## How to test

Before starting, create an environment named "test" under the "default" namespace.

Set `url_prefix` to "/":

- localhost:8080/ -> 200, should load app
- localhost:8080/foo -> 200, should see 404 page (but HTTP status code should be 200)
- localhost:8080/foo/bar -> 200, UI loads with sidebar and blank main panel (this needs to be handled better by the client, the client thinks this is a route for namespace/environment, tries to request namespace=foo, environment=bar, but currently doesn't handle the error case, so it's basically a silent error)
- localhost:8080/default/test -> 200, should load the app page for the default/test environment
- localhost:8080/not-found -> 404 with page generated from React app, not the server. How can you tell? The server returns JSON.

Set `url_prefix` to "/conda-store", go to the following URLs, behavior should be exactly the same as above:

- localhost:8080/conda-store 
- localhost:8080/conda-store/foo
- localhost:8080/conda-store/foo/bar
- localhost:8080/conda-store/default/test
- localhost:8080/conda-store/not-found


conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html
```diff
@@ -28,8 +28,8 @@
              REACT_APP_URL_BASENAME: new URL("{{ url_for('get_conda_store_ui') }}").pathname
          };
         </script>
-        <script defer src="static/conda-store-ui/main.js"></script>
-        <link href="static/conda-store-ui/main.css" rel="stylesheet">
+        <script defer src="{{ url_for('get_conda_store_ui') }}static/conda-store-ui/main.js"></script>
+        <link href="{{ url_for('get_conda_store_ui') }}static/conda-store-ui/main.css" rel="stylesheet">
```
gabalafou, line 32: Adding `url_for('get_conda_store_ui')` was the only way I could figure out how to put the url_prefix here. 

Why is this change needed? Previously, the server would only serve the UI app at `url_prefix` (e.g., either at "/" or "/conda-store", depending on configuration).

However, this PR allows the UI app to be loaded at /conda-store/default/test, for example. At that URL, this page would try to fetch the JavaScript and CSS at:

- /conda-store/default/static/conda-store-ui/main.js
- /conda-store/default/static/main.css

These routes end up matching the catch-all, which means the server serves this template page for those requests instead of the desired JavaScript and CSS.

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
```diff
@@ -9,10 +9,11 @@
 
 router_conda_store_ui = APIRouter(tags=["conda-store-ui"])
 
-
 @router_conda_store_ui.get("/")
+@router_conda_store_ui.get("/{full_path:path}")
 async def get_conda_store_ui(
     request: Request,
+    full_path: str,
```
gabalafou, line 17: Is there a conventional way to mark this function argument as unused in the function body?

conda-store-server/conda_store_server/_internal/server/app.py
```diff
@@ -293,15 +293,32 @@ async def exception_handler(request, exc):
...
+            )
+            app.mount(
+                self.conda_store.storage.storage_url,
+                StaticFiles(directory=self.conda_store.storage.storage_path),
+                name="static-storage",
+            )
+
+        # This needs to come at the end because if the UI is enabled,
+        # it becomes the catch all route
+        if self.enable_ui:
```
gabalafou, line 318: The diff view here is confusing. What I did was move this entire if-block (`if self.enable_ui`) down to the bottom of this function (`init_fastapi_app`). And within the if-block, I moved the UI router from near the top to the very end.

Why does this need to be moved around? Because the order in which the routes are defined matters.

For example, if I put the `/favicon.ico` definition after the UI app, then a request for `/favicon.ico` will end up loading the UI app instead of the favicon.

gabalafou (commented):
self review

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
```diff
@@ -11,8 +11,10 @@
 
 
 @router_conda_store_ui.get("/")
+@router_conda_store_ui.get("/{full_path:path}")
```
gabalafou, line 14: This is the line that adds a catch-all route to the UI.

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
```diff
@@ -11,11 +11,16 @@
...
+    full_path: str,
     templates=Depends(dependencies.get_templates),
 ):
     context = {
         "request": request,
     }
-    return templates.TemplateResponse("conda-store-ui.html", context)
+    response = templates.TemplateResponse("conda-store-ui.html", context)
+    if full_path.endswith("not-found"):
+        response.status_code = 404
```
gabalafou, line 25: Decided to sneak this one in to support https://github.com/conda-incubator/conda-store-ui/pull/440

gabalafou:
It would be great to turn the manual test plan I put in the PR description into a set of automated tests

soapy1:
> It would be great to turn the manual test plan I put in the PR description into a set of automated tests

+100 that would be awesome! I would make a new test file in https://github.com/conda-incubator/conda-store/tree/main/conda-store-server/tests/_internal/server/views called `test_conda_store_ui.py`.  There are a bunch of fixutres defined in [conftest](https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/tests/conftest.py) that you can use to setup tests.

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
peytondmurray: Yeah, a preceding underscore:

```suggestion
    _full_path: str,
```

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
peytondmurray: Perfect time for it :rocket:

peytondmurray (approved):
Tested and the routes/html codes are as expected. Thanks!

peytondmurray:
Hmm, looks like something here is making the integration tests fail.

conda-store-server/conda_store_server/_internal/server/views/conda_store_ui.py
```diff
@@ -3,18 +3,34 @@
...
+    # docs. The docs show the endpoints as returning HTML rather than JSON.
+    default_response_class=HTMLResponse,
+)
 
 
+# Yes, "{path:path}"" makes "/" redundant. But if we don't define "/", then we
+# always have to pass `path` to the `url_for` function like so:
+# url_for("get_conda_store_ui", path="")
 @router_conda_store_ui.get("/")
+@router_conda_store_ui.get("{path:path}")
```
gabalafou, line 22: @peytondmurray I'm not entirely sure why, but this seems to be the line that is causing at least some of the CI checks to fail.

If you remove this line and go to localhost:8080/api/v1/namespace, you'll see that it redirects to same URL but with a forward slash at the end. If you then put this line back in and go to the same URL, you'll notice that it tries to load the client app. You get an HTML page back that tries to load the client app at that URL, but the client renders an error since it doesn't recognize that route (api/v1/namespace).

Any idea how to fix it?

gabalafou: Recap

With line 22 removed:

- localhost:8080/api/v1/namespace -> localhost:8080/api/v1/namespace/
- localhost:8080/api/v1/namespace/ -> JSON containing list of namespaces

With line 22 in place:

- localhost:8080/api/v1/namespace -> index.html for client app
- localhost:8080/api/v1/namespace/ -> JSON containing list of namespaces

gabalafou: This seems to suggest to me if FastAPI tries all the routes and cannot find a match, then it will add a trailing slash as a last-ditch attempt to not return a 404. 

I see there is an option `redirect_slashes` for the [FastAPI class](https://fastapi.tiangolo.com/reference/fastapi/#fastapi.FastAPI) whose default value is True. 

I will have to look at the FastAPI source code to see how and when it uses that option.

I wonder if the solution is to explicitly map both routes to the same handler. The following would code would allow the user to get the internal admin UI via `/admin` or via `/admin/`

```
@router_ui.get("")
@router_ui.get("/")
async def ui_list_environments(
```

peytondmurray: I don't have context on this, but it seems like this is intended behavior on the FastAPI side of things. Others have [wondered](https://github.com/fastapi/fastapi/discussions/7298) about how to prevent this and [about the impacts of redirection on http[s]](https://github.com/fastapi/fastapi/discussions/9328).

marcelovilla:
@gabalafou @peytondmurray is there any chance this PR will make the next conda-store release?

We're trying to upgrade conda-store in Nebari and would like to include this fix if possible.

peytondmurray:
@marcelovilla We're in the midst of exploring options here - yesterday I proposed to @gabalafou the idea of simply overriding FastAPI's default HTTPException handler with something like

```python
from fastapi import FastAPI, HTTPException
from fastapi.responses import RedirectResponse

app = FastAPI()

@app.exception_handler(HTTPException):
async def handle_bad_route(request, exc):
    return RedirectResponse("/")
```

Will get back to you on whether this can get in the next release!

marcelovilla:
@peytondmurray good to know, thanks for the prompt reply!

gabalafou:
I also considered that idea, but I wasn't sure it would work because some of the API endpoints raise a 404 if they cannot find the resource in the database. For example, the /api/v1/namespace/{namespace} endpoint raises 404 if it cannot find the namespace in the database:

https://github.com/conda-incubator/conda-store/blob/feac99d5c6c3c5fa1e17430e63d6ab7e8d2053da/conda-store-server/conda_store_server/_internal/server/views/api.py#L282-L299

My suspicion was that if we add a custom exception handler for 404s to serve the React app, then if someone makes a request to `/api/v1/namespace/nonexistent_namespace_example/`, they would get a 200 with an HTML page for the React app, but they should get a 404 with a JSON error payload instead.

I was imagining something like this:

```py
@app.exception_handler(404):
async def get_conda_store_ui(
    request: Request,
    templates=Depends(dependencies.get_templates),
    url_prefix=Depends(dependencies.get_url_prefix),
):
    context = {
        "request": request,
        "url_prefix": url_prefix,
        "ui_prefix": router_conda_store_ui.prefix,
    }
   return templates.TemplateResponse("conda-store-ui.html", context)
```

I created a minimal FastAPI app to double check my suspicion:

```py
# main.py

from fastapi import FastAPI, APIRouter, HTTPException
from fastapi.responses import HTMLResponse, RedirectResponse


app = FastAPI()

router_api = APIRouter(
    tags=["api"],
    prefix="/api/v1",
)


@router_api.get("/namespace/{namespace}")
async def get_namespace(namespace):
    if (namespace == "foobar"):
        raise HTTPException(status_code=404, detail="namespace does not exist")
    else:
        return {"message": f"namespace {namespace} found"}


@app.exception_handler(404)
async def handle_bad_route(request, exc):
    html_content = """
    <html>
        <head>
            <title>Conda Store</title>
        </head>
        <body>
            <h1>Pretend I'm a React app</h1>
        </body>
    </html>
    """
    return HTMLResponse(content=html_content, status_code=200)


app.include_router(router_api)
```

Result:

- 🟢 localhost:8000 => 200 HTML "Pretend I'm a React app"
- 🟢 localhost:8000/some/other/random/route => 200 HTML "Pretend I'm a React app"
- ❌ localhost:8000/api/v1/namespace/raise404 => 200 HTML "Pretend I'm a React app". (This should be a 404 w/JSON.)
- 🟢 localhost:8000/api/v1/namespace/foobar => 200 JSON `{"message":"namespace foobar found"}`

The only way I see around this is by subclassing or somehow annotating exceptions raised in our code so that we can re-raise them in the exception handler.

-----

Out of curiosity, do you have any concerns about the approach I've taken in my most recent commits, which is to put the app at `/ui`?

peytondmurray:
No, I don't have a problem with setting the default to `/ui/`. I tested locally

| URL Entered | Result |
| --- | --- |
| / | Redirected to /conda-store/ui/ |
| /conda-store/ui | (no redirect) |
| /conda-store/ui/ | (no redirect) |
| /conda-store/ui/username/foo (existing environment) | (no redirect) |
| /conda-store/ui/username/foo2 (nonexistent) | (no redirect, but the /ui/ interface appeared) |

gabalafou:
Great, thanks @peytondmurray!

1. This is good, should redirect
2. Should show the UI in the "no environment selected" state
3. Ditto
4. Should show the UI in the "view environment" state (for the "foo" environment)
5. I'm not exactly sure what the UI should show given the current state of the codebase; however, we'll have to add some client-side code that shows the user a 404 page when they try to view an environment that doesn't exist or that they don't have access to

peytondmurray:
Only one thing `test_conda_store_server_url_prefix`: can you update the `200` status code to be `307`?

gabalafou:
Cross-reference: 

- https://github.com/conda-incubator/conda-store/pull/993

PR #933, opened 2024-10-30
[MAINT] Update lint tooling

### LINKED ISSUE #934: [MAINT] Update lint tooling
### Context

The lint tooling could use some attention:

- Let's use `ruff format` instead of `black` for speed
- Let's get rid of `isort` in favor of `ruff`, again for speed and because all configuration will be unified
- It would be good to establish a docstring convention. Although some of our docstrings are unformatted, others follow numpydoc so I'd propose that.
- Let's enable more lint rules. Currently we don't enable any extra rules, and according to the `ruff` documentation:

> By default, Ruff enables Flake8's F rules, along with a subset of the E rules, omitting any stylistic rules that overlap with the use of a formatter, like ruff format or [Black](https://github.com/psf/black).

So while it's a good start, we could really benefit by enabling more rulesets. I propose:

```
    "E",   # pycodestyle
    "W",   # pycodestyle
    "F",   # Pyflakes
    "UP",  # pyupgrade
    "B",   # flake8-bugbear
    "SIM", # flake8-simplify
    "I",   # isort
    "N",   # pep8 naming
    "D",   # pydocstyle
    "ANN", # annotations
    "T10", # debugger
    "PT",  # flake8-pytest
    "RET", # flake8-return
    "ARG", # flake8-unused-arguments
    "FIX", # flake8-fixme
    "ERA", # flake8-eradicate
    "PD",  # pandas-vet
    "NPY", # numpy-specific rules
```

### Value and/or benefit

These checks will help us avoid a few more of the common issues that we see without tooling, and will help establish uniformity for our docstrings.

### Anything else?

_No response_

peytondmurray:
Closes #934.

## Description

This PR implements a couple of improvements to the lint system.

- `black` has been swapped out for the more performant `ruff format`
- `isort` has been removed in favor of enabling ruff's isort ruleset
- A few other rulesets have been enabled as well. However, due to existing violations, a number of individual rules have been disabled until we can fix them one by one.

The main changes were:

- `.pre-commit-config.yaml` has been updated to use `ruff` and `ruff format`
- `conda-store-server/pyproject.toml` has a list of updated rules (and ignores, which were added because of existing rule violations in code. These will be fixed in followup PRs). I've also formally made the docstring style numpydoc in the pydocstyle configuration; most of the (well formatted) docstrings in the code follow numpydoc anyway.
- Fixed some type annotations in `conda_store_server/__init__.py`; needed to add `from __future__ import annotations` to make this work for python<3.11
- The rest of the changes are considered "safe" changes introduced by `ruff --fix`.

## Pull request checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,12 +13,6 @@ ci:
     [pre-commit.ci] Apply automatic pre-commit fixes
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-      - id: black
-        exclude: "examples|tests/assets"
-
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
     rev: "v0.6.3"
@@ -26,6 +20,7 @@ repos:
       - id: ruff
         exclude: "examples|tests/assets"
         args: ["--fix"]
+      - id: ruff-format
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.6.0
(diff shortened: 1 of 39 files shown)


soapy1 (approved):
Neat!

PR #945, opened 2024-10-31
Move build to worker

soapy1:
Building should only be done by workers. Other pieces of the application should not be accessing this module.

## Pull request checklist

## Additional information

Previously, the coverage for `build.py`
```
conda_store_server/_internal/build.py         175     47    73%
```

Marginally increases
```
conda_store_server/_internal/worker/build.py      175     43    75%
```


conda-store-server/tests/_internal/worker/test_build.py
```diff
@@ -0,0 +1,40 @@
...
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+from conda_store_server import api
+from conda_store_server._internal import schema
+from conda_store_server._internal.worker import build
+
+
+def test_build_started(db, seed_conda_store):
+    test_build = api.get_build(db, build_id=4)
```
peytondmurray, line 11: Might be good to check that it _isn't_ started before you do the thing that is supposed to start it.

peytondmurray (approved):
I like the reorganization. I can see why this stuff wasn't tested before (it just sets values on `Build` items in the database in a pretty minimal way), but it's definitely good to have a more robust test suite here.

conda-store-server/tests/_internal/worker/test_build.py
soapy1: good idea! pushed up a change for this

PR #968, opened 2024-11-08
REL - 2024.11.1

soapy1:


peytondmurray (approved):
Looks good to me. Do we need to bump the `conda-store-ui` version for the release?

soapy1:
There isn't a new release of conda-store-ui, so I don't think we need to bump it here :balloon:

Issue #931, opened 2024-10-28
[MAINT] - sqlalchemy 2.0 migration


peytondmurray:
### Context

`sqlalchemy==2.0` has been out since January 2023. This issue is to track work on an upgrade path before support for <2.0 is dropped.

### Value and/or benefit

`sqlalchemy<2.0` is still seeing releases, but at some point legacy support will be dropped. It would be nice to get ahead of this so that we're not caught off guard by it when it gets announced.

PR #970, opened 2024-11-08
Bump sqlalchemy version

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/931

## Description

The upgrade path from sqlalchemy 1.4 -> 2.0 was designed to have a seamless upgrade path. This PR updates:

* the project dependency
* the conda recipe
* updates how some tests are formed to ensure that test objects are properly created

Docs: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#the-1-4-2-0-migration-path

## Pull request checklist


conda-store-server/pyproject.toml
```diff
@@ -69,7 +69,7 @@ dependencies = [
   # pinning them; conda-lock depends on this, but also doesn't pin the setuptools
   # version. See https://github.com/pypa/setuptools/issues/4478 for details
   "setuptools<70",
-  "sqlalchemy<2.0a0",
+  "sqlalchemy>=2.0,<2.1",
```
peytondmurray, line 72: what's the incompatibility with 2.1?

conda-store-server/tests/test_api.py
```diff
@@ -126,29 +125,37 @@ def test_namespace_role_mapping(db):
...
-        namespace_id=namespace.id,
-        entity="org/*",
-        role="editor",
+    db.add(
+        NamespaceRoleMapping(
+            namespace=namespace,
+            namespace_id=namespace.id,
+            entity="org/*",
+            role="editor",
+        )
```
peytondmurray, line 134: :rocket: Wayyyyyy better!

peytondmurray (approved):
Looks good - the tests will tell us whether there were any issues with the annotations. A few other minor comments, but on the whole this is brilliant - thank you! :rocket:

conda-store-server/pyproject.toml
soapy1: I think there shouldn't be any issue based on the [changelog](https://docs.sqlalchemy.org/en/21/changelog/changelog_21.html). But, I never tested this (sqlalchemy 2.1 is still in development) . Was setting this restriction out of an abundance of caution.

Issue #962, opened 2024-11-07
[ENH] - Add attestations for pip packages


peytondmurray:
### Feature description

We currently publish PYPI packages for `conda-store-server` (and `conda-store`, but that's not as important). Let's add support for [github attestations](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds).

### Value and/or benefit

Provenance will be established for our builds. I don't know if these will be needed by anyone, but it's extremely easy to do with a few extra github action calls and hardly any effort on our side.

trallard:
Just FYI since we are using the pypi-publish action with a trusted publisher, this adds attestations by default already https://github.com/pypa/gh-action-pypi-publish/tree/release/v1/?tab=readme-ov-file#generating-and-uploading-attestations (which are different from the GH attestations)

PR #978, opened 2024-11-13
MAINT - Do not use `nodefaults` moving forward

trallard:
Fixes https://github.com/conda-incubator/conda-store/issues/962

## Description

- While reviewing another PR, I realised there are a lot of warnings regarding the use of `nodefaults` due to changes in conda and miniconda-setup related to removing the `default` population in channels. This PR makes the necessary updates to remove these warnings.
- Since I was touching workflows, I decided to add the `attest-build-provenance-github` configuration to our release action to also generate GH attestions

## Pull request checklist

### DIFF
--- a/.github/workflows/generate_api_docs.yaml
+++ b/.github/workflows/generate_api_docs.yaml
@@ -22,6 +22,7 @@ jobs:
         uses: conda-incubator/setup-miniconda@v3
         with:
           environment-file: conda-store-server/environment-dev.yaml
+          conda-remove-defaults: "true"
 
       - name: "Install conda-store-server 📦"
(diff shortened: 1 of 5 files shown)


peytondmurray (approved):
Looks good to me.

I saw your comment on the target issue - thanks for pointing out that attestations are already being generated. My one question: now we have two different sets of attestations, right? One that is produced by GH and the other that is produced by `pypa/gh-action-pypi-publish` - should we just stick to one, or there no harm in providing both?

trallard:
There is no harm in providing both, as they have _slightly_ different schemas/metadata so let's just go with it.

PR #980, opened 2024-11-14
Bump dev version

soapy1:
ref: https://github.com/conda-incubator/conda-store/issues/967

### DIFF
--- a/conda-store-server/conda_store_server/__init__.py
+++ b/conda-store-server/conda_store_server/__init__.py
@@ -14,7 +14,7 @@ if typing.TYPE_CHECKING:
     from .app import CondaStore
 
 
-__version__ = "2024.11.1"
+__version__ = "2024.11.2-dev"
 
 
 CONDA_STORE_DIR = platformdirs.user_data_path(appname="conda-store")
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -43,6 +43,8 @@ jobs:
       - name: "Build and check package - ${{ matrix.directory }} 📦"
         uses: hynek/build-and-inspect-python-package@v2
         id: baipp
+        permissions:
+          id-token: write # required for attest-build-provenance-github
```
peytondmurray, line 47: I think we don't need this, as it is set on L20 above.

peytondmurray (changes requested):
Small change requested!

.github/workflows/release.yaml
soapy1: without this addition we get this error https://github.com/conda-incubator/conda-store/actions/runs/11843467867/job/33004483291

peytondmurray:
Merging as it seems the permissions set here don't apply to forks.

peytondmurray (approved):

Issue #995, opened 2024-11-20
[BUG] - getting artifacts from the ui returns not found error


soapy1:
### Describe the bug

conda-store `build/{build_id}/<build artifact>` endpoint returns a not found error. 

For example:
  * for a valid build `67`, going to the endpoint `/api/v1/build/67/logs/` should return the logs for the build. 
  * or, the endpoint `/api/v1/build/67/lockfile/` should return the lockfile for the build. 
  
 However, in both cases conda-store returns the message `{"detail":"Not Found"}`

Looks like this bug is maybe related to https://github.com/conda-incubator/conda-store/pull/932

### Expected behavior

When you initiate a build thru the ui, some links to artifacts related to the build pop up at the bottom of the page. For example:
![Image](https://github.com/user-attachments/assets/c1ba7af0-0f84-48ef-ab99-266f70a9e2d8)

When you click one of the links, conda-store should send a request to the `/build/<build-id>/<>` endpoint, for example `http://localhost:8080/conda-store/api/v1/build/35/lockfile/`. And then redirect to the object storage url, for example `http://localhost:9000/conda-store/logs/db2456c0-1732135951-35-sdsd.log?...more args...`.
![Image](https://github.com/user-attachments/assets/6bd4b24b-7991-4bc5-9a1f-9bfa8e3ee1ab)

### How to Reproduce the problem?

1. Create a new build in the ui
2. Wait for the `Logs and Artifacts` section to be populated
3. Click one of the links in the `Logs and Artifacts` section
4. observe, the error message  `{"detail":"Not Found"}` opposed to the logs or artifacts

gabalafou:
I'm looking into this now. I have noticed both the backend (#997) and frontend (conda-incubator/conda-store-ui#442) PRs for this.

PR #997, opened 2024-11-21
Ensure api url setting has trailing slash

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/995

## Description

This is required for the "log and artifact" links in the ui to work. 

For example, follow the instructions in #995 to reproduce the issue. With this change, the links provided in the `Logs and Artifacts` section correctly point to `/conda-store/api/v1/<rest of the api path>`.

## Pull request checklist

- x Did you update the documentation (if required)?
- x Did you add/update relevant tests for this change (if required)?

### DIFF
--- a/conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html
+++ b/conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html
@@ -22,7 +22,7 @@ https://conda.store/conda-store-ui/how-tos/configure-ui -->
              REACT_APP_AUTH_TOKEN: "",
              REACT_APP_STYLE_TYPE: "green-accent",
              REACT_APP_SHOW_LOGIN_ICON: "true",
-             REACT_APP_API_URL: "{{ url_for('redirect_root_to_ui') }}",
+             REACT_APP_API_URL: "{{ url_for('redirect_root_to_ui') }}/",
              REACT_APP_LOGIN_PAGE_URL: "{{ url_for('get_login_method') }}?next=",
              REACT_APP_LOGOUT_PAGE_URL: "{{ url_for('post_logout_method') }}?next=/",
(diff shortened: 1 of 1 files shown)


conda-store-server/conda_store_server/_internal/server/templates/conda-store-ui.html
```diff
@@ -22,7 +22,7 @@
              REACT_APP_AUTH_TOKEN: "",
              REACT_APP_STYLE_TYPE: "green-accent",
              REACT_APP_SHOW_LOGIN_ICON: "true",
-             REACT_APP_API_URL: "{{ url_for('redirect_root_to_ui') }}",
+             REACT_APP_API_URL: "{{ url_for('redirect_root_to_ui') }}/",
```
soapy1, line 25: I'm not sure if there is a better way to do this, like maybe on the ui side? The rest of the interactions between the ui and api seem to work properly with and without this change.

peytondmurray:
Looks like this is not being tested, otherwise we would have caught this. Would it be possible to write a test to hit these logs/artifacts endpoints with and without trailing slashes?

soapy1:
> Would it be possible to write a test to hit these logs/artifacts endpoints with and without trailing slashes?

Thinking about this issue more, I think it would be better for this issue to be fixed in conda-store-ui opposed to conda-store. 

The crux of the issue is that the logs/artifacts in the ui are links to some api endpoint. Without this slash the `url_prefix` is ignored for building these links. But, the rest of the api interactions between conda-store and conda-store-ui work without this trailing slash. So, that makes me think that there is an issue with how the ui is generating the links. 

I took a quick look at the ui code, but I don't really understand how it works so I could be really off base here. @peytondmurray @gabalafou if you have some time could you take a look?

soapy1:
ok so, I think this fixes things on the ui side https://github.com/conda-incubator/conda-store-ui/pull/442 and this PR can be closed.

Issue #969, opened 2024-11-08
[MAINT] - Improve release process


soapy1:
### Context

Currently the[ release process for conda-store](https://github.com/conda-incubator/conda-store/blob/main/.github/ISSUE_TEMPLATE/new-release.md) has a lot of manual tasks. It would be nice to automate some of the manual steps so that the release process is more streamlined. An ideal release process might looks something like

1. check out the main branch, build and test locally
2. create a tag in github for the release

Here, step 2 should do:
* all the required version bumping for conda-store
* pull the latest version number for conda-store-ui
* run the automated tests
* build and push all the relevant artifacts

### Value and/or benefit

Makes cutting a release take less engineering time

peytondmurray:
> all the required version bumping for conda-store

I'd vote for either `hatch-vcs` or `setuptools_scm` for versioning. Manually setting version strings is both unnecessary and is a recipe for human error, since they are tied precisely to the git tag we have most recently released. I know others feel differently about this, but I haven't yet been convinced that there's a reason not to use automated versioning.

In the release issue template for `conda-store`, there's a note about needing to make another PR to bump the `conda-store` and `conda-store-server` versions to the next dev-release number. This will not be needed if we use automated versioning.

---

We should also have a workflow that triggers on successful release publication which bumps the `conda-store-server` version in the [conda-store-ui compose file](https://github.com/conda-incubator/conda-store-ui/blob/main/docker-compose.yml).

trallard:
While I am generally in favour of automating redundant manual steps I am -1 on fully automating the release of conda-store since I do not believe we have good enough confidence or coverage (not only of unit tests but also functional and integration tests) on our test suite and QA processes.

> Here, step 2 should do:

> all the required version bumping for conda-store

I started working on this in https://github.com/conda-incubator/conda-store/pull/861, though I did not complete this as I had to shift to other more urgent tasks and fixes. 
TBH, some of the complexity comes from the fact that we are necessarily maintaining and releasing two packages. So, IMO, we should work on https://github.com/conda-incubator/conda-store/issues/911. This will reduce some of the maintenance and release toil (we use this as a mono repo and are separating the CLI from the API unnecessarily) and reduce the cognitive load on users.

We could see https://github.com/conda-incubator/conda-store/pull/861 to the end as this will avoid needing to bump the version to a development one through a PR (which was not being done until recently when I added this step to the release process), and would also fix the currently broken step of releasing `dev` versions to test-pypi on merge to `main`.

[...]

PR #998, opened 2024-11-21
[MAINT] - Don't keep `_version.py` file in version control

soapy1:
related to https://github.com/conda-incubator/conda-store/issues/969

As a pre req to taking up https://github.com/conda-incubator/conda-store/pull/861, clean up the `_version.py` file which should not be tracked in version control.


peytondmurray:
I like this, but is anyone using `conda_store_server.__version__` right now? I guess as a workaround we can do something like

```python
# conda_store_server/__init__.py

from importlib.metadata import version

__version__ = version(__package__)
```

if we need to...?

soapy1:
The conda store version is meant to come from here https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/__init__.py#L17

As we move towards using setuptools-scm (and hatch-vcs), that'll look more like https://github.com/conda-incubator/conda-store/pull/861/files#diff-48a7f6b1b439ff32ff7b62092a2d6a49fb3c672e387714f659b6b492d22a3375R27.

Also want to call out this comment in the file
>file generated by setuptools_scm
 don't change, don't track in version control

peytondmurray:
Oh you're right - let's get rid of this!

peytondmurray (approved):

Issue #941, opened 2024-10-31
Remove docker related build tasks


soapy1:
### Context

Conda-store was once able to build docker images. This feature has not been working for some time, and is not currently being used. It would be nice to remove this feature to reduce cruft in the code base. If it becomes clear that it would be helpful to have this feature again, it should be re-implemented with a more modern approach.

### Value and/or benefit

The goal of this is to remove cruft and simplify the conda-store app.

soapy1:
@peytondmurray do you have opinions about if this should go thru a proper deprecation cycle, or if it's ok to yank out? It looks like this is not really in a working state, so I'd opt to yank it.

Issue #942, opened 2024-10-31
Remove docker registry


soapy1:
### Context

Conda-store was once able to build docker images and host them in it's own docker registry. The docker workflows for conda store have fallen into disrepair and is have not been used by anyone in some time.  It would be nice to remove this feature to reduce cruft in the code base. If it becomes clear that it would be helpful to have this feature again, it should be re-implemented with a more modern approach.

### Value and/or benefit

The goal of this is to remove cruft and simplify the conda-store app.

related to https://github.com/conda-incubator/conda-store/issues/941

soapy1:
@peytondmurray do you have opinions about if this should go thru a proper deprecation cycle, or if it's ok to yank out? I'm not sure anyone is using it, but the feature does still seem to work when I tested in locally. I'd be ok with deprecating and removing over a few releases.

peytondmurray:
@trallard @kcpevey @Adam-D-Lewis @krassowski Bringing this to your attention - any concerns about removing this iteration of the docker feature?

Adam-D-Lewis:
I'm not aware of anyone using it, but could we at least document what seems to be going wrong at the moment?  I wasn't aware that it no longer was working.

soapy1:
@Adam-D-Lewis I believe the registry is working, but building docker images has been disabled https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/_internal/worker/build.py#L423, looks like for the past year. I think if conda-store isn't building docker images, there isn't really too much use for a registry.

peytondmurray:
Looks like it never really went through a formal deprecation cycle anyway, it just got turned off. But it looks like it should have been generating warning messages about the feature not working...so maybe it's fine to remove at this point?

PR #1001, opened 2024-11-25
Remove docker build tasks

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/941

## Description

This PR removes actions for building docker images and related functionality, including:
* removing the action `action_generate_conda_docker`
* removing docker artifacts from the list of artifacts generated by default
* remove docker artifacts from artifact queries
* deprecate the api endpoint `/build/<build id>/docker`
  * this endpoint will continue to work, but will have an added header `deprecation: true` 
  * https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-deprecation-header

We'll leave in the docker registry functionality. This requires the following docker related bits to stay (to be removed in https://github.com/conda-incubator/conda-store/issues/942)
* the `DOCKER_` build artifact types
* all docker db related items

## Pull request checklist


conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -1341,14 +1340,15 @@ async def api_get_build_archive(
         return RedirectResponse(conda_store.storage.get_url(build.conda_pack_key))
 
 
-@router_api.get("/build/{build_id}/docker/")
+@router_api.get("/build/{build_id}/docker/", deprecated=True)
```
peytondmurray, line 1343: :+1: Thanks for making sure this appears in both the docs and the response.

peytondmurray (approved):
Let's go! :rocket:

PR #1002, opened 2024-11-26
REL - 2024.11.2

soapy1:
## Pull request checklist


gabalafou (approved):

gabalafou:
I guess it’s okay that the server and UI versions are not the same?

soapy1:
> I guess it’s okay that the server and UI versions are not the same?

yep. We did a release of the server earlier in the month. So the next version should be `2024.11.2`.

Issue #994, opened 2024-11-19
[MAINT]  - Remove redundant column "channel" in `conda_package_build` table


soapy1:
### Context

Follows up https://github.com/conda-incubator/conda-store/pull/961

Currently, the `conda_package_build` table has a column for the channel the package comes from (ref: https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/_internal/orm.py#L786). However, this is redundant information because each entry must also have a `conda_package` entry associated to it, which already has the channel information (https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/_internal/orm.py#L743). 

This duplication of information can lead to inconsistencies, for example the bug resolved by #961.

### Value and/or benefit

This change will simplify the database relations.

PR #961, opened 2024-11-07
Update conda-package-build causes an integrity error on conda_package_build table

soapy1:
related to https://github.com/conda-incubator/conda-store/issues/852

## Description

Previously, whenever the packages for a channel would get updated, duplicated packages across channels would cause an sql error.

```
conda-store-worker-1  | (Background on this error at: https://sqlalche.me/e/14/gkpj)
conda-store-worker-1  | [2024-11-07 05:27:06,310: ERROR/ForkPoolWorker-5] Task task_update_conda_channel[73d1f6c4-e39e-4d77-a73d-0a61c5824648] raised unexpected: IntegrityError('(psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "_conda_package_build_uc"\nDETAIL:  Key (channel_id, package_id, subdir, build, build_number, sha256)=(3, 5618, linux-64, py27_0, 0, d65b4ca2fa3e2e9f6dc84b4883255c395a381bb11a3009c2f6fefa24aa284db8) already exists.\n')
conda-store-worker-1  | Traceback (most recent call last):
conda-store-worker-1  |   File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/sqlalchemy/engine/base.py", line 1890, in _execute_context
conda-store-worker-1  |     self.dialect.do_executemany(
conda-store-worker-1  |   File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/sqlalchemy/dialects/postgresql/psycopg2.py", line 982, in do_executemany
conda-store-worker-1  |     context._psycopg2_fetched_rows = xtras.execute_values(
conda-store-worker-1  |                                      ^^^^^^^^^^^^^^^^^^^^^
conda-store-worker-1  |   File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/psycopg2/extras.py", line 1299, in execute_values
conda-store-worker-1  |     cur.execute(b''.join(parts))
conda-store-worker-1  | psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "_conda_package_build_uc"
conda-store-worker-1  | DETAIL:  Key (channel_id, package_id, subdir, build, build_number, sha256)=(3, 5618, linux-64, py27_0, 0, d65b4ca2fa3e2e9f6dc84b4883255c395a381bb11a3009c2f6fefa24aa284db8) already exists.
conda-store-worker-1  | 
conda-store-worker-1  | 
[...]

Issue #733, opened 2024-01-16
Test migrations as part of CI


nkaretnikov:
### Context

One thing that takes a lot of time right now is checking that existing installations are not broken. This is currently done by hand, which takes a lot of time and is error-prone.

_The goal of this issue is to shift responsibility from the PR author to the testing infrastructure._

It needs to be done as part of CI instead. It could work something like this:
1. Run the whole testsuite on main to _populate the DB with realistic data_ (this is the important part -- to make sure that data is there) and make sure the tests pass (both SQLite and Postgres)
2. _With the populated DB_, run the update migration added in the new PR and run the whole testsuite again + new tests added in the new PR (to make sure the migration is successful and new data is not conflicting with the old data/schema)
3. Then run the downgrade migration with the DB added on step 2 and make sure that tests from main pass.
4. During merges, the CI needs to check if there are issues with migrations (migrations have hashes and if two PRs both add migrations, one of them needs to be updated -- the CI needs to detect that and run the testsuite again). If there are problems, the PR should be reverted automatically.

### Value and/or benefit

Goals:
- After this issue is implemented and closed, _we stop relying on manual testing_ because it's too slow and error-prone anyway.
- If the CI is green, it's considered tested and fine. _The author of the PR is not responsible if there are any issues._
- If problems are discovered later, we add new testing infrastructure.
- _At no point we go back to manual testing._

This will improve both the quality and the development/iteration speed, which will translate into getting more features in after this initial time investment.

### Anything else?

Our unit-tests currently run only on SQLite. We need to add a Postgres DB that will execute the same tests.

conda-store can be configured to use postgres like this:

```
c.CondaStore.database_url = 'postgresql+psycopg2://postgres:password@localhost:5432/conda-store'
```

It first requires:
- configuring Postgres to use the password `password` listed above
- adding the `conda-store` database to Postgres.

Note that we do have API tests that use Docker/Postgres, but that's different because not everything can be tested via those APIs.

Both SQLite and Postgres need to be tested via unit tests because these DBs have some incompatibilities.

For example:
- SQLite doesn't have Enum support and represents Python ORM Enums as VARCHAR
- Postgres has Enums, but they are immutable, which requires making a temporary table if you want to add a new field and copying the data
- SQLite doesn't have ALTER COLUMN support
- SQLite doesn't warn on casts to invalid/unknown types.
- etc.

trallard:
[...]

PR #1010, opened 2024-12-05
Remove redundant channel column

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/994

## Description

The primary focus of this pr is to remove the `channel_id` column and `channel` association in the `conda_package_build` table. This is redundant information, also available in the associated `package`. In order to get channel information from a `conda_package_build` users may join the `conda_package` table and `conda_package_build` table. For example:
```
from conda_store_server._internal import orm

db.query(orm.CondaPackageBuild)
        .join(orm.CondaPackage)
        .filter(orm.CondaPackage.channel_id == 1)
```

Things changed in this PR:
* removal of the `channel_id` column in the `conda_package_build` table
* a migration script. It does 2 things:
  *  updates the `conda_package_build` to have the correct package associated with each row. See #961 for more details about this issue
  * updates the database table column + unique constraint
* updates the docs for how to write a migration
* adds a dependency on [pytest-alembic
](https://pytest-alembic.readthedocs.io/en/latest/running.html)
* sets up running tests for alembic migrations

## Pull request checklist

## Additional Info

Part of the migration is to correct some of the conda_package_build entries that have the wrong conda_package associated with them. Since this needs to sort thru all elements of this table, this migration takes a while to run. Users should probably opt to run this migration directly with alembic.

### DIFF
--- a/conda-store-server/conda_store_server/_internal/alembic/script.py.mako
+++ b/conda-store-server/conda_store_server/_internal/alembic/script.py.mako
@@ -1,6 +1,6 @@
-// Copyright (c) conda-store development team. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
+# Copyright (c) conda-store development team. All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
 
 """${message}
 
--- a/conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
+++ b/conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
@@ -1,0 +1,189 @@
+# Copyright (c) conda-store development team. All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
+
+import pytest
+from sqlalchemy import text
+from sqlalchemy.exc import OperationalError
+
+from conda_store_server import api
+from conda_store_server._internal import orm
+
+
+def setup_bad_data_db(conda_store):
+    """A database fixture populated with
+    * 2 channels
+    * 2 conda packages
+    * 5 conda package builds
+    """
+    with conda_store.session_factory() as db:
+        # create test channels
+        api.create_conda_channel(db, "test-channel-1")
+        api.create_conda_channel(db, "test-channel-2")
+        db.commit()
+
+        # create some sample conda_package's
+        # For simplicity, the channel_id's match the id of the conda_package.
+        # So, when checking that the package build entries have been reassembled
+        # the right way, check that the package_id in the conda_package_build is
+        # equal to what would have been the channel_id (before the migration is run)
+        conda_package_records = [
+            {
+                "id": 1,
+                "channel_id": 1,
+                "name": "test-package-1",
+                "version": "1.0.0",
+            },
+            {
+                "id": 2,
+                "channel_id": 2,
+                "name": "test-package-1",
+                "version": "1.0.0",
+            },
+        ]
+        for cpb in conda_package_records:
+            conda_package = orm.CondaPackage(**cpb)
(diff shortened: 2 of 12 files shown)


peytondmurray:
At least partially address https://github.com/conda-incubator/conda-store/issues/733 with the addition of `pytest-alembic`.

conda-store-server/conda_store_server/_internal/alembic/script.py.mako
```diff
@@ -1,6 +1,6 @@
-// Copyright (c) conda-store development team. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
+# Copyright (c) conda-store development team. All rights reserved.
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file.
```
peytondmurray, line 3: Thanks, this has been breaking new migrations I was adding for another PR :+1:

conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def test_remove_conda_package_build_bad_data(
+    conda_store, alembic_config, alembic_engine, alembic_runner
+):
+    """Simply run the upgrade and downgrade for this migration"""
+    # migrate all the way to the target revision
+    alembic_runner.migrate_up_to("89637f546129")
+
+    # try downgrading
```
peytondmurray, line 139: Since the goal of the migration to delete the `channel_id` off the `conda_package_build` table, it might be good to assert that there's no `channel_id` column in that table here...

conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
```diff
@@ -0,0 +1,183 @@
...
+    alembic_runner.migrate_down_one()
+
+    # seed db with data that has broken data
+    setup_bad_data_db(conda_store)
```
peytondmurray, line 143: ...and that there is a `channel_id` in the table here.

conda-store-server/tests/conftest.py
```diff
@@ -290,6 +290,20 @@ def plugin_manager():
...
+    from conda_store_server._internal.dbutil import write_alembic_ini
+
+    ini_file = pathlib.Path(__file__).parent / "alembic.ini"
+    write_alembic_ini(ini_file, conda_store.database_url)
+    return {"file": ini_file}
+
+
+@pytest.fixture
+def alembic_engine(db):
+    return db
```
peytondmurray, line 304: I see the `pytest-alembic` docs mention something about defining an `alembic_engine`:

> Override this fixture to provide pytest-alembic powered tests with a database handle.

But this function doesn't seem to do anything. Do the tests run without this? If not, it would be helpful to have a docstring here about it being necessary to avoid future maintainer confusion.

peytondmurray (approved):
This one seemed like a bit of a tricky migration. The changes look great, thank you for this! Only a few minor suggestions here.

conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
soapy1: good idea!

conda-store-server/tests/_internal/alembic/version/test_89637f546129_remove_conda_package_build_channel.py
soapy1: can do!

conda-store-server/tests/conftest.py
soapy1: good eye, tests continue to work without this. Will remove it :x:

PR #1011, opened 2024-12-06
Fix conda_package_build "build" column type

soapy1:
## Description

This should be a string instead of an int, for example "py310h06a4308_0". This was probably updated to the wrong type by accident as part of https://github.com/conda-incubator/conda-store/pull/970

## Pull request checklist


peytondmurray (approved):
Thanks for this - did the problematic commit get released?

soapy1:
> Thanks for this - did the problematic commit get released?

yep, should have. Though it doesn't seem to have broken anything.

PR #1013, opened 2024-12-09
Fix deprecation warnings

soapy1:
## Description

There are currently lots of deprecation warning when running the test suite. Some of them come from upstream dependencies. But many are from conda-store. This PR aims to resolve some of them.

not fixed:
* `datetime.datetime.utcnow` is deprecated, should be replaced with `datetime.datetime.now(datetime.UTC)`. However, this is not supported by python 3.10. Will address this when we drop python 3.10 support

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/_internal/environment.py
+++ b/conda-store-server/conda_store_server/_internal/environment.py
@@ -14,7 +14,7 @@ from conda_store_server._internal import conda_utils, orm, schema, utils
 
 def validate_environment(specification):
     try:
-        specification = schema.CondaSpecification.parse_obj(specification)
+        specification = schema.CondaSpecification.model_validate(specification)
         return True
     except pydantic.ValidationError:
         return False
--- a/conda-store-server/conda_store_server/_internal/plugins/lock/conda_lock/conda_lock.py
+++ b/conda-store-server/conda_store_server/_internal/plugins/lock/conda_lock/conda_lock.py
@@ -40,7 +40,7 @@ class CondaLock(lock.LockPlugin):
         lockfile_filename = pathlib.Path.cwd() / "conda-lock.yaml"
 
         with environment_filename.open("w") as f:
-            json.dump(spec.dict(), f)
+            json.dump(spec.model_dump(), f)
 
         context.log.info(
             "Note that the output of `conda config --show` displayed below only reflects "
--- a/conda-store-server/conda_store_server/_internal/schema.py
+++ b/conda-store-server/conda_store_server/_internal/schema.py
@@ -443,7 +443,7 @@ class LockfileSpecification(BaseModel):
     def __str__(self):
         # This makes sure the format is suitable for output if this object is
         # converted to a string, which can also happen implicitly
-        return str(self.dict())
+        return str(self.model_dump())
 
 
 ###############################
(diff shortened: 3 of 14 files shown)


trallard:
I did not go through all the changes in detail but this is much appreciated.

> datetime.datetime.utcnow is deprecated, should be replaced with datetime.datetime.now(datetime.UTC). However, this is not supported by python 3.10. Will address this when we drop python 3.10 support

Can you add a TODO note on this in the code so we do not forget to do this when needed

conda-store-server/conda_store_server/_internal/orm.py
```diff
@@ -173,6 +173,7 @@ def __init__(self, specification, is_lockfile: bool = False):
     name: Mapped[str] = mapped_column(Unicode(255), nullable=False)
     spec: Mapped[dict] = mapped_column(JSON, nullable=False)
     sha256: Mapped[str] = mapped_column(Unicode(255), unique=True, nullable=False)
+    # TODO: change to datetime.datetime.now(datetime.UTC) when python 3.10 is dropped
```
peytondmurray, line 176: Can we make an issue for this instead of inline TODOs? They have a habit of getting lost in the repo if there aren't issues for them...

To mitigate datetime deprecation warnings, we could do something like this at the top of the file (or in a `utils.py`) so that the correct function is called on whatever version is currently running:

```python
import sys

if sys.version_info[:2] > (3, 10):
    from datetime.datetime import now
else:
    from datetime.datetime import utcnow as now
```

peytondmurray (approved):
See comment below about python 3.10/utcnow issue. I vote for a github issue and the suggested snippet instead of TODOs everywhere, but I'd be okay if you wanted to merge this as is too.

conda-store-server/conda_store_server/_internal/orm.py
soapy1: ooo, I like the suggestion of making this an issue opposed to TODO's. 

For the deprecation warnings, I don't think this snippet is going to quite work. [`datetime.datetime.now`](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) will use the default TZ provided by the machine it's running on if it's not given the timezone parameter. And `utcnow` does not accept any arguments. So on my machine (I'm on Pacific time) I get 
```
>>> datetime.datetime.now()
datetime.datetime(2024, 12, 16, 12, 10, 50, 437991)

>>> datetime.datetime.utcnow()
. . .  deprecation warning removed for brevity . . .
datetime.datetime(2024, 12, 16, 20, 11, 2, 237286)

>>> datetime.datetime.utcnow(datetime.UTC)
  File "<stdin>", line 1, in <module>
TypeError: datetime.utcnow() takes no arguments (1 given)
```

peytondmurray: Ahh, I guess it would have to be something like

```python
import sys

if sys.version_info[:2] > (3, 10):
    from datetime.datetime import now
else:
    import datetime
    def now():
        return datetime.datetime.utcnow(datetime.UTC)
```

But I'm fine with just merging as is if that's preferable.

soapy1: Cool, I'm going to go for a merge so we have less work in the future when we can drop python 3.10

PR #1015, opened 2024-12-10
Run unit tests on all supported python versions

soapy1:
Currently we run integration tests on all supported python versions. We should do the same with unit tests.

### DIFF
--- a/.github/workflows/test_conda_store_server_unit.yaml
+++ b/.github/workflows/test_conda_store_server_unit.yaml
@@ -46,12 +46,13 @@ jobs:
           hatch env run -e lint lint
 
   test-conda-store-server:
-    name: "unit-test - ${{ matrix.os}} "
+    name: "unit-test - ${{ matrix.python-version }} "
     strategy:
       matrix:
         # ubuntu 22.04, macos 14, windows 2022
         # os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         os: ["ubuntu-latest"]
+        python-version: ["3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.os }}
     defaults:
       run:
@@ -61,10 +62,6 @@ jobs:
       - name: "Checkout Repository 🛎"
         uses: actions/checkout@v4
 
-      - name: "Get project's default Python version 🏷️"
-        run: |
-          echo "PYTHON_VERSION_DEFAULT=$(cat .python-version-default)" >> $GITHUB_ENV
-
       - name: "Set up env ${{ matrix.os }} 🐍"
         uses: conda-incubator/setup-miniconda@v3
         with:
@@ -72,7 +69,7 @@ jobs:
             miniforge-version: latest
             auto-activate-base: false
             activate-environment: conda-store-server-dev
-            python-version: ${{ env.PYTHON_VERSION_DEFAULT }}
+            python-version: ${{ matrix.python-version }}
             conda-remove-defaults: "true"
 
       # This fixes a "DLL not found" issue importing ctypes from the hatch env
@@ -82,7 +79,7 @@ jobs:
           timeout_minutes: 9999
           max_attempts: 6
           command:
-            conda install --channel=conda-forge --quiet --yes python=${{ env.PYTHON_VERSION_DEFAULT }}
+            conda install --channel=conda-forge --quiet --yes python=${{ matrix.python-version }}
         if: matrix.os == 'windows-latest'
 
       - name: "Install conda-store server 📦"


.github/workflows/test_conda_store_server_unit.yaml
```diff
@@ -46,12 +46,13 @@ jobs:
...
 
   test-conda-store-server:
-    name: "unit-test - ${{ matrix.os}} "
+    name: "unit-test - ${{ matrix.python-version }} "
     strategy:
       matrix:
         # ubuntu 22.04, macos 14, windows 2022
         # os: ["ubuntu-latest", "macos-latest", "windows-latest"]
         os: ["ubuntu-latest"]
+        python-version: ["3.10", "3.11", "3.12"]
```
trallard, line 55:
```suggestion
        python-version: ["3.10", "3.11", "3.12", "3.13"]
```

Let's add 3.13 so we start testing against this version.
Note we might want to add this to all our actions actually

soapy1: conda [does not currently support python 3.13](https://github.com/conda/conda/pull/14187). So, this[ produces an error in the setup stage for ci ](https://github.com/conda-incubator/conda-store/actions/runs/12281008876/job/34268821539?pr=1015)

soapy1: I was investigating this a little further, and I don't think we will be able to run test against python 3.13 until conda is installable in a python 3.13 environment. 

The step the ci is stuck on is `Updating 'conda-store-server-dev' env from conda env update...` this step will install the conda environment specified in `conda-store-server/environment.yaml`. The requested python version will get patched into the env specification, so it'll try to install an environment that looks like:

```
name: conda-store-server-dev
channels:
  - conda-forge
dependencies:
  - python >=3.10
  - conda
  - aiohttp>=3.8.1
  - pytest
  - pytest-celery
  - pytest-mock
  - pytest-cov
  - docker-py<=7
  - docker-compose
  - hatch
  - twine>=5.0.0
  - pkginfo>=1.10 # needed to support metadata 2.3
  - constructor # runtime dep but must be conda installed
  - pre_commit
  - pip
  - pip:
      - pytest-playwright
      - build
  - python=3.12
```

You can verify that installing this environment locally fails in the same way that it fails in ci. Solver messaging is a little confusing to read (for me), but I think these 2 pieces of info are most important in this case. 

This piece of the error message shows the requested version of python that is trying to be installed. Obviously, the env files requested python 3.13, so that's what it'll try to install. 
```
 python 3.13.*,>=3.10  is not installable because there are no viable options
     ├─ python [3.13.0|3.13.1] conflicts with any installable versions previously reported;
     └─ python [3.13.0rc1|3.13.0rc2|3.13.0rc3] would require
        └─ _python_rc, which does not exist (perhaps a missing channel).
```

There are a bunch of these lines that show different versions of conda with different python requirements. Here is one of those lines, showing some conda's that require python 3.12. If you look at all of these lines, there is no version that provides a version with python 3.13. So this is unsatisfiable.
```
 └─ conda [23.10.0|23.11.0|...|24.9.2] would require
  │     └─ python_abi 3.12.* *_cp312, which requires
  │        └─ python 3.12.* *_cpython, which can be installed;
```

### Exploring the issue

#### update pyproject.toml

I found that changing the `pyproject.toml` python version does not change the output. This is expected, since the issue arises before a step that uses the `pyproject.toml`.

#### add channel with python_rc 

I also tried adding the channel ` conda-forge/label/python_rc`. So, for example this env file:
```
name: conda-store-server-dev-2
channels:
  - conda-forge
  - conda-forge/label/python_rc
dependencies:
  # must be kept in sync with the min supported version in pyproject.toml
  - python >=3.10
  # conda builds
  - conda
  # dev dependencies
  - aiohttp>=3.8.1
  - pytest
  - pytest-celery
  - pytest-mock
  - pytest-cov
  - docker-py<=7
  - docker-compose
  # build dependencies
  - hatch
  - twine>=5.0.0
  - pkginfo>=1.10 # needed to support metadata 2.3
  # installer
  - constructor # runtime dep but must be conda installed
  # linting
  - pre_commit
  - pip
  - pip:
      - pytest-playwright
      - build
  - python=3.13
```
As expected, this fails because of the unsatisfiable condition shown above.

#### removing dependencies

To be thorough, I also tried to remove `conda` and `constructor` as dependencies in the `conda-store-server/environment.yaml`. This allowed the CI to successfully setup the tests and run the test suite. However, as expected, lots of the tests fail because dependencies are missing ([link to ci run](https://github.com/conda-incubator/conda-store/actions/runs/12302934584/job/34336899081)).

trallard: This confused me since they have the same pins as us, and the conda outputs can be rather annoying to read/debug.

It seems we will have to wait until constructor and conda sort their 3.13 releases for this then. Thanks for the thorough investigation @soapy1

trallard (approved):
Thanks @soapy1. I will go ahead and approve this 🚀

PR #1016, opened 2024-12-10
Deprecate docker registry

soapy1:
fixes https://github.com/conda-incubator/conda-store/issues/942

## Description

This aims to start the process of removing the `/registry` endpoint. It:
* disables the registry endpoint by default
* adds deprecation headers to all the endpoints under `/registry`
* target deprecation date March 17, 2025

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/_internal/server/views/api.py
+++ b/conda-store-server/conda_store_server/_internal/server/views/api.py
@@ -3,14 +3,14 @@
 # license that can be found in the LICENSE file.
 
 import datetime
-from typing import Any, Dict, List, Optional, Callable
+from functools import wraps
+from typing import Any, Callable, Dict, List, Optional
 
 import pydantic
 import yaml
 from celery.result import AsyncResult
 from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request
 from fastapi.responses import JSONResponse, PlainTextResponse, RedirectResponse
-from functools import wraps
 
 from conda_store_server import __version__, api
 from conda_store_server._internal import orm, schema
@@ -105,6 +105,7 @@ def paginated_api_response(
         "count": count,
     }
 
+
 def deprecated(sunset_date: datetime.date) -> Callable:
     """Decorator to add deprecation headers to a HTTP response. These will include:
         {
@@ -120,14 +121,19 @@ def deprecated(sunset_date: datetime.date) -> Callable:
     sunset_date : datetime.date
         the date that the endpoint will have it's functionality removed
     """
+
     def decorator(func):
         @wraps(func)
         def add_deprecated_headers(*args, **kwargs):
             response = func(*args, **kwargs)
             response.headers["Deprecation"] = "True"
-            response.headers["Sunset"] = sunset_date.strftime("%a, %d %b %Y 00:00:00 UTC")
+            response.headers["Sunset"] = sunset_date.strftime(
+                "%a, %d %b %Y 00:00:00 UTC"
+            )
             return response
+
         return add_deprecated_headers
+
     return decorator
 
 
@@ -1380,7 +1386,8 @@ async def api_get_build_docker_image_url(
                 "message": f"Build {build_id} doesn't have a docker manifest",
             }
             return JSONResponse(
-                status_code=400, content=content,
+                status_code=400,
+                content=content,
             )
 
 
--- a/conda-store-server/conda_store_server/_internal/server/views/api.py
+++ b/conda-store-server/conda_store_server/_internal/server/views/api.py
@@ -1353,7 +1353,7 @@ async def api_get_build_archive(
 
 
 @router_api.get("/build/{build_id}/docker/", deprecated=True)
-@deprecated(sunset_date=datetime.date(2025, 2, 17))
+@deprecated(sunset_date=datetime.date(2025, 3, 17))
 async def api_get_build_docker_image_url(
     build_id: int,
     request: Request,
--- a/conda-store-server/conda_store_server/_internal/server/views/api.py
+++ b/conda-store-server/conda_store_server/_internal/server/views/api.py
@@ -10,6 +10,7 @@ import yaml
 from celery.result import AsyncResult
 from fastapi import APIRouter, Body, Depends, HTTPException, Query, Request
 from fastapi.responses import JSONResponse, PlainTextResponse, RedirectResponse
+from functools import wraps
 
 from conda_store_server import __version__, api
 from conda_store_server._internal import orm, schema
@@ -103,6 +104,26 @@ def paginated_api_response(
         "size": paginated_args["limit"],
         "count": count,
     }
+
+def deprecated(sunset_date):
+    """
+    Decorator to add deprecation headers to a response. These will include
+        {
+            Deprecation: True
+            Sunset: <sunset_date>
+        }
+
+    :param sunset_date: the sunset date in HTTP-Date format, for example "Mon, 16 Feb 2025 23:59:59 UTC"
+    """
+    def decorator(func):
+        @wraps(func)
+        def add_deprecated_headers(*args, **kwargs):
+            response = func(*args, **kwargs)
+            response.headers["Deprecation"] = "True"
+            response.headers["Sunset"] = sunset_date
+            return response
+        return add_deprecated_headers
+    return decorator
 
 
 @router_api.get(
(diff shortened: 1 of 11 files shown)


conda-store-server/conda_store_server/_internal/server/views/registry.py
```diff
@@ -79,6 +81,10 @@ def replace_words(s, words):
 
 def get_docker_image_manifest(conda_store, image, tag, timeout=10 * 60):
     namespace, *image_name = image.split("/")
+    response_headers = {
+        "Deprecation": "True",
+        "Sunset": "Mon, 16 Feb 2025 23:59:59 UTC",
+    }
```
peytondmurray, line 87: An alternative to this might be to make a `DeprecatedResponse` object which includes the correct response headers, and either wrap the current responses with it or make a decorator that does this automatically. That would help avoid typos in these keys, but I don't know if implementing deprecation machinery like that is worth the effort.

peytondmurray (approved):
Very nice, thank you!

conda-store-server/conda_store_server/_internal/server/views/registry.py
soapy1: Oh ya, I like this approach. I added just a decorator that adds the two deprecation headers. Could also extend this to create a `DeprecatedResponse` if you think that's more appropriate.

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -441,8 +441,7 @@ endpoints. Default True.
 `CondaStoreServer.enable_api` a Boolean on whether to expose the API
 endpoints. Default True.
 
-`CondaStoreServer.enable_registry` a Boolean on whether to expose the
-registry endpoints. Default True.
+`CondaStoreServer.enable_registry` (deprecated) a Boolean on whether to expose the registry endpoints. Default False.
```
trallard, line 444: Perhaps deprecated configuration options should be in a separate heading to keep things tidy.

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -453,7 +452,7 @@ to. The default is all IP addresses `0.0.0.0`.
 `CondaStoreServer.port` is the port for conda-store server to
 use. Default is `8080`.
 
-`CondaStoreServer.registry_external_url` is the external hostname and
+`CondaStoreServer.registry_external_url` (deprecated) is the external hostname and
```
trallard, line 455: Same comment about having a deprecated section

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -128,6 +129,20 @@ def paginated_api_response(
     }
 
 
+def deprecated(sunset_date):
```
trallard, line 132: It needs docstrings, it would be particularly helpful to document the expected format of the `sunset_date`)

trallard (changes requested):

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -128,6 +129,30 @@ def paginated_api_response(
...
+    """
+    Decorator to add deprecation headers to a response. These will include
+        {
+            Deprecation: True
+            Sunset: <sunset_date>
+        }
+
+    :param sunset_date: the sunset date in HTTP-Date format, for example "Mon, 16 Feb 2025 23:59:59 UTC"
```
peytondmurray, line 140: Small nit: the majority of the repo is using numpydoc docstring format. In the future I'd really like to enable some docstring lint rules to make this more uniform. No worries about updating this; as we continue to update the code base we'll get more serious about ensuring docstring uniformity and completeness in the future.

As an aside, it would also be nice to have type annotations here. But again, most of the code base isn't annotated, so maybe we can leave this for the future?

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -128,6 +129,30 @@ def paginated_api_response(
     }
 
 
+def deprecated(sunset_date):
```
peytondmurray, line 132: It might be good to accept a `datetime.datetime` object and then format it however you want inside this function. That moves the responsibility of ensuring the right format away from the person using the decorator, to the decorator itself.

peytondmurray (approved):

conda-store-server/conda_store_server/_internal/server/views/api.py
soapy1: thanks for pointing this out! Lets update it now

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -128,6 +129,34 @@ def paginated_api_response(
     }
 
 
+def deprecated(sunset_date: datetime.date):
+    """Decorator to add deprecation headers to a response. These will include
```
trallard, line 133:
```suggestion
def deprecated(sunset_date: datetime.date):
    """Decorator to add deprecation headers to a HTTP response. These will include:
```

trallard: Also we should add a link to the section in the docs re deprecations here.

docusaurus-docs/conda-store/explanations/artifacts.md
```diff
@@ -63,7 +63,7 @@ Click **"Download archive"** button to download the archive of your conda enviro
 
 To install the tarball, follow the [instructions for the target machine in the conda-pack documentation][conda-pack-usage].
 
-## Docker images
+## Docker images (deprecated)
```
trallard, line 67: Since I cannot comment on non-changed lines, we should replace the `warning` below to:

```
:::warning[Deprecation notice]

<Details on the deprecation>

:::
```

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -210,14 +210,16 @@ setting is useful if you want to protect environments from
...
 `CondaStore.post_update_environment_build_hook` is an optional configurable to
 allow for custom behavior that will run after an environment's current build changes.
 
 `CondaStore.lock_backend` is the name of the default lock plugin to use
 when locking a conda environment. By default, conda-store uses [conda-lock](https://github.com/conda/conda-lock).
 
+#### Deprecated configuration options for `conda_store_server._internal.app.CondaStore`
+
+`CondaStore.serialize_builds` no longer has any effect
+
```
trallard, line 222: Is there a reason why this is a H4 instead of a H3 like the rest of the configuration options? Being an H4 means this does not show in the in-page TOC like the rest of the config options.

Also, line 509 has `### (deprecated) `conda_store_server.registry.ContainerRegistry`, so we have inconsistent formatting; we should stick to one throughout.

docusaurus-docs/conda-store/references/configuration-options.md
```diff
@@ -486,7 +484,15 @@ to serve in form `[(path, method, function), ...]`. `path` is a
 string, `method` is `get`, `post`, `put`, `delete` etc. and function
 is a regular python fastapi function.
 
-### `conda_store_server.._internal.worker.app.CondaStoreWorker`
+#### Deprecated configuration options for `conda_store_server._internal.server.app.CondaStoreServer`
```
trallard, line 487: Same comment re heading formatting and level as above.

docusaurus-docs/conda-store/references/configuration-options.md
trallard, line 1: This is aside, but it would be good to address here. 
We skip H2 and go straight for H3 for the config option headers, we should change these to H2 instead and address the issues regarding inconsistent formatting.

trallard (changes requested):

conda-store-server/conda_store_server/_internal/server/views/api.py
```diff
@@ -133,6 +134,37 @@ def paginated_api_response(
     }
 
 
+def deprecated(sunset_date: datetime.date):
```
peytondmurray, line 137: Missing return type annotation, but I don't think this should be a blocker here.

peytondmurray (approved):
The changes look good - thanks for making that deprecation decorator.

trallard (approved):
Aside from the note regarding the deprecation policies this looks good to me, thanks @soapy1

Issue #911, opened 2024-10-22
[ENH] - consider consolidating conda-store and conda-server into one package


trallard:
### Feature description

We currently have two packages in this repo, `conda-store` and `conda-store-server`, where the former is but a convenience package for the conda-store CLI.

This increases complexity, and users should ideally get the CLI and UI-based versions installed simultaneously (since there are no scenarios in which users would only install the CLI and use it without the server).

We should consolidate the two in a single package.

### Value and/or benefit

- Reduced complexity (what is conda-store? What is conda-store-server? which makes it hard for users to grasp conda-store as a whole).
- Reduced maintenance overhead: we would only release and maintain one package

### Anything else?

We would need to think of how to communicate this change to users.

trallard:
@conda-incubator/conda-store team as we discussed this recently - it would be good to note any pros/cons/questions on your end.

peytondmurray:
A note about timeline and how this change will be implemented: @smeragoel is currently enumerating user workflows that are important to capture for the forthcoming CLI. @soapy1 and I will take a few days to provide input about these workflows. When CLI work begins, that will be a good opportunity for consolidation.

soapy1:
I've started looking into what needs to be done to make this happen. In particular, there are a few changes to be made:
* on the conda-store cli side
  * stop publishing the conda-store docker container
    *  there are a few conda-store docs/demos that depend on this container being available (and jupyter being set up in them)
  * remove all the `jupyter-*` dependencies from conda-store cli
* on the conda-store server side
  * it would be nice to be able to use the python api exposed in conda-store-server to de-duplicate code between the cli and public api
  * it would be nice to have a clear place for the cli module to live

A little bit of a rogue suggestion, but since we aren't really pushing any updates to the conda-store cli (until the new cli work begins) it might be nice to just drop/deprecate this package (stop building the conda-store packages and delete the `./conda-store` directory). 

Pros:
* we don't have to keep releasing this code that we are not making changes to
* users will still be able to use older version of the package 
* it will give use the opportunity to prepare the conda-store-server side of things to take a different approach at building the cli 

Cons:
* will not be able to provide bug fixes

soapy1:
Few more tasks that need to be done before we can close this issue:

- [ ] (1) remove the conda-store module
- [ ] (2) rename conda-store-server to conda-store + move the new conda-store module one level up in the file structure
[...]

PR #1019, opened 2024-12-10
Dont build conda store cli

soapy1:
part 2 to address https://github.com/conda-incubator/conda-store/issues/911

## Description

This PR removes steps for building and releasing conda-store (the cli). This includes:
* removing the split package from the conda build recipe - now the recipe will just build the conda-store-server
* Removing conda-store (cli) build tasks from the CI (this also removes the release build tasks)
* Removing conda-store integration tests from the CI

## Pull request checklist


.github/workflows/build_pypi_package.yaml
```diff
@@ -25,7 +24,6 @@ jobs:
     strategy:
       matrix:
         directory:
-          - "conda-store"
           - "conda-store-server"
     defaults:
```
trallard, line 28: If we are removing `conda-store`, it would be best to remove the `matrix` strategy altogether and refactor the workflow

.github/workflows/release.yaml
trallard, line 1: Same comment as in the previous workflow, we might as well remove the whole matrix strategy altogether

docusaurus-docs/conda-store/how-tos/install-standalone.md
```diff
@@ -51,10 +51,9 @@ Access conda-store UI at [`localhost:8080`](https://localhost:8080/) 🎉
...
 ## Next steps
 
-Learn to use the [conda-store UI with these tutorials][conda-store-ui-tutorials].
+Learn more about [conda-store ui][conda-store-ui].
 
 
 <!-- Internal links -->
 
 [conda-store-ui]: ../../conda-store-ui/introduction
-[conda-store-ui-tutorials]: ../../conda-store-ui/tutorials
```
trallard, line 60: Is there a reason to remove the tutorials?

trallard (changes requested):
Thank you @soapy1 this is going in the right direction,  I think since you are already going through removing the build for conda-store (cli) we might as well refactor the workflows to remove redundant matrix bits

trallard:
Another thought we would need to add a note somewhere about stopping releasing the cli as a separate package

docusaurus-docs/conda-store/how-tos/install-standalone.md
soapy1: oops, over edit. Thanks, will add these back.

soapy1:
> Another thought we would need to add a note somewhere about stopping releasing the cli as a separate package

There are a few public places that users will be able to see this change:
* this pr
* release notes (for the next release)
* pr to conda-forge package that removes the conda-store package

I think release notes is an appropriate place to be broadcasting this change to communicate with users that this change is coming. 

Are there any other stream of news that a conda-store user would be looking at to get this type of information?

peytondmurray (approved):
Looks great. Did the `recipe/meta.yaml` just come from the `conda-store-server` conda package?

peytondmurray (approved):

.github/workflows/build_docker_image.yaml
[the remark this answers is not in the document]
soapy1: I would say it would be better to remove all the conda-store directory at the same time opposed to piecemeal removing pieces.

.github/workflows/build_pypi_package.yaml
```diff
@@ -22,27 +21,22 @@ jobs:
...
-        working-directory: ${{ matrix.directory }}
+        working-directory: conda-store-server
     steps:
       - name: "Checkout Repository 🛎"
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - run: echo "Running on ${{ matrix.directory }}"
+      - run: echo "Running on conda-store-server"
```
trallard, line 33: This is now redundant, as it was only used to check it was actually building one of the sub-packages from the correct dir. Let's remove it.

.github/workflows/build_pypi_package.yaml
```diff
@@ -22,27 +21,22 @@ jobs:
...
 
-      - name: "Build and check package - ${{ matrix.directory }} 📦"
+      - name: "Build and check package - conda-store-server 📦"
```
trallard, line 35:
```suggestion
      - name: "Build and check package 📦"
```
name is redundant here too now

.github/workflows/release.yaml
```diff
@@ -20,28 +20,23 @@ jobs:
...
-        working-directory: ${{ matrix.directory }}
+        working-directory: conda-store-server
     steps:
       - name: "Checkout Repository 🛎"
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - run: echo "Running on ${{ matrix.directory }}"
+      - run: echo "Running on conda-store-server"
```
trallard, line 32:
```suggestion
```

.github/workflows/release.yaml
```diff
@@ -20,28 +20,23 @@ jobs:
...
 
-      - name: "Build and check package - ${{ matrix.directory }} 📦"
+      - name: "Build and check package - conda-store-server 📦"
```
trallard, line 34:
```suggestion
      - name: "Build and check package 📦"
```
name is redundant here too now

.github/workflows/release.yaml
trallard, line 1: There are a lot of places where we use `conda-store-server`, which might change as we consolidate the packages. I suggest setting a job-level env variable for this and reusing the variable instead.

README.md
```diff
@@ -14,7 +14,7 @@ Flexible. Reproducible. Governable.
...
-|Releases | [![PyPI release conda-store](https://img.shields.io/pypi/v/conda-store)](https://badge.fury.io/py/conda-store?label=pypi%20conda-store&style=flat.svg) [![PyPI release conda-store-server](https://img.shields.io/pypi/v/conda-store-server?label=pypi%20conda-store-server)](https://badge.fury.io/py/conda-store-server) [![conda-forge release conda-store](https://img.shields.io/conda/vn/conda-forge/conda-store?label=conda-forge%20conda-store)]((https://anaconda.org/conda-forge/conda-store)) [![conda-forge release conda-store-server](https://img.shields.io/conda/vn/conda-forge/conda-store-server?label=conda-forge%20conda-store-server)]((https://anaconda.org/conda-forge/conda-store-server)) |
+|Releases | [![PyPI release conda-store-server](https://img.shields.io/pypi/v/conda-store-server?label=pypi%20conda-store-server)](https://badge.fury.io/py/conda-store-server) [![conda-forge release conda-store-server](https://img.shields.io/conda/vn/conda-forge/conda-store-server?label=conda-forge%20conda-store-server)]((https://anaconda.org/conda-forge/conda-store-server)) |
```
trallard, line 17: I think we should add a note in the README indicating the last release similar to the one in the `conda-store/README`

docusaurus-docs/community/contribute/local-setup-core.md
```diff
@@ -8,7 +8,7 @@ description: Local development setup for conda-store
 There are two main ways to set up your local environment and conda-store services (web UI, API server, database, etc.) for development:
 
 - Using [Docker and Docker compose](#docker-setup-recommended): This is the recommended approach for working on `conda-store-server` library.
-- Using [standalone mode](#standalone-setup): Required for running tests and for working on the `conda-store` (client) library.
+- Using [standalone mode](#standalone-setup): This approach avoids using docker and starts the conda-store services as one process
```
trallard, line 11:
```suggestion
- Using [standalone mode](#standalone-setup): This approach avoids using Docker and starts the conda-store services as one process
```

docusaurus-docs/conda-store/introduction.md
```diff
@@ -6,18 +6,16 @@ description: Introduction to conda-store (core) documentation
 
 # conda-store (core)
 
-The [`conda-store` repository on GitHub][conda-store-repo] consists of two separate, yet related, packages:
+The [`conda-store` repository on GitHub][conda-store-repo] contains the conda-store application. It has two entrypoints
```
trallard, line 9:
```suggestion
The [`conda-store` repository on GitHub][conda-store-repo] contains the conda-store application. It has two entrypoints:
```

trallard (changes requested):

.github/workflows/release.yaml
soapy1: I think the amount of effort it will take to introduce a job-level variable will be pretty similar to the amount of effort it will take to explicitly rename things. Since (1) we should only be renaming `conda-store-server` -> `conda-store` once, and (2) it is easier to read + debug yaml files (and ci) that use explicit names opposed to variables, I would prefer not to introduce this variable. 

But open to doing it if I'm missing an important upside that it will provide. I think it would be appropriate to do that work in a follow up PR

Issue #1005, opened 2024-11-27
Rethink how settings get applied to conda-store servers/workers


soapy1:
### Context

Currently there are some issues with how conda store server/workers are configured and how settings are persisted. There are 2 places where settings come from:
* the configuration file (provided at application startup)
* the settings db table - contains only a subset of settings to apply globally or to a namespace/environment

However, it is not entirely clear which of these get precedent when there is overlap, or how changes to settings in one place affect the other.

This leads to unpredictable beheviour, for example when you are running the service in a cloud environment. Sometimes redeploying a worker with new settings does not work, etc.

It would be nice to have:
* a way to configure conda-store servers/workers with a yaml/config language type file
* a clear understanding of which settings take precedent from each config source
* a unified way of checking settings in conda-store

### Value and/or benefit

This change will:
* make it more clear to users where their config is coming from and how to change their deployment
* make the code for retrieving settings in conda-store-server more clear
* (hopefully) lay some ground work for enabling per environment/namespace settings

PR #1023, opened 2024-12-13
Separate conda-store app instance from config

soapy1:
pre req for resolving https://github.com/conda-incubator/conda-store/issues/1005

## Description

This PR refactors `conda_store_server/app.py` to split the config portion and app instance portion of the class. The goal of this refactor is more clearly delineate responsibility between the 2 classes/pieces of functionality. Eventually, this should also make how to approach changes to the config more clear.

2 new objects are introduced:
* `conda_store_config:CondaStore`
  * this is the config class for conda store
  * it uses traitlets to get config
  * it's only responsibility is to read the config file and provide that info to `conda_store:CondaStore`
* `conda_store:CondaStore`
  *  this has common app functions like `get_settings`, `set_settings`, etc
  * it gets it's configuration from `conda_store_config:CondaStore`
  * it's responsibility is to have shared instance information and provide common functionality

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/conda_store.py
+++ b/conda-store-server/conda_store_server/conda_store.py
@@ -3,26 +3,26 @@
 # license that can be found in the LICENSE file.
 
 import datetime
+import logging
 import os
 from contextlib import contextmanager
 from typing import Any, Dict
-import logging
 
 import pydantic
 from celery import Celery, group
 from sqlalchemy.orm import Session, sessionmaker
 from sqlalchemy.pool import QueuePool
 
-from conda_store_server import CONDA_STORE_DIR, api, storage, conda_store_config
-from conda_store_server._internal import conda_utils, orm, schema, utils, environment
-from conda_store_server.plugins import hookspec, plugin_manager
+from conda_store_server import CONDA_STORE_DIR, api, conda_store_config, storage
+from conda_store_server._internal import conda_utils, orm, schema, utils
 from conda_store_server.exception import CondaStoreError
+from conda_store_server.plugins import hookspec, plugin_manager
 from conda_store_server.plugins.types import lock
 
 
 class CondaStore:
     """This class provides a set of common functionality to be used by
-    conda store servers and workers. 
+    conda store servers and workers.
 
     Attributes
     ----------
@@ -32,6 +32,7 @@ class CondaStore:
     log : logging.Logger
         global logger
     """
+
     def __init__(self, config: conda_store_config.CondaStore):
         self.config = config
         self.log = logging.getLogger(__name__)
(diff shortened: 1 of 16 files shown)


conda-store-server/conda_store_server/conda_store.py
```diff
@@ -3,365 +3,34 @@
...
-        config=True,
-    )
-
-    post_update_environment_build_hook = Callable(
-        default_value=None,
-        help="callable function taking conda_store and `orm.Environment` object as input arguments. This function can be used to add custom behavior that will run after an environment's current build changes.",
-        config=True,
-        allow_none=True,
-    )
+class CondaStore:
```
peytondmurray, line 22: Would it be possible to add some documentation about _what this is_? What does this class do, how is it different from the traitlets/FastAPI `CondaStoreServer` application, etc would be very helpful here.

Really thing this seems like a celery worker manager - maybe a different name could provide a clearer idea what this class is for. Or maybe the webserver should be named differently :thinking: What do you think?

peytondmurray (approved):
A few general questions about this:

- `conda_store_server._internal.server.app.CondaStoreServer` still inherits from `traitlets.config.Application`. Is this needed anymore? Like does `traitlets` actually do anything for the `CondaStoreServer` after this change?
- There are still configuration options in the `CondaStoreServer` class. Does it make sense to move these to `CondaStoreServer.config`? For someone unfamiliar with `conda-store`, it will be hard to understand why these config values live on the class while others live on `.config`.

I'm definitely in favor of this, but I'll avoid merging until we get feedback from @trallard and possibly others.

soapy1:
> conda_store_server._internal.server.app.CondaStoreServer still inherits from traitlets.config.Application. Is this needed anymore? Like does traitlets actually do anything for the CondaStoreServer after this change?

Yep, it's still needed in order for there to be `CondaStoreServer` config object. That config object is described in the [docs](https://conda.store/conda-store/references/configuration-options#conda_store_server_internalserverappcondastoreserve), it provides the ability to specify server level configuration like `url_prefix` or `port`. That's in contrast to the `CondaStore` config which is separated out in this PR from the `CondaStore` class. The CondaStore config allows users to set stuff like the storage class and specify the build directory. This config is used by both the conda store server and conda store worker.

> There are still configuration options in the CondaStoreServer class. Does it make sense to move these to CondaStoreServer.config? For someone unfamiliar with conda-store, it will be hard to understand why these config values live on the class while others live on .config.

Ya, I think we can do that. But I would say it's out of scope for this PR. I think conda-store should be working towards:
* not using traitlets for config
* being able to provide config from plugins

in order to achieve these 2 ends we'll need to separate out configuring the `CondaStoreServer` instance from starting the webserver. But, I think not much is gained from doing that before conda-store has a more robust way of reading configs and making them available in the right places. This PR is meant to be a stepping stone to get to that point.

peytondmurray:
Awesome, thanks for the explanation, I'm all clear on this now :rocket:

PR #1026, opened 2024-12-17
Consolidate exception classes :broom: 

soapy1:
## Description

This PR does a little bit of refactoring to consolidate all the `Exception` related classes into one exception module in `conda_store_server.exception`. This is to increase legibility of the code.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/exception.py
+++ b/conda-store-server/conda_store_server/exception.py
@@ -2,18 +2,43 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
+import typing
+
 
 class CondaStoreError(Exception):
-    pass
+    """Exception raised by conda store
+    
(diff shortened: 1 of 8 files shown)


conda-store-server/conda_store_server/exception.py
```diff
@@ -2,18 +2,44 @@
...
+    """Exception raised by conda store
+
+    Attributes
+    ----------
+    message : str
+        message to output to users
+    """
+
+    def __init__(self, msg: str):
+        self.message = msg
```
peytondmurray, line 18: Is the only place this is used `conda_store_server/_internal/server/views/api.py`? If so is there much advantage in introducing the `message` attribute over the standard `.args`?

peytondmurray (approved):
This is a useful consolidation. Thank you!

PR #1033, opened 2024-12-31
Don't output db secret to logs

soapy1:
## Description

The `database_url` also includes the username + password for accessing the db. Outputting this to the logs is not good.

## Pull request checklist


peytondmurray (approved):
Oof, thanks for catching this. I'm going to make an issue for creating a logging formatter to redact sensitive information as a follow up.

Issue #1034, opened 2025-01-02
[BUG] - conda-store requires UI to be enabled in order to use auth functionality


soapy1:
### Describe the bug

When running conda-store without the ui enabled auth no longer works. For example, given the following relevant config (notice the ui is turned off)

```
c.CondaStoreServer.enable_ui = False
c.CondaStoreServer.enable_api = True
c.CondaStoreServer.enable_registry = False
c.CondaStoreServer.enable_metrics = False
```

trying to login to conda store will produce a 500 error. The full error from the logs is shown below. Relevant part of the error:

```
conda-store-server-1  |   File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/requests.py", line 186, in url_for
conda-store-server-1  |     url_path = url_path_provider.url_path_for(name, **path_params)
conda-store-server-1  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
conda-store-server-1  |   File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/routing.py", line 662, in url_path_for
conda-store-server-1  |     raise NoMatchFound(name, path_params)
conda-store-server-1  | starlette.routing.NoMatchFound: No route exists for name "ui_list_environments" and params "".
```

### Expected behavior

conda store auth functionality should not be tied to if the ui is enabled (eg. using the config option `c.CondaStoreServer.enable_ui`).

### How to Reproduce the problem?

To reproduce the problem launch conda-store with docker compose. You'll need the docker-compose.yml and condastore_config.py given by

#### Config files 

`docker-compose.yml`
```
services:
  conda-store-worker:
    image: quansight/conda-store-server:${CONDA_STORE_SERVER_VERSION:-latest}
    volumes:
      - ./docker/assets/environments:/opt/environments:ro
      - ./docker/assets/conda_store_config.py:/opt/conda_store/conda_store_config.py:ro
    depends_on:
      conda-store-server:
        condition: service_healthy
    platform: linux/amd64
    command:
      [
        "conda-store-worker",
        "--config",
        "/opt/conda_store/conda_store_config.py"
      ]

  conda-store-server:
    image: quansight/conda-store-server:${CONDA_STORE_SERVER_VERSION:-latest}
    depends_on:
      postgres:
        condition: service_healthy
      redis:
        condition: service_healthy
    volumes:
[...]

PR #1039, opened 2025-01-07
Login page without UI enabled

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1034

## Description

This PR updates the auth module to be agnostic to the admin ui. It does this by:
* separating the login page from the `base.html` template
* removing references to admin ui endpoints from the auth module

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/_internal/server/templates/login.html
+++ b/conda-store-server/conda_store_server/_internal/server/templates/login.html
@@ -4,16 +4,44 @@
  license that can be found in the LICENSE file.
 -->
 
-{% extends 'base.html' %}
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 
-{% block title %}Login{% endblock %}
+        <!-- Bootstrap CSS -->
+        <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css" integrity="sha384-JcKb8q3iqJ61gNV9KGb8thSsNjpSL0n8PARn9HuZOnIxN0hoP+VmmDGMN5t9UJ0Z" crossorigin="anonymous">
 
-{% block content %}
+        <title>Login</title>
+    </head>
+    <body>
+        <!-- jQuery first, then Popper.js, then Bootstrap JS -->
+        <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js" integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj" crossorigin="anonymous"></script>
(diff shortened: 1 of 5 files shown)


peytondmurray:
I tried testing this by editing the default `conda_store_config.py` and instead of the error I was getting on `main` no exception is raised when I navigate to `localhost:8080`, I just get a 404, as intended:

image

So this seems like a good start. But it sounds like there's other functionality here which I should be testing too? From the issue

> trying to login to conda store will produce a 500 error. The full error from the logs is shown below. Relevant part of the error

When you say login, you mean login to the admin UI?

---

Should there be a test for this? Something like:

1. Use the existing conda-store testclient
2. Set these settings on the testclient:

```python
c.CondaStoreServer.enable_ui = False
c.CondaStoreServer.enable_api = True
c.CondaStoreServer.enable_registry = False
c.CondaStoreServer.enable_metrics = False
```

3. Send a request to the testclient to login URL (check that there's no UI returned)
4. Send requests or call functions from `conda_store_server.server.auth` to make sure they work

soapy1:
> When you say login, you mean login to the admin UI?

Yep. Sorry, I could have been more clear in the issue. Will update that now. 
To see the issue, you'll need to navigate to `localhost:8080/ui` and then click on the `Login` button. Instead of getting to the login screen, there will be an error page.

> Should there be a test for this? Something like:

Do you mean like an integration test to ensure that loging in works when the ui is disabled?

peytondmurray:
Actually I can't get to the login screen at all. When I get when I open up `localhost:8080/` or `localhost:8080/ui` is this:

image

Which I think is expected; there's no UI at all with the options set to the above :point_up:. The REST API appears to work as intended though.

> Do you mean like an integration test to ensure that loging in works when the ui is disabled?

What I mean is that the only way that I know whether this PR fixes #1034 or not is by manually editing my `conda_store_config.py`, running `docker compose up --build`, then checking via the browser that the UI doesn't work. Then I use curl to probe the REST API. I feel like there might be a simpler way of checking this; I don't think a full blown integration test is necessary, is it? Maybe if we just use the `testclient` fixture?

soapy1:
Ah, that's my bad again, you need to go to `localhost:8000`,  not `localhost:8080`. It's important to update your docker compose to include the conda-store-ui
```
conda-store-ui:
    image:  ghcr.io/conda-incubator/conda-store-ui:sha-2469531
    command: "yarn run start:ui"
    profiles:
      - local-dev
    ports:
      - "8000:8000"
    depends_on:
      conda-store-server:
        condition: service_healthy
    platform: linux/amd64
    volumes:
      - ./src:/usr/src/app/src
      - ./style:/usr/src/app/style
      - ./.env:/usr/src/app/.env
    healthcheck:
      test:
        ["CMD", "curl", "--fail", "http://localhost:8000"]
      interval: 20s
      timeout: 10s
      retries: 8
```

This way you have a separate service providing the ui functionality.

soapy1:
>I feel like there might be a simpler way of checking this; I don't think a full blown integration test is necessary, is it? Maybe if we just use the testclient fixture?

Ah, I see what you mean. Great idea! Added :flamingo:

peytondmurray (approved):
Thank you for adding the extra test. Looks good!

PR #1040, opened 2025-01-08
Add env var setting to disable ssl verification for user journey tests

soapy1:
## Description

This PR adds an option when running user journey tests to disable SSL verification. This is useful when running tests against a host that has ssl enabled but is does not have valid certs (eg. using self signed certs). 

In order to disable SSL verification in the user journey tests, set the env var `CONDA_STORE_TEST_VERIFY_SSL=0`. By default, SSL verification will be enabled.

## Pull request checklist


soapy1:
The failing test is unrelated to these changes and is resolved in #1041

peytondmurray:
Related: https://github.com/nebari-dev/nebari/pull/2895

Nebari runs tests against their conda-store deployment. In their test environment, `conda-store` is accessed over https, but does not have valid certs.

peytondmurray (approved):
This looks like it'll be useful for the nebari team :)

PR #1044, opened 2025-01-09
Add settings module

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1005

## Description

This PR introduces a settings module for interacting with conda store settings. It also adds some documentation and enforces rules for how settings get merged at each of the different levels of abstraction that setting apply to. 

There are multiple levels which settings config can be applied to conda store. 

* deployment defaults (controlled by the config file)
* global defaults (controlled by the admin console)
* namespace defaults (controlled by the admin console)
* environment default (controlled by the admin console)

The most specific settings config will take precedence over the more  general setting.

The introduction of this settings object will allow:
* defining a common pattern for accessing settings
* enforcing rules around setting precedence
* providing a common implementation for accessing settings.

 With this change, plugin authors will have better patterns for accessing settings. For example, consider the lock plugin which requires running a [conda command](https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/conda_store_server/_internal/plugins/lock/conda_lock/conda_lock.py#L21)

```
def _conda_command(self, conda_store) -> str:
        with conda_store.session_factory() as db:
            settings = conda_store.get_settings(db=db)
        return settings.conda_command
```

can now become

```
def _conda_command(self, conda_store, namespace=None, environment=None) -> str:
        return conda_store.get_setting(
            "conda_command", namespace=namespace, environment=environment
        ) or "mamba"
```
 
## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/_internal/settings.py
+++ b/conda-store-server/conda_store_server/_internal/settings.py
@@ -2,7 +2,8 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-from typing import Any, Dict
+from typing import Any, Dict, Callable
+import functools
 
 from sqlalchemy.orm import Session
 import pydantic
@@ -16,6 +17,15 @@ class Settings:
         self.db = db
         self.deployment_default = deployment_default.model_dump()
 
+    def _ensure_closed_session(func: Callable):
+        @functools.wraps(func)
+        def wrapper(self, *args, **kwargs):
+            result = func(self, *args, **kwargs)
+            self.db.close()
+            return result
+        return wrapper
+
+    @_ensure_closed_session
     def set_settings(
         self,
         namespace: str = None,
@@ -65,6 +75,7 @@ class Settings:
 
         api.set_kvstore_key_values(self.db, prefix, data)
 
+    @_ensure_closed_session
     def get_settings(
         self, namespace: str = None, environment_name: str = None
     ) -> schema.Settings:
--- a/conda-store-server/conda_store_server/_internal/settings.py
+++ b/conda-store-server/conda_store_server/_internal/settings.py
@@ -129,7 +129,10 @@ class Settings:
 
     @_ensure_closed_session
     def get_setting(
-        self, key: str, namespace: str | None = None, environment_name: str | None = None
+        self,
+        key: str,
+        namespace: str | None = None,
+        environment_name: str | None = None,
     ) -> Any:
         """Get a given setting at the given level of specificity. Will short
         cut and look up global setting directly even if a namespace/environment
--- a/conda-store-server/conda_store_server/_internal/settings.py
+++ b/conda-store-server/conda_store_server/_internal/settings.py
@@ -122,7 +122,7 @@ class Settings:
 
     @_ensure_closed_session
     def get_setting(
-        self, name: str, namespace: str = None, environment_name: str = None
+        self, key: str, namespace: str = None, environment_name: str = None
     ) -> Any:
         """Get a given setting at the given level of specificity. Will short
         cut and look up global setting directly even if a namespace/environment
@@ -130,7 +130,7 @@ class Settings:
 
         Parameters
         ----------
-        name : str
+        key : str
             name of the setting to return
         namespace : str, optional
             namespace to use to retrieve settings
@@ -143,7 +143,7 @@ class Settings:
             setting value, merged for the given level of specificity
         """
         
-        field = schema.Settings.model_fields.get(name)
+        field = schema.Settings.model_fields.get(key)
         if field is None:
             return
         
@@ -155,9 +155,9 @@ class Settings:
                 prefixes.append(f"setting/{namespace}/{environment_name}")
 
         # start building settings with the least specific defaults
-        result = self.deployment_default.get(name)
+        result = self.deployment_default.get(key)
         for prefix in prefixes:
-            value = api.get_kvstore_key(self.db, prefix, name)
+            value = api.get_kvstore_key(self.db, prefix, key)
             if value is not None:
                 result = value
 
--- a/conda-store-server/conda_store_server/_internal/settings.py
+++ b/conda-store-server/conda_store_server/_internal/settings.py
@@ -29,8 +29,8 @@ class Settings:
     @_ensure_closed_session
     def set_settings(
         self,
-        namespace: str = None,
-        environment_name: str = None,
+        namespace: str | None = None,
+        environment_name: str | None = None,
         data: Dict[str, Any] = {},
     ):
         """Persist settings to the database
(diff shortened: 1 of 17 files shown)


conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+                    k: v for k, v in new_settings.items() if k not in global_fields
+                }
+                settings.update(new_settings)
+
+        return schema.Settings(**settings)
+
+    @_ensure_closed_session
+    def get_setting(
+        self, key: str, namespace: str = None, environment_name: str = None
+    ):
```
peytondmurray, line 133:
```suggestion
    ) -> Any:
```

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+                new_settings = {
+                    k: v for k, v in new_settings.items() if k not in global_fields
+                }
+                settings.update(new_settings)
+
+        return schema.Settings(**settings)
+
+    @_ensure_closed_session
+    def get_setting(
+        self, key: str, namespace: str = None, environment_name: str = None
```
peytondmurray, line 132:
```suggestion
        self, key: str, namespace: str | None = None, environment_name: str | None = None
```

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+            self.db.close()
+            return result
+
+        return wrapper
+
+    @_ensure_closed_session
+    def set_settings(
+        self,
+        namespace: str = None,
+        environment_name: str = None,
```
peytondmurray, line 33:
```suggestion
        namespace: str | None = None,
        environment_name: str | None = None,
```

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+        data: Dict[str, Any] = {},
```
peytondmurray, line 34:
```suggestion
        data: Dict[str, Any] | None = None,
```

I know it's not getting modified below, but it's still best to avoid mutable default arguments (it'll generate a lint error anyway if we ever turn on more rules). We need an `if data is None: ...` block below if this suggestion is made.

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+            raise ValueError(f"Invalid setting keys {invalid_keys}")
+
+        for key, value in data.items():
+            field = schema.Settings.model_fields[key]
+            global_setting = field.json_schema_extra["metadata"]["global"]
+            if global_setting and (
+                namespace is not None or environment_name is not None
+            ):
+                raise ValueError(
+                    f"Setting {key} is a global setting cannot be set within namespace or environment"
```
peytondmurray, line 59:
```suggestion
                    f"Setting {key} is a global setting and cannot be set within a namespace or environment"
```

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+                )
+
+            try:
+                validator = pydantic.TypeAdapter(field.annotation)
+                validator.validate_python(value)
+            except Exception as e:
```
peytondmurray, line 65:
```suggestion
            except pydantic.ValidationError as e:
```

Should we just let validation errors propagate instead of catching them? The default validation error message should be very similar, I think.

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+                raise ValueError(
+                    f"Invalid parsing of setting {key} expected type {field.annotation} ran into error {e}"
+                )
+
+        if namespace is not None and environment_name is not None:
+            prefix = f"setting/{namespace}/{environment_name}"
+        elif namespace is not None:
+            prefix = f"setting/{namespace}"
+        else:
+            prefix = "setting"
```
peytondmurray, line 75: I think the setting can inadvertently be set at the global level if `namespace is None` and `environment_name is not None`. I think something similar to the prefix logic in `get_settings` would work though.

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,170 @@
...
+
+        api.set_kvstore_key_values(self.db, prefix, data)
+
+    @_ensure_closed_session
+    def get_settings(
+        self, namespace: str = None, environment_name: str = None
```
peytondmurray, line 81:
```suggestion
        self, namespace: str | None = None, environment_name: str | None = None
```

peytondmurray (changes requested):
Looks good, just one minor change and a bunch of annotations.

Thanks for this, it's going to make this much clearer for users :+1:

conda-store-server/conda_store_server/_internal/settings.py
soapy1: Sounds good to me!

conda-store-server/conda_store_server/_internal/settings.py
soapy1: oh, good catch. I think that is probably a programmer error and this function should probably throw an exception. Will also add a test for that!

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,178 @@
...
+from typing import Any, Callable, Dict
+
+import pydantic
+from sqlalchemy.orm import Session
+
+from conda_store_server import api
+from conda_store_server._internal import schema
+
+
+class Settings:
```
peytondmurray, line 15: This is currently part of the private API - is there a scenario in which these settings would be needed by e.g. nebari?

conda-store-server/conda_store_server/_internal/settings.py
```diff
@@ -0,0 +1,178 @@
...
+        field = schema.Settings.model_fields.get(key)
+        if field is None:
+            return None
+
+        prefixes = ["setting"]
+        if field.json_schema_extra["metadata"]["global"] is False:
+            if namespace is not None:
+                prefixes.append(f"setting/{namespace}")
+            if namespace is not None and environment_name is not None:
+                prefixes.append(f"setting/{namespace}/{environment_name}")
```
peytondmurray, line 169:
```suggestion
            if namespace is not None:
                prefixes.append(f"setting/{namespace}")
                if environment_name is not None:
                    prefixes.append(f"setting/{namespace}/{environment_name}")
```

peytondmurray (approved):
Minor changes. One question: should setting an invalid setting raise an exception? Right now if you set a setting that isn't real there's no way for you to know other than checking whether it is `None`.

conda-store-server/conda_store_server/_internal/settings.py
soapy1: I think this should be private. Users are able to interact with settings using the public [`CondaStore`](conda-store-server/conda_store_server/conda_store.py) object. For example, plugins or other consumers can do stuff like

```
def do_something(self, conda_store) -> str:
    . . .
    conda_store.get_setting("conda_command", namespace=namespace, environment=environment)
   . . . 
```

PR #1045, opened 2025-01-09
Get database url from conda-store config in tests

soapy1:
https://github.com/conda-incubator/conda-store/pull/1023 introduced splitting conda store config from the conda store class. This pr updates tests that got left out.


peytondmurray (approved):
Thanks for the fix :+1:

peytondmurray:
Merging, as it was only the cleanup that had failed before, and anyway GH actions is having issues today.

Issue #1042, opened 2025-01-09
[ENH] - Add new endpoint(s) to avoid needing internal knowledge of conda-store's database


peytondmurray:
### Feature description

See https://github.com/nebari-dev/nebari/pull/2815 for more information about what internal functions are needed by nebari.

We should create whatever REST endpoints are necessary for nebari to avoid having to call internal conda-store functions.

### Value and/or benefit

Access to the internal python API should not be necessary; fixing this will reduce the chance of breaking changes for users.

marcelovilla:
@peytondmurray thanks for opening this issue!

Just to clarify something, are you considering adding the functionality Nebari relies on as REST endpoints, or expose them in the conda-store Python API? If it's the former, I think it's already covered, at least judging from what's available in the [REST API docs](https://conda.store/rest-api.html). Here's a small table outlining the Python functions we're using and (what I think) are the equivalent REST endpoints:

| Python function                                                                                                                                                                                                                                    | REST API endpoint                           |
|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------|
| [`get_namespace`](https://github.com/nebari-dev/nebari/blob/e7480a096d01f23c500cf73e3a232399884096e3/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L424)                    | `GET /api/v1/namespace/{namespace}/`        |
| [`ensure_namespace`](https://github.com/nebari-dev/nebari/blob/e7480a096d01f23c500cf73e3a232399884096e3/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L426C24-L426C41)      | `POST /api/v1/namespace/{namespace}/`       |
| [`create_namespace_role`](https://github.com/nebari-dev/nebari/blob/e7480a096d01f23c500cf73e3a232399884096e3/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L191C17-L191C38) | `POST /api/v1/namespace/{namespace}/role`   |
[...]

Issue #1031, opened 2024-12-23
Plugin-ize authentication backend


soapy1:
### Context

Currently, conda-store supports a few different auth backends. However, projects that use conda-store may require extending the available authentication options. For example, nebari defines it's own [KeyCloakAuthentication](https://github.com/nebari-dev/nebari/blob/b42a90352f6b38ffa0b46413900039e7de507165/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L153) class. It would be better if nebari could install an authentication plugin for keycloak instead of defining the class within the config file.

### Value and/or benefit

conda-store consumers can extend conda-store to use custom authentication providers.

PR #1046, opened 2025-01-14
Move auth related schema to inside the auth module

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1042

## Description

This will allow users to define their own auth modules without dependings on internal schemas. 

In particular this enables nebari to not depend on the `conda_store_server._internal.schema` module for setting up the keycloak auth backend (ref. https://github.com/nebari-dev/nebari/blob/f39467679ce2dac84385f1d6d4ec698befd7c126/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L14).

This is intended as a stop gap solution until [conda-store support authentication plugins](https://github.com/conda-incubator/conda-store/issues/1031). As such, this PR aims to not make any changes that will make it more difficult to move towards auth plugins.

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/server/auth.py
+++ b/conda-store-server/conda_store_server/server/auth.py
@@ -33,6 +33,7 @@ from traitlets import (
 from traitlets.config import LoggingConfigurable
 
 from conda_store_server import api
+from conda_store_server.server import schema as auth_schema
 from conda_store_server._internal import environment, orm, schema, utils
 from conda_store_server._internal.server import dependencies
 
@@ -56,7 +57,7 @@ class AuthenticationBackend(LoggingConfigurable):
         config=True,
     )
 
-    def encrypt_token(self, token: schema.AuthenticationToken):
+    def encrypt_token(self, token: auth_schema.AuthenticationToken):
         return jwt.encode(token.model_dump(), self.secret, algorithm=self.jwt_algorithm)
 
     def decrypt_token(self, token: str):
@@ -68,7 +69,7 @@ class AuthenticationBackend(LoggingConfigurable):
                 authentication_token = self.predefined_tokens[token]
             else:
                 authentication_token = self.decrypt_token(token)
-            return schema.AuthenticationToken.model_validate(authentication_token)
+            return auth_schema.AuthenticationToken.model_validate(authentication_token)
         except Exception:
             return None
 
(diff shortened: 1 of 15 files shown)


conda-store-server/conda_store_server/server/auth.py
```diff
@@ -35,6 +35,7 @@
 from conda_store_server import api
 from conda_store_server._internal import environment, orm, schema, utils
 from conda_store_server._internal.server import dependencies
+from conda_store_server.server import schema as auth_schema
```
soapy1, line 38: I think this module being called `conda_store_server.server` is pretty misleading. This is left over from some other related work around moving modules into an `_internal` module. Punting on renaming this for now.

peytondmurray: I agree we should leave this alone for now - but it's not a helpful name.

docusaurus-docs/conda-store/references/auth.md
```diff
@@ -34,7 +34,7 @@ user. Similar to JupyterHub
 method. This method is the primary way to customize authentication. It
 is responsible for checking that the user credentials to login are
 correct as well as returning a dictionary following the schema
-`conda_store_server._internal.schema.AuthenticationToken`. This stores a
+`conda_store_server.auth.schema.AuthenticationToken`. This stores a
```
peytondmurray, line 37:
```suggestion
`conda_store_server.server.schema.AuthenticationToken`. This stores a
```

peytondmurray (approved):
Let me see if I understand correctly: now `conda_store_server._internal.schema` contains _only_ db schema, right? Not REST API schema for the authorization etc? If so, it would be useful to have a module-level docstring there that emphasizes that it is only for db schema. I guess in the future we can rename this do `dbschema.py` or similar.

Did you consider putting the authorization schema into `conda_store_server.server.auth`?

soapy1:
> Let me see if I understand correctly: now conda_store_server._internal.schema contains only db schema, right? Not REST API schema for the authorization etc? 

I don't think so. I think it's mostly API schema in `conda_store_server._internal.schema`. And also, some DB schema. We'll need to do some more work to clean this up. I think this change is a good start with removing all the auth related schema from that module.

> Did you consider putting the authorization schema into conda_store_server.server.auth? I can't tell if there's an advantage of having it in a separate module.

I did consider it. I decided against it because I think `conda_store_server.server.auth` has a bunch of stuff that I'd like to break down. I think it makes sense to do that as part of https://github.com/conda-incubator/conda-store/issues/1031. I don't think it makes too much difference, other than cleaning up the import statements some. Open to moving things around if you think that's better.

peytondmurray:
All good, thank you for the response!

Issue #947, opened 2024-11-04
[ENH] - Implement archiving of old builds


peytondmurray:
### Feature description

This issue for tracking work on archiving old builds. Archived builds should only keep the lockfile; if they become activated again, the lockfile should be used to reinstall the environment from scratch. Proposed approach:

1. Create an endpoint to archive a build: `/api/v1/build/{build_id}/archive/`
2. Add a button to `conda-store-ui` to archive old builds of an environment.

### Value and/or benefit

Archiving old builds will save disk space for deployments with many old builds.

asmacdo:
I'd like to be able to backup using this functionality, but without deleting the artifacts.

PR #1049, opened 2025-01-16
Archive builds

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/947

Depends on #1047  and #1048 

## Description

This PR adds the ability to archive builds. To archive a build means to:
* delete all artifacts except the build log and the lockfile
* marks the build status as `ARCHIVED`
* sets the build `archived_on` field

### Using the REST api
```
$ export CS_TOKEN=<get a conda-store api token>

# look at existing builds
$ curl  --header "Authorization: bearer ${CS_TOKEN}" http://localhost:8080/conda-store/api/v1/build/ | jq

# choose a build from the output above to archive
$ curl -X POST  --header "Authorization: bearer ${CS_TOKEN}" http://localhost:8080/conda-store/api/v1/build/<build_id>/archive/ | jq

# inspect the build to verify that artifacts have been deleted + metadata has been updated
$ curl  --header "Authorization: bearer ${CS_TOKEN}" http://localhost:8080/conda-store/api/v1/build/<build_id>/ | jq
```

## Pull request checklist


conda-store-server/conda_store_server/_internal/worker/tasks.py
```diff
@@ -298,11 +298,41 @@ def task_delete_build(self, build_id):
...
         db.commit()
 
 
+@shared_task(base=WorkerTask, name="task_archive_build", bind=True)
+def task_archive_build(self, build_id):
+    conda_store = self.worker.conda_store
+    with conda_store.session_factory() as db:
+        build = api.get_build(db, build_id)
+
+        archive_save_artifacts = [
```
soapy1, line 313: there is a setting to allow users to configure what artifacts should be deleted when a build is deleted. Should we add a setting that allows users to customize what artifacts are deleted on archive as well?

PR #1050, opened 2025-01-16
Fix user-journey tests for running against non-default targets

soapy1:
## Description

This PR makes a variety of fixes to the user journey tests in order to run them against non default targets (eg. like nebari's conda store deployment). This is required for https://github.com/nebari-dev/nebari/pull/2895

### Fix `CONDA_STORE_TEST_VERIFY_SSL`
Previously setting any value for `CONDA_STORE_TEST_VERIFY_SSL`  resulted in enabling ssl verification. So, setting `CONDA_STORE_TEST_VERIFY_SSL=0` or `CONDA_STORE_TEST_VERIFY_SSL=false` enabled ssl verification. This is because strings are truthy. 

This PR fixes this by checking if the passed in env var is one of `0`, `false` or `False` in order to determine if tests should verify SSL.

### Use `CONDA_STORE_TOKEN` 
Previously all but one conda-store user journey tests used the `CONDA_STORE_TOKEN` as auth when running tests. This PR extends this so that all user journey tests can use the provided auth token for auth.

### Ensure new api contexts are valid
Previously, some tests created new utils.API instances. However, the generated tokens had invalid expiry values and did not respect the SSL settings. 

## Pull request checklist

### DIFF
--- a/conda-store-server/tests/user_journeys/test_user_journeys.py
+++ b/conda-store-server/tests/user_journeys/test_user_journeys.py
@@ -96,6 +96,7 @@ def test_user_login_and_create_shared_environment(
 
     dev_api = utils.API(
         base_url=base_url,
+        verify_ssl=verify_ssl,
         token=api.create_token(
             namespace,
             "developer",
--- a/conda-store-server/tests/user_journeys/test_user_journeys.py
+++ b/conda-store-server/tests/user_journeys/test_user_journeys.py
@@ -29,7 +29,8 @@ def verify_ssl() -> bool:
     If CONDA_STORE_TEST_VERIFY_SSL is 0, ssl verification is disabled. For all
     other inputs, ssl verification is enabled.
     """
-    return bool(os.environ.get("CONDA_STORE_TEST_VERIFY_SSL", True))
+    verify = os.environ.get("CONDA_STORE_TEST_VERIFY_SSL", "true")
+    return not verify.lower() in ("0", "false")
 
 
 @pytest.mark.user_journey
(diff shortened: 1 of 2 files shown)


conda-store-server/tests/user_journeys/test_user_journeys.py
```diff
@@ -29,7 +29,8 @@ def verify_ssl() -> bool:
     If CONDA_STORE_TEST_VERIFY_SSL is 0, ssl verification is disabled. For all
     other inputs, ssl verification is enabled.
     """
-    return bool(os.environ.get("CONDA_STORE_TEST_VERIFY_SSL", True))
+    verify = os.environ.get("CONDA_STORE_TEST_VERIFY_SSL", "true")
+    return verify.lower() not in ("0", "false")
```
peytondmurray, line 33: Nice catch :grimacing:

conda-store-server/tests/user_journeys/utils/api_utils.py
```diff
@@ -136,7 +134,7 @@ def create_token(
         """Create a token with a specified role in a specified namespace."""
         json_data = {
             "primary_namespace": default_namespace,
-            "expiration": time_utils.get_iso8601_time(1),
+            # "expiration": time_utils.get_iso8601_time(1),
```
peytondmurray, line 137: Hmm, I thought we _needed_ an expiration time? If we don't, let's just remove this line.

peytondmurray (approved):
Thanks for this! Just one minor change requested.

conda-store-server/tests/user_journeys/utils/api_utils.py
soapy1: Nope, by default the api will set the expiry to the same expiration as the token that is making the request. 

This line in particular was not generating the expiration in the right format so it was causing an error. By removing this line the generated token will get the default expiration value.

PR #1054, opened 2025-01-22
Ensure settings are retrieved for the right level

soapy1:
## Description

Follow up to https://github.com/conda-incubator/conda-store/pull/1044

It ensures that there is consistency in how settings are retrieved throughout the code base and ensures that settings are pulled at the right specificity. 

## Pull request checklist

### DIFF
--- a/conda-store-server/conda_store_server/_internal/worker/tasks.py
+++ b/conda-store-server/conda_store_server/_internal/worker/tasks.py
@@ -82,10 +82,10 @@ def task_initialize_worker(self):
 @shared_task(base=WorkerTask, name="task_watch_paths", bind=True)
 def task_watch_paths(self):
     conda_store = self.worker.conda_store
+    # filesystem_namespace is a global setting, don't need to specify namespace/environment
+    filesystem_namespace = conda_store.get_setting("filesystem_namespace")
 
     with conda_store.session_factory() as db:
-        settings = conda_store.get_settings()
-
         conda_store.configuration(db).update_storage_metrics(
             db, conda_store.config.store_directory
         )
@@ -96,7 +96,7 @@ def task_watch_paths(self):
                 conda_store.register_environment(
                     db,
                     specification=yaml.safe_load(f),
-                    namespace=settings.filesystem_namespace,
+                    namespace=filesystem_namespace,
                     force=False,
                 )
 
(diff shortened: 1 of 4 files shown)


peytondmurray:
See also #1005.

conda-store-server/conda_store_server/_internal/worker/tasks.py
```diff
@@ -153,33 +153,33 @@ def task_update_conda_channels(self):
...
+
+    is_locked = False
+
+    if conda_store.config.redis_url is not None:
+        lock = conda_store.redis.lock(task_key, timeout=60 * 15)  # timeout 15min
+    else:
+        lockfile_path = os.path.join(f"/tmp/task_lock_{task_key}")
+        lock = FileLock(lockfile_path, timeout=60 * 15)
 
+    with conda_store.session_factory() as db:
```
peytondmurray, line 182: I guess if this line is being moved to the place where it is actually needed, it can go further down inside `if is_locked`?

conda-store-server/conda_store_server/conda_store_config.py
```diff
@@ -51,13 +51,15 @@ def conda_store_validate_action(
     namespace: str,
     action: auth_schema.Permissions,
 ) -> None:
-    settings = conda_store.get_settings()
+    # get global setting storage thershold, no need to specify a namespace/environment
+    storage_threshold = conda_store.get_setting("storage_threshold")
```
peytondmurray, line 55: Hmm, this is the second place this comment has shown up. Would it be useful to have `conda_store.get_setting("settings/<namespace>/<environment>/storage_threshold")` raise an exception that says "storage_threshold is a global setting, and does not exist on <namespace>/<environment>"?

Otherwise it is entirely up to the maintainers to know which settings are global and which aren't (which actually isn't that bad IMO).

peytondmurray (commented):
A few minor changes/comments, otherwise this is a solid improvement.

conda-store-server/conda_store_server/conda_store_config.py
peytondmurray: Okay, I just realized there's already an exception for this. I'd be fine if the comments about this were removed, in that case. I'm also fine if they stay.

soapy1: Currently the error get's thrown if you try to set the setting for a particular namespace/environment. 

Getting settings works by collecting all the places that a setting is defined and then choosing the one that has the highest level of specificity. So, doing `storage_threshold = conda_store.get_setting("storage_threshold")` is the same as `storage_threshold = conda_store.get_setting("storage_threshold", "mynamespace", "myenv")`. 

I think this it would be pretty cumbersome to force an error if the user specified the namespace and environment erroneously. Like, then the programmer will need to know if things are global or not and call the function keeping that in mind. Using the current approach, the programmer doesn't need to know about what settings are global when retrieving settings.

Hmm, I think I can come up with some better comment/docstring to make this more clear. Thanks for pointing this out!

PR #1055, opened 2025-01-22
Set appropriate settings from a plugin context

soapy1:
## Description

Follow up to https://github.com/conda-incubator/conda-store/pull/1044

In order for a plugin to get settings at the right level of specificity, they must have information about the running environment and namespace. This PR:
* adds environment and namespace info into the plugin context
* updates the conda-lock plugin to use this info to get the correct conda command

## Pull request checklist


conda-store-server/tests/_internal/plugins/lock/test_conda_lock.py
```diff
@@ -74,6 +74,33 @@ def test_solve_lockfile_simple(conda_store, simple_specification):
...
 
+@mock.patch("conda_store_server._internal.plugins.lock.conda_lock.conda_lock.run_lock")
+def test_solve_right_conda_command(mock_run_lock, conda_store, simple_specification):
+    # Update conda_command settings
+    conda_store.set_settings(data={"conda_command": "conda"})
+
+    # Dump dummy data to the expected lockfile output location
+    def run_lock_side_effect(lockfile_path, **kwargs):
+        with open(lockfile_path, "w") as f:
+            yaml.dump({"foo": "bar"}, f)
```
peytondmurray, line 85: Do we even need to write to a file here?

soapy1: Good call out! Yes, we need to write a file because the last part of the `lock_environment` function is to read the output lockfile and return the contents to the user. So if the file isn't there an error is thrown about the missing file, for example, `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpajob34th/conda-lock.yaml'`.

I think it makes sense for this to be an error. But open to change if that seems overly strict to you.

soapy1: Actually, this is used in another test as well. I'll refactor some so we can reuse this in all the places it's required.

peytondmurray: > So if the file isn't there an error is thrown about the missing file, for example, FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmpajob34th/conda-lock.yaml'

Yeah, I was wondering if there would be an issue here. Can we just mock out the call to read the file too?

soapy1: Ah ya, not relying on opening/closing files for unit tests would be good. So I looked into this some, and I think it's a bit tricky. I tried a couple of things:
* the file that is getting read is created inside the `run_lock` function. So I tried mocking out `open`, but this interferes with running `subrocess.Popen`.
* I considered passing in the lock file destination (so that it can be mocked out more directly). I think that isn't really appropriate for this function signature, so I don't think is a good option.

I'm open to more suggestions to try out. Otherwise, will mull it over a bit more.

peytondmurray: Okay, if it's impractical to do, let's just not worry about it if that works for you.

Issue #1053, opened 2025-01-22
[BUG] - conda-store pins very old (unused?) jupyterhub


minrk:
### Describe the bug

conda-store [pins](https://github.com/conda-incubator/conda-store/blob/c8f8a8123469805a8235c12e160db9e321f79697/conda-store/pyproject.toml#L49) `jupyterhub<2.0`, which is outdated since 2021. That's super old and shouldn't be used ~anywhere. It also doesn't appear to be used by conda-store, so I'm not sure why it's there. The dependency was _added_ in 2024.11 by #827 which shifts some environment requirements for the image to strict package runtime dependencies, meaning they conflict for every installation, not just a specific environment.

I assume the right answer is to remove the dependency from the package, but I don't know why it was added so didn't open a PR, and don't know if jupyterhub should be re-added somewhere else.

### Expected behavior

conda-store doesn't conflict with recent jupyterhub

### How to Reproduce the problem?

`pip install 'conda-store>=2024.11' 'jupyterhub>=4'`

### Output

```
ERROR: Cannot install conda-store==2024.11.1, conda-store==2024.11.2 and jupyterhub>=4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested jupyterhub>=4
    conda-store 2024.11.2 depends on jupyterhub<2.0.0
    The user requested jupyterhub>=4
    conda-store 2024.11.1 depends on jupyterhub<2.0.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip to attempt to solve the dependency conflict
```

soapy1:
Heya 👋 thanks for bringing this up! We'll have this fixed up for next release.

trallard:
It seems the dependencies from `conda-store/environment.yaml` were somehow added as a direct dependency of conda-store (but shouldn't have as JHub, JLab, and the other Jupyter-related ones are not needed, this env should only be used in the examples).

@soapy1 can you do a full check of the deps in `pyproject.toml` in any case?

PR #1056, opened 2025-01-23
Trim conda-store package dependencies

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1053

## Description

The conda-store (cli) package does not have dependencies on jupyter - cleaning up these dependencies. 

For reference, you can also check against the conda package recipe https://github.com/conda-incubator/conda-store/blob/main/recipe/meta.yaml#L34-L43 (which does not express any dependencies on jupyter).

## Pull request checklist

### DIFF
--- a/conda-store/pyproject.toml
+++ b/conda-store/pyproject.toml
@@ -46,10 +46,6 @@ dependencies = [
   "yarl",
   "aiohttp>=3.8.1",
   "ruamel.yaml",
-  "jupyterhub<2.0.0",
-  "jupyterlab>=4.0.0",
-  "jupyter-launcher-shortcuts",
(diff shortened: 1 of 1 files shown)


trallard:
Looks good, but while you are it can you do a thorough check of all our deps in case there are others that should be removed or bumped?

soapy1:
> can you do a thorough check of all our deps in case there are others that should be removed or bumped?

Yep, done. The pyproject.yoml and conda recipes are also now matching.

peytondmurray (approved):
I wonder if there's some automated tooling to check whether certain dependencies are used...

Anyway, looks good :rocket:

PR #1061, opened 2025-01-28
[MAINT] Deprecate `/api/v1/environments/`

peytondmurray:
## Description

This PR deprecates the `/api/v1/environments/` endpoint, which has been replaced with cursor-based pagination in `/api/v2/environments/`.

## Pull request checklist

### DIFF
--- a/conda-store-server/tests/_internal/server/views/test_api.py
+++ b/conda-store-server/tests/_internal/server/views/test_api.py
@@ -240,6 +240,8 @@ def test_api_list_environments_auth(
 
     r = model.model_validate(response.json())
     assert r.status == schema.APIStatus.OK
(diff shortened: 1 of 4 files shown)


conda-store-server/tests/_internal/server/views/test_api.py
```diff
@@ -1189,7 +1195,6 @@ def test_api_list_environments_paginate(
     cursor = None
     cursor_param = ""
     while cursor is None or cursor != Cursor.end():
-        # breakpoint()
```
soapy1, line 1192: Looks like this can be removed

soapy1 (approved):
Looks great, just one little line to clean up! Thanks :palm_tree:

Issue #1062, opened 2025-01-29
[BUG] - GET /api/v1/namespace/ returns a 500


dcmcand:
### Describe the bug

when we deployed our latest rc to https://nebari.quansight.dev/ conda-store won't load any namespaces. Looking at the logs showed the following stack trace
```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
    await self.app(scope, receive, _send)
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/base.py", line 185, in __call__
    with collapse_excgroups():
         ^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/conda-store-server/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/_utils.py", line 82, in collapse_excgroups
    raise exc
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/base.py", line 187, in __call__
    response = await self.dispatch_func(request, call_next)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/conda_store_server/_internal/server/app.py", line 256, in conda_store_middleware
    response = await call_next(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/base.py", line 163, in call_next
    raise app_exc
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/base.py", line 149, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/sessions.py", line 85, in __call__
    await self.app(scope, receive, send_wrapper)
  File "/opt/conda/envs/conda-store-server/lib/python3.12/site-packages/starlette/middleware/cors.py", line 85, in __call__
    await self.app(scope, receive, send)
[...]

PR #1063, opened 2025-01-29
Allow Namespace pydantic model to have a null metadata_ entry

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1062

## Description

This PR updates the pydantic model for `Namespace` so that the `metadata_` field can be null. This field is nullable, and the default value is an empty dict:

```
class Namespace(Base):
    """Namespace for resources"""

    . . .

    metadata_: Mapped[Optional[dict]] = mapped_column(JSON, default=dict, nullable=True)

    . . .
```

In order for pydantic models to allow values to be `None`, they must set them as `Optional`. If not, an error will be thrown. 

In older deployments on conda-store it is possible for this field to actually be `NULL` (see the explanation https://github.com/conda-incubator/conda-store/issues/1062#issuecomment-2622435285), resulting in an error.

### Additional notes

While sorting this out I also checked for other orm fields that were explicitly set to `nullable=True` and ensured that their pydantic counterparts are set to `Optional`.

## Pull request checklist

### DIFF
--- a/conda-store-server/tests/_internal/server/views/test_api.py
+++ b/conda-store-server/tests/_internal/server/views/test_api.py
@@ -175,7 +175,9 @@ def test_api_list_namespace_auth(testclient, seed_conda_store, authenticate):
     assert sorted([_.name for _ in r.data]) == ["default", "namespace1", "namespace2"]
 
 
-def test_api_list_namespace_including_missing_metadata_(testclient, seed_namespace_with_edge_cases, authenticate):
+def test_api_list_namespace_including_missing_metadata_(
+    testclient, seed_namespace_with_edge_cases, authenticate
+):
     response = testclient.get("api/v1/namespace")
     response.raise_for_status()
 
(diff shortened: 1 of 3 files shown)


conda-store-server/tests/_internal/server/views/test_api.py
```diff
@@ -175,6 +175,18 @@ def test_api_list_namespace_auth(testclient, seed_conda_store, authenticate):
     assert sorted([_.name for _ in r.data]) == ["default", "namespace1", "namespace2"]
 
 
+def test_api_list_namespace_including_missing_metadata_(
+    testclient, seed_namespace_with_edge_cases, authenticate
+):
```
peytondmurray, line 180:
```suggestion
):
    """Test that a namespace with metadata_ = None can be retrieved. 
    
    See https://github.com/conda-incubator/conda-store/issues/1062
    for additional context.
    """
```

peytondmurray (approved):
I added a comment that provides context here, but otherwise looks good. :ship:

PR #1067, opened 2025-01-29
Build and push docker images for multiple platforms in one step

soapy1:
ref: https://docs.docker.com/build/ci/github-actions/multi-platform/

### DIFF
--- a/.github/workflows/build_docker_image.yaml
+++ b/.github/workflows/build_docker_image.yaml
@@ -44,6 +44,8 @@ jobs:
 
       - name: "Set up Docker Buildx 🏗"
         uses: docker/setup-buildx-action@v3
+        # Required for building multi-platform images. See
+        # https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images
         with:
           daemon-config: |
             {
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -150,6 +147,14 @@ jobs:
 
       - name: "Set up Docker Buildx 🏗"
         uses: docker/setup-buildx-action@v3
+        with:
```
peytondmurray, line 150:
```suggestion
        # Required for building multi-platform images. See
        # https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images
        with:
```

.github/workflows/build_docker_image.yaml
```diff
@@ -47,6 +44,14 @@ jobs:
 
       - name: "Set up Docker Buildx 🏗"
         uses: docker/setup-buildx-action@v3
+        with:
```
peytondmurray, line 47:
```suggestion
        # Required for building multi-platform images. See
        # https://docs.docker.com/build/ci/github-actions/multi-platform/#build-and-load-multi-platform-images
        with:
```

peytondmurray (approved):
Thanks for the fix! I added some comments to the configuration options, which are otherwise extremely unintuitive.

Issue #1068, opened 2025-01-30
conda-store-server conda package dependencies are out of sync with pyproject.toml dependencies


soapy1:
### Context

There is an inconsistency between the dependencies required by conda-store-server as defined in the [conda-forge package](https://github.com/conda-forge/conda-store-feedstock/blob/main/recipe/meta.yaml#L68) and the [conda-store-server pyproject.toml](https://github.com/conda-incubator/conda-store/blob/main/conda-store-server/pyproject.toml#L39).

For example, you can see the issue if you use `pip check`
```bash
# create a clean environment to test
$ conda create -n tmp-test python=3.11     
$ conda activate tmp-test

# install the conda-store-server package
$ conda install conda-store-server
$ conda-store-server --version   
2025.1.1

# check
$ python -m pip check                                             
conda-store-server 2025.1.1 requires flower, which is not installed.
conda-store-server 2025.1.1 requires psycopg2-binary, which is not installed.
conda-store-server 2025.1.1 has requirement setuptools<70, but you have setuptools 75.8.0.

```

ref: https://github.com/conda-forge/conda-store-feedstock/pull/52#issuecomment-2625309277

### Value and/or benefit

fixes the issue exposed in https://github.com/conda-forge/conda-store-feedstock/pull/52

PR #1069, opened 2025-01-30
Fix conda store server deps

soapy1:
Fixes https://github.com/conda-incubator/conda-store/issues/1068

## Description

* moves `flower` into dev dependencies (since it is not actually used by the conda-store-server package)
* add dependencies missing in the conda package recipe
* add `pip check` test to conda package recipe

## Pull request checklist


peytondmurray (approved):
Thanks!

